Issue #78, opened 2014-03-29
Support BigTextStyle notifications


juliocbcotta:
Hi, I would like to know if it is possible to make the BigTextStyle feature the default style of text when available. I mean, many push notifications needs more than one line to display the whole text, so in Android 4.1+ it could be archived using BigTextStyle.
Can you consider this for the library?

Thanks.

PR #148, opened 2014-12-06
Add support for big style notifications

### LINKED ISSUE #145: GCM Receiver / Android Style Guides
Google is starting to be vigilant about Push notifications following the [style guide](http://developer.android.com/design/patterns/notifications.html) and clients are starting to get nastygrams about being in violation.

Our standard [gcm receiver](https://github.com/mixpanel/mixpanel-android/blob/c772a91149ca1f7799f722b6e0aa6242a6a8afd0/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java) class does not do a few of the things required; most egregiously nesting subsequent notifications but also things like needing a white on transparent app icon image.

anas-ambri:
Add support for big style notifications for API level >= 16
Pass custom data along to the PendingIntent and thus handling activity
Fixes #145 #78

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
@@ -12,6 +12,7 @@ import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.PackageManager.NameNotFoundException;
 import android.os.Build;
+import android.os.Bundle;
 import android.util.Log;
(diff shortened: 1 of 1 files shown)


pranitsawantAtBwellthy:
is this really fixed, text is not getting enlarge in version v5.2.1

patedit:
@pranilkanderi Can you create a separate ticket and explain your issue? This ticket was solved a long ago and maybe there is something else the SDK is missing. An screenshot would be very helpful. Thanks!

PR #158, opened 2015-01-07
New Notification Features in 4.5.2

joeatwork:
Changes include
- Motivating change: When building notifications, look for a set of magic keys in addition to "mp_message" that allow for choosing specific title, notification bar icon, and cta url in the notification
- New tests for GCMReceiver
- Minor refactor of system for reading resources by name
- Adding a collection of "default" icons to library for use as notification icons


mixpanelsteve:
This is exciting. Looking forward to playing with this!

joeatwork:
@mixpanelsteve please take a look and let me know what you think!

src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
```diff
@@ -96,6 +101,118 @@ public void onReceive(Context context, Intent intent) {
...
+        if (notificationIcon == -1) {
+            notificationIcon = android.R.drawable.sym_def_app_icon;
+        }
+
+        if (null == notificationTitle && null != appInfo) {
+            notificationTitle = manager.getApplicationLabel(appInfo);
+        }
+
+        if (null == notificationTitle) {
+            notificationTitle = "A Message For You";
```
jbwyme: In standard Mixpanel form we should probably not Title Case this (even though it's technically a title)

src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
```diff
@@ -96,6 +101,118 @@ public void onReceive(Context context, Intent intent) {
...
+        if (MPConfig.DEBUG) Log.d(LOGTAG, "MP GCM notification received: " + notificationData.message);
+        final PendingIntent contentIntent = PendingIntent.getActivity(
+                context,
+                0,
+                notificationData.intent,
+                PendingIntent.FLAG_UPDATE_CURRENT
+        );
+
+        final Notification notification;
+        if (Build.VERSION.SDK_INT >= 16) {
```
jbwyme: i hate android

jbwyme:
This looks great. Will adding this data to our message payload have implications on max message lengths?

joeatwork:
As things stand, these properties *will* affect message payload size, particularly if they're sent by adding custom data in the "custom data" box in the UI. In the short term I've tried to work around this with short-ish names (our data names live in the same namespace as customer-provided custom data, so I've kept the "mp_" prefixes on the properties, but tried to keep the names shortish (thus "mp_icnm" rather than "mp_icon_resource_name").

In the medium term, we could look into modifying the server side and UI to send these properties when pushing to android devices - the byte limit is an iOS issue (The max GCM message size is documented at 4096 bytes, and folklore says we can actually send much larger messages) and iOS devices won't be interested in the values of these properties.

PR #159, opened 2015-01-09
WIP handle failures in the API, or completely insane floods of events

joeatwork:
Deal with goofy/extreme circumstances, don't hammer endpoints when they're down.


joeatwork:
Holding off on this change

PR #168, opened 2015-02-05
Localization of alert dialog for taking survey

hypeastrum:
Hardcoded strings extracted to strings.xml, russian translation added


seanpont:
Looks great. This will be merged into 4.5.4 and pushed out with the next release.

PR #170, opened 2015-02-09
alooma-mixpanel-android SDK

ashacked:
Driver is a basic android app to test alooma-mixpanel-android SDK.
Driver is located under  alooma-mixpanel-android  directory


Driver/app/src/main/java/com/alooma/driver/MainActivity.java
```diff
@@ -0,0 +1,89 @@
...
+
+import android.content.Context;
+import com.mixpanel.android.mpmetrics.MixpanelAPI;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class MainActivity extends ActionBarActivity {
+
+    public static final String MIXPANEL_TOKEN = "d36769c7f898550cbcadc7b659bb9647";
```
joeatwork: We probably shouldn't merge your token into a public github repository. You may want to refresh your token now.

joeatwork:
Thanks for the Pull request! We're interested in application level testing, and this is a nice demonstration for us!

Unfortunately, I can't merge integrations with non-mixpanel services into our client without some higher level discussions- you may want to reach out to support for further discussion. We'd also consider pull requests that allows for more abstract integrations with user defined services (you can see the Consumer mechanism in our lightweight libraries as a possible example of an integration point we might consider)

Thanks again!

PR #173, opened 2015-02-23
Observe retry after

yinfeiru:


src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -6,6 +6,7 @@
 import android.os.HandlerThread;
 import android.os.Looper;
 import android.os.Message;
+import android.os.SystemClock;
```
joeatwork: This is better than how we do things elsewhere- we use currentTimeMillis() inappropriately in other code! Terrific for finding this-I'm going to file a new bug to fix it elsewhere!

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
```diff
@@ -102,6 +102,13 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
             db.execSQL(PEOPLE_TIME_INDEX);
         }
 
+        public boolean belowMemThreshold() {
```
joeatwork: I like this logic, but I wish that the MPDbAdapter wasn't responsible for knowing how much is too much? Perhaps it would be better to provide something like MPDbAdapter.getUsableSpace and MPDbAdapter.fileSize, and then have the comparison logic in AnalyticsMessages?

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -233,7 +235,11 @@ public void handleMessage(Message msg) {
                         logAboutMessageToMixpanel("Queuing people record for sending later");
                         logAboutMessageToMixpanel("    " + message.toString());
 
-                        queueDepth = mDbAdapter.addJSON(message, MPDbAdapter.Table.PEOPLE);
+                        if (mDbAdapter.belowMemThreshold()) {
+                            queueDepth = mDbAdapter.addJSON(message, MPDbAdapter.Table.PEOPLE);
+                        } else {
+                            Log.e(LOGTAG, "Dropping events queries due to short of memory");
```
joeatwork: Since these error messages will be visible to the user, we might want to be super specific- perhaps something like
```
"There is not enough space left on the device to store Mixpanel data, so a people analytics update was not sent"
```

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -241,7 +247,11 @@ else if (msg.what == ENQUEUE_EVENTS) {
                             final JSONObject message = prepareEventObject(eventDescription);
                             logAboutMessageToMixpanel("Queuing event for sending later");
                             logAboutMessageToMixpanel("    " + message.toString());
-                            queueDepth = mDbAdapter.addJSON(message, MPDbAdapter.Table.EVENTS);
+                            if (mDbAdapter.belowMemThreshold()) {
+                                queueDepth = mDbAdapter.addJSON(message, MPDbAdapter.Table.EVENTS);
+                            } else {
+                                Log.e(LOGTAG, "Dropping events queries due to short of memory");
```
joeatwork: Maybe an error message like this?
```
"There is not enough space left on the device to store Mixpanel data, so an event was not sent"
```

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -249,14 +259,26 @@ else if (msg.what == ENQUEUE_EVENTS) {
                     else if (msg.what == FLUSH_QUEUE) {
                         logAboutMessageToMixpanel("Flushing queue due to scheduled or forced flush");
                         updateFlushFrequency();
-                        sendAllData(mDbAdapter);
-                        mDecideChecker.runDecideChecks(getPoster());
+                        if (SystemClock.uptimeMillis() >= mRetryAfter) {
```
joeatwork: I think this should be `SystemClock.elapsedRealtime` rather than `uptimeMillis`, since we want to send even if we were asleep when the time passed.

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -249,14 +259,26 @@ else if (msg.what == ENQUEUE_EVENTS) {
...
+                            try {
+                                sendAllData(mDbAdapter);
+                                mDecideChecker.runDecideChecks(getPoster());
+                            } catch (ServiceUnavailableException e) {
+                                mRetryAfter = SystemClock.uptimeMillis() + e.getRetryAfter() * 1000;
```
joeatwork: elapsedRealtime

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -564,14 +590,15 @@ private void updateFlushFrequency() {
...
-    private static int REGISTER_FOR_GCM = 13; // Register for GCM using Google Play Services
+    private static final int ENQUEUE_PEOPLE = 0; // submit events and people data
+    private static final int ENQUEUE_EVENTS = 1; // push given JSON message to people DB
+    private static final int FLUSH_QUEUE = 2; // push given JSON message to events DB
+    private static final int KILL_WORKER = 5; // Hard-kill the worker thread, discarding all events on the event queue. This is for testing, or disasters.
+    private static final int INSTALL_DECIDE_CHECK = 12; // Run this DecideCheck at intervals until it isDestroyed()
+    private static final int REGISTER_FOR_GCM = 13; // Register for GCM using Google Play Services
 
     private static final String LOGTAG = "MixpanelAPI.AnalyticsMessages";
+    private static long mRetryAfter = -1;
```
joeatwork: I'd prefer it if mRetryAfter was an instance variable, and was only visible to the Handler - because this class is multithreaded, should limit visibility of mutable values to the innermost class (particularly we want to be 100% sure only one thread ever consults this value.)

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -564,14 +590,15 @@ private void updateFlushFrequency() {
...
-    private static int FLUSH_QUEUE = 2; // push given JSON message to events DB
-    private static int KILL_WORKER = 5; // Hard-kill the worker thread, discarding all events on the event queue. This is for testing, or disasters.
-    private static int INSTALL_DECIDE_CHECK = 12; // Run this DecideCheck at intervals until it isDestroyed()
-    private static int REGISTER_FOR_GCM = 13; // Register for GCM using Google Play Services
+    private static final int ENQUEUE_PEOPLE = 0; // submit events and people data
+    private static final int ENQUEUE_EVENTS = 1; // push given JSON message to people DB
+    private static final int FLUSH_QUEUE = 2; // push given JSON message to events DB
+    private static final int KILL_WORKER = 5; // Hard-kill the worker thread, discarding all events on the event queue. This is for testing, or disasters.
+    private static final int INSTALL_DECIDE_CHECK = 12; // Run this DecideCheck at intervals until it isDestroyed()
+    private static final int REGISTER_FOR_GCM = 13; // Register for GCM using Google Play Services
```
joeatwork: These finals are GREAT! You fixed another bug!

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -249,14 +259,26 @@ else if (msg.what == ENQUEUE_EVENTS) {
...
+                                mRetryAfter = SystemClock.uptimeMillis() + e.getRetryAfter() * 1000;
+                            }
+                        }
                     }
                     else if (msg.what == INSTALL_DECIDE_CHECK) {
                         logAboutMessageToMixpanel("Installing a check for surveys and in app notifications");
                         final DecideMessages check = (DecideMessages) msg.obj;
                         mDecideChecker.addDecideCheck(check);
-                        mDecideChecker.runDecideChecks(getPoster());
+                        if (SystemClock.uptimeMillis() >= mRetryAfter) {
```
joeatwork: elapsedRealtime

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -275,11 +297,15 @@ else if (msg.what == KILL_WORKER) {
 
                     ///////////////////////////
 
-                    if (queueDepth >= mConfig.getBulkUploadLimit()) {
+                    if (queueDepth >= mConfig.getBulkUploadLimit() && SystemClock.uptimeMillis() >= mRetryAfter) {
```
joeatwork: There is an interesting and strange edge case- what if we have only a few records, but they're very, very large? I'd like to try and push here under the following circumstances
```
if ( (queueDepth >= mConfig.getBulkUploadLimit() || we are dropping events due to lack of space) &&
    SystemClock.elapsedRealtime() >= mRetryAfter) {
...
}
```

src/main/java/com/mixpanel/android/mpmetrics/ServerMessage.java
```diff
@@ -148,3 +155,20 @@ public boolean isOnline(Context context) {
...
+        } catch (NumberFormatException e) {
+            mRetryAfter = 0;
+        }
+    }
+
+    public int getRetryAfter() {
+        return mRetryAfter;
+    }
+
+    private int mRetryAfter;
```
joeatwork: This can be final, since it will never change

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
```diff
@@ -102,6 +102,13 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
...
+            if (mDatabaseFile.exists()) {
+                return Math.min(mDatabaseFile.getUsableSpace(), 20 * 1024 * 1024) >= mDatabaseFile.length();
```
joeatwork: It might be nice to put this constant (the minimum) in MPConfig - thats where configuration values can live, and it means you can give it a nice human readable name (and allow library users to adjust it)

joeatwork:
This is *great*!

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
yinfeiru: hmmm.. I don't get the point of not doing it in the MPDbAdapter. I think - 
1. mDatabaseFile is a private member of MPDatabaseHelper. so MPDatabaseHelper should manage the size of this guy.
2. to me, all the data base related stuff should be encapsulated in the MPDatabaseHelper and MPDbAdapter classes. and when we do the memory check, we don't care about the type/content/count or any specific attribute of the message, the only thing we care about is how much space we are using. so I don't think AnalyticsMessages should be involved.
3. AnalyticsMessages shouldn't know the logic of how the database is storing the messages. in the current code, all it does is insert new messages into the database by calling an API, without knowing the details of the database.

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: yes!

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: this is a good point!! we care only about the real time passed.

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: 54cfcf2

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: 54cfcf2

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: 47e116f

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: 47e116f

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: 47e116f

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: c17b1cb

yinfeiru: (if we don't need to change the implementation of belowMemThreshold)

src/main/java/com/mixpanel/android/mpmetrics/ServerMessage.java
yinfeiru: I actually made it a final in the very beginning.. but the java syntax checker says "Variable might already have been assigned to" at line 165.

joeatwork: I see. Perhaps this is the way around that issue:
```
public ServiceUnavailableException(String message, String strRetryAfter) {
    super(message);
    int retry;
    try {
        retry = Integer.parseInt(strRetryAfter);
    } catch (NumberFormatException e) {
        retry = 0;
    }
    mRetryAfter = retry;
}
```

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
joeatwork: In that case, perhaps we should put all of the "database full" logic in MPDbAdapter? Perhaps by simply ignoring the write (and logging an error) in addJSON if the memory is too full? That would mean that MPDbAdapter was responsible for not only the "How big is too big" part of the policy, but also the "what to do if you're too big" part of the policy.

src/main/java/com/mixpanel/android/mpmetrics/ServerMessage.java
yinfeiru: nice!

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
yinfeiru: 4623e17

src/main/java/com/mixpanel/android/mpmetrics/ServerMessage.java
yinfeiru: 1f5a72c

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
yinfeiru: 9cffc4c c07e210 1f5a72c 0612317

sorry... they are in different branches dues to the weird failure of the test...

yinfeiru: 9fcbea8

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
```diff
@@ -39,12 +39,15 @@ public String getName() {
         private final String mTableName;
     }
 
-    private static final String DATABASE_NAME = "mixpanel";
-    private static final int DATABASE_VERSION = 4;
-
     public static final String KEY_DATA = "data";
     public static final String KEY_CREATED_AT = "created_at";
 
+    public static final int DBUpdateError = -1;
```
joeatwork: Please name finals known at compile time in ALL_CAPS

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -225,7 +228,7 @@ public void handleMessage(Message msg) {
                 }
 
                 try {
-                    int queueDepth = -1;
+                    int returnCode = MPDbAdapter.DBUpdateError;
```
joeatwork: We might want to add a code for "Nothing happened" or "undefined" and use it for the default value rather than defaulting to UPDATE_ERROR

src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
```diff
@@ -56,6 +56,7 @@ public static MPConfig getInstance(Context context) {
         mBulkUploadLimit = metaData.getInt("com.mixpanel.android.MPConfig.BulkUploadLimit", 40); // 40 records default
         mFlushInterval = metaData.getInt("com.mixpanel.android.MPConfig.FlushInterval", 60 * 1000); // one minute default
         mDataExpiration = metaData.getInt("com.mixpanel.android.MPConfig.DataExpiration",  1000 * 60 * 60 * 24 * 5); // 5 days default
+        mDBMaxSize = metaData.getInt("com.mixpanel.android.MPConfig.DBMaxSize",  20 * 1024 * 1024); // 20 Mb
```
joeatwork: This name isn't really the Max size (we'll allow it to take up half the available storage) - I'm not sure what the name should be, maybe "MinimumDatabaseLimit"? Then if the user sets it to zero, we will demand that there is at least half the available space left.

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
```diff
@@ -121,10 +133,16 @@ public MPDbAdapter(Context context, String dbName) {
      * @return the number of rows in the table, or -1 on failure
      */
     public int addJSON(JSONObject j, Table table) {
+        // we are aware of the race condition here, but what can we do..?
+        if (!this.belowMemThreshold()) {
```
joeatwork: Great! I love that this logic is in here - this is a really nice interface!

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -275,19 +290,24 @@ else if (msg.what == KILL_WORKER) {
 
                     ///////////////////////////
 
-                    if (queueDepth >= mConfig.getBulkUploadLimit()) {
+                    if ((returnCode >= mConfig.getBulkUploadLimit() || returnCode == MPDbAdapter.DBOutOfMemoryError) &&
+                            SystemClock.elapsedRealtime() >= mRetryAfter) {
```
joeatwork: I'd like to see a test for the back-off behavior- provide a mock poster to AnalyticsMessages (that throws a ServiceUnavailableException) and somehow see that the AnalyticsMessages actually back off. I'd also like to be sure it behaves like we expect in the presence of error codes from a mock DB.

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
yinfeiru: f8388a3

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: good point  473be86

src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
yinfeiru: fbed0c5

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: As we discussed, these test cases are already covered in HttpTest (503 test and short of memory test).

joeatwork:
Merged into v4.5.4dev!

PR #177, opened 2015-03-05
OnMixpanelUpdatesReceivedListener should be public to use it outside of ...

aguillard:
...the sdk


joeatwork:
Thanks for the suggestion! We'll need to think a bit about this API before we push it public- We may change the communication model for updates in the future, and listeners might end up behaving in a strange way.

I'd love to hear a bit more about your use case, though! We're in the midst of a new release of the library (You can see what we're doing in the v4.5.4dev branch) and this might be a good time for us to rethink how advanced users can hook into Mixpanel updates. Are you in a situation where getSurveyIfAvailable and getNotificationIfAvailable aren't working for what you need?

joeatwork:
I am completely wrong about this. Merging presently.

joeatwork:
This has been merged into branch v4.5.4dev for inclusion in the next release. Sorry about my clueless response (and thanks for your help!)

PR #179, opened 2015-03-06
Fix worker crash when fallabck endpoint disabled

baolongnt:


joeatwork:
Thanks! This is already cooking for the next release, and will be included in 4.5.4. In the mean time the workaround is to disable debug output in your app by removing the com.mixpanel.android.MPConfig.EnableDebugLogging directive from your AndroidManifest.xml

joeatwork:
(Also, thank you for sending this fix! We really appreciate the help!)

baolongnt:
No pb. When do you think 4.5.4 will be out? Right now, that debug feature is not usable.

joeatwork:
We're working on it now (you can actually watch our progress on branch v4.5.4dev ( https://github.com/mixpanel/mixpanel-android/tree/v4.5.4dev )!

If debug output is helpful to you, can also work around the issue by turning on HTTP fallback if you need the debug output during development- while you probably shouldn't ship with fallback on unless your app is targeting particular markets or unique circumstances, the same is true for shipping with debug output.

baolongnt:
Thanks for the workaround.

PR #182, opened 2015-03-10
send default color as an integer

yinfeiru:


joeatwork:
Looks great! Go ahead and merge into v4.5.4dev!

PR #190, opened 2015-03-23
Tweaks processor for 4.5.4 release

joeatwork:
- Hooks for `@Tweaks` annotation processor, including discovery of `$$TWEAK_REGISTRAR` classes and `registerForTweaks` for instances.
- Removing defaults from individual Tweak accessors
- Drive-by fixes and cleanup

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -118,14 +118,9 @@ uploadArchives {
             }
             developers {
                 developer {
-                    id 'jbwyme'
-                    name 'Josh Wymer'
-                    email 'josh@mixpanel.com'
-                }
-                developer {
-                    id 'joeatwork'
-                    name 'Joe Bowers'
-                    email 'joe@mixpanel.com'
+                    id 'mixpanel_dev'
+                    name 'Mixpanel Developers'
+                    email 'dev@mixpanel.com'
                 }
             }
         }
--- a/build.gradle
+++ b/build.gradle
@@ -120,7 +120,7 @@ uploadArchives {
                 developer {
                     id 'mixpanel_dev'
                     name 'Mixpanel Developers'
-                    email 'dev@mixpanel.com'
+                    email 'dev+android@mixpanel.com'
                 }
             }
         }
--- a/build.gradle
+++ b/build.gradle
@@ -7,13 +7,13 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.0'
+        classpath 'com.android.tools.build:gradle:1.1.0'
     }
 }
 
 android {
     compileSdkVersion 21
-    buildToolsVersion "21.1.1"
+    buildToolsVersion "21.1.2"
 
     defaultConfig {
         // If you change minSdkVersion or targetSdkVersion below, you must also change
--- a/src/androidTest/java/com/mixpanel/android/mpmetrics/TweaksTest.java
+++ b/src/androidTest/java/com/mixpanel/android/mpmetrics/TweaksTest.java
@@ -1,0 +1,85 @@
+package com.mixpanel.android.mpmetrics;
+
+import android.test.AndroidTestCase;
+import android.util.Pair;
+
+import com.mixpanel.android.mpmetrics.test_autotweak_package.RegistrarRuns;
+import com.mixpanel.android.mpmetrics.test_autotweak_package.TweakableThing;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class TweaksTest extends AndroidTestCase {
+    @Override
+    public void setUp() {
+        mTweaks = new Tweaks(new TestUtils.SynchronousHandler(), "RegistrarRuns");
+    }
+
+    public void testNoSuchTweakPoll() {
+        // Should return null for tweaks we know nothing about
+        assertNull(mTweaks.getString("Some Tweak"));
+    }
+
+    public void testTweakWithDefault() {
+        mTweaks.bind("Some Tweak", "Default Value", null);
+        assertEquals("Default Value", mTweaks.getString("Some Tweak"));
+    }
+
+    public void testNonStringTweakWithWrongType() {
+        mTweaks.bind("Some Tweak", 100.0, null);
+        assertNull(mTweaks.getString("Some Tweak"));
+    }
+
+    public void testPrimitiveTweakWithWrongType() {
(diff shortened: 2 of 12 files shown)


src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -109,12 +145,39 @@ public void set(Map<String, Object> tweakUpdates) {
...
+                    final ClassLoader loader = klass.getClassLoader();
+                    final Package registrantPackage = klass.getPackage();
+
+                    try {
+                        final Class found = loader.loadClass(registrantPackage.getName() + "." + mTweakClassName);
+                        final Field instanceField = found.getField("TWEAK_REGISTRAR");
+                        final TweakRegistrar registrar = (TweakRegistrar) instanceField.get(null);
+                        sRegistrars.put(klass, registrar);
+                    } catch (ClassNotFoundException e) {
+                        ; // Ok, no such class.
```
yinfeiru: why don't we log it?

joeatwork: This will happen more often than not, and it isn't an error condition- we just don't have a `ClassLoader.classExists` method to call.

joeatwork: The rest of the exceptions are much more likely to be errors- if there is a class named `$$TWEAK_REGISTRAR` but it doesn't conform to our assumptions, a programmer (us or the customer) has likely made an error.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -906,18 +908,20 @@ public JSONObject update(JSONObject in) {
...
-    private static final int MESSAGE_HANDLE_EDITOR_BINDINGS_RECEIVED = 8;
-    private static final int MESSAGE_SEND_EVENT_TRACKED = 9;
-    private static final int MESSAGE_HANDLE_EDITOR_CLOSED = 10;
-    private static final int MESSAGE_VARIANTS_RECEIVED = 11;
+    private static final @MessageType int MESSAGE_INITIALIZE_CHANGES = 0;
+    private static final @MessageType int MESSAGE_CONNECT_TO_EDITOR = 1;
+    private static final @MessageType int MESSAGE_SEND_STATE_FOR_EDITING = 2;
+    private static final @MessageType int MESSAGE_HANDLE_EDITOR_CHANGES_RECEIVED = 3;
+    private static final @MessageType int MESSAGE_SEND_DEVICE_INFO = 4;
+    private static final @MessageType int MESSAGE_EVENT_BINDINGS_RECEIVED = 6;
```
yinfeiru: I know it doesn't matter and I'm being picky, but why do we keep 5 and 7 here?

build.gradle
```diff
@@ -7,13 +7,13 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.0'
+        classpath 'com.android.tools.build:gradle:1.1.0'
```
yinfeiru: do we have a doc for tracking which gradle/buildToolsVersion/etc. the library is on?

build.gradle
```diff
@@ -118,14 +118,9 @@ uploadArchives {
...
-                    name 'Josh Wymer'
-                    email 'josh@mixpanel.com'
-                }
-                developer {
-                    id 'joeatwork'
-                    name 'Joe Bowers'
-                    email 'joe@mixpanel.com'
+                    id 'mixpanel_dev'
+                    name 'Mixpanel Developers'
+                    email 'dev@mixpanel.com'
```
yinfeiru: do we want to have an android_dev mailing list instead of spamming all the devs?

src/androidTest/java/com/mixpanel/android/mpmetrics/TweaksTest.java
```diff
@@ -0,0 +1,85 @@
...
+        TweakableThing thing = new TweakableThing();
+        mTweaks.registerForTweaks(thing);
+        assertEquals(1, RegistrarRuns.TWEAK_REGISTRAR.wasRegistered.size());
+        Pair<Tweaks, Object> found = RegistrarRuns.TWEAK_REGISTRAR.wasRegistered.get(0);
+        assertEquals(mTweaks, found.first);
+        assertEquals(thing, found.second);
+    }
+
+    Tweaks mTweaks;
+}
```
yinfeiru: I don't see any test case for "public synchronized void bind(String tweakName, Object defaultValue, Object gcScope, TweakChangeCallback callback)". are you skipping it intentionally and why?

build.gradle
joeatwork: Unfortunately, no- I've been upgrading it every time that Android Studio tells me to, rolling to the latest version. The good news is that this shouldn't matter to customers (they don't have to upgrade their tools to use the library)

build.gradle
joeatwork: That sounds like a great idea! I'll ask Josh to set it up!

joeatwork: Actually, I can just change this to `dev+android@mixpanel.com` and folks can filter it appropriately

src/androidTest/java/com/mixpanel/android/mpmetrics/TweaksTest.java
joeatwork: I'm skipping this because it's really tough to test WeakReference behavior- The gcScope argument is there to prevent leaks, and it's hard to stimulate garbage collection on a particular object.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: I think you're right! I've been reluctant to change these constants (and there were some constants that have gone away) but these should *never* be persisted. I'll make that change!

build.gradle
joeatwork: Fixed 2d64c66

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: Fixed in b537fae

yinfeiru:
looks good to me

PR #191, opened 2015-03-25
add all required files

yinfeiru:


acceptance/run.sh
```diff
@@ -0,0 +1,123 @@
...
+function npm_package_check {
+	for package in $NPM_PACKAGES_REQUIRED
+	do
+		ls node_modules 2> /dev/null | grep $package > /dev/null
+		if [ $? != 0 ]
+		then
+			echo "$package is missing. installing.."
+			if [ $VERBOSE != 0 ]
+			then
+				sudo npm install $package --save
```
joeatwork: Let's not put sudo into these scripts- I believe we should be able to install npm packages locally.

acceptance/README.txt
```diff
@@ -0,0 +1,17 @@
+1.  install selenium webdriver lib by "sudo easy_install selenium"
```
joeatwork: We could leave out the sudo here, we should have virtualenv and pip installed.

acceptance/run.sh
```diff
@@ -0,0 +1,123 @@
...
+	NODE_SERVER_PID=$!
+}
+
+function start_selendroid_server {
+	export ANDROID_HOME=~/Library/Android/sdk/
+	if [ $VERBOSE != 0 ]
+	then
+		java -jar selendroid-standalone-0.15.0-with-dependencies.jar -app sample-android-mixpanel-integration/build/outputs/apk/sample-android-mixpanel-integration-debug.apk&
+	else
+		java -jar selendroid-standalone-0.15.0-with-dependencies.jar -app sample-android-mixpanel-integration/build/outputs/apk/sample-android-mixpanel-integration-debug.apk > /dev/null 2> /dev/null&
```
joeatwork: Let's always do this in verbose mode for now.

acceptance/test.py
```diff
@@ -0,0 +1,25 @@
+import unittest
+from selenium import webdriver
+
+class FindElementTest(unittest.TestCase):
```
joeatwork: Three languages (shell, node, and python) is a lot for these things.

acceptance/run.sh
joeatwork: Let's not install packages in the script. If we need packages, we should fail and let users install them (but we should minimize the number of packages and external software that we need)

joeatwork:
As discussed, here's my big list of suggestions :)

- Let's avoid sudo. We can expect users to have virtualenv installed. In particular, let's not use sudo inside of shell scripts.
- Python, shell, and node is a lot of different languages. I'd bet there is a quick way to serve content with python, for example http://effbot.org/librarybook/simplehttpserver.htm
- Let's add the app under test to the repository rather than download and edit it.
- Let's use the AndroidManifest.xml for mDecideEndpoint, mDecideFallbackEndpoint and mEditorUrl instead of editing code- these are manifest configuration directives
- If we need to build, I'd prefer to use ./gradlew rather than open Android Studio
- Why do we need a good MIXPANEL_API_TOKEN? Can we just leave this with a dummy value?

acceptance/run.sh
```diff
@@ -0,0 +1,123 @@
...
+			then
+				sudo npm install $package --save
+			else
+				sudo npm install $package --save > /dev/null
+			fi
+		fi
+	done
+}
+
+function parse_config {
```
joeatwork: Let's see if we can get away with a dummy value here, so you can eliminate the config file and editing Java with sed

acceptance/run.sh
```diff
@@ -0,0 +1,123 @@
...
+			'MIXPANEL_API_TOKEN')
+				sed -i -e "s/MIXPANEL_API_TOKEN = \"YOUR API TOKEN\"/MIXPANEL_API_TOKEN = \"${value}\"/g" $SAMPLE_APP_DIR/src/com/mixpanel/example/hello/MainActivity.java
+			;;
+		esac
+	done < $CONFIG_FILE
+}
+
+function build_apk {
+	pushd ${SAMPLE_APP_DIR} > /dev/null
+	if [ $VERBOSE != 0 ]
```
joeatwork: Let's not have switches- let's either always or never run it as verbose, and let users use grep if they need to see special stuff in the output

acceptance/run.sh
```diff
@@ -0,0 +1,123 @@
...
+	done
+
+	echo "Total test suite(s): ${total_cnt}"
+	echo "Failed test suite(s): ${failure_cnt}"
+}
+
+function clean_up {
+	if [ $NODE_SERVER_PID != NULL ]
+	then
+		kill -9 $NODE_SERVER_PID
```
joeatwork: Let's not kill -9 in a script- bugs could make this do crazy stuff. Can we just kill it with SIGTERM?

acceptance/run.sh
```diff
@@ -0,0 +1,123 @@
...
+	fi
+	
+	if [ $SELEDROID_PID != NULL ]
+	then
+		kill -9 $SELEDROID_PID
+	fi
+}
+
+echo "looking for required module - express.."
```
joeatwork: We probably don't need this output

acceptance/server.js
```diff
@@ -0,0 +1,12 @@
+var express = require("express");
```
joeatwork: I'd prefer to see this in python, since we'll be using python for the tests.

acceptance/test.py
```diff
@@ -0,0 +1,25 @@
...
+        self.driver.quit()
+
+    def test_change_text_size(self):
+        text_size_change_response = '{"variants":[{"experiment_id":3585,"id":7647,"tweaks":[],"actions":[{"path":[{"id":16908290,"index":0,"prefix":"shortest"},{"view_class":"android.widget.LinearLayout","index":0},{"index":0,"mp_id_name":"edit_first_name"}],"args":[[4289362242,"java.lang.Integer"]],"property":{"get":{"selector":"getTextColors","result":{"type":"android.content.res.ColorStateList"},"parameters":[]},"set":{"selector":"setTextColor","parameters":[{"type":"java.lang.Integer"}]},"name":"textColor","classname":"android.widget.TextView"},"name":"c117"},{"path":[{"id":16908290,"index":0,"prefix":"shortest"},{"view_class":"android.widget.LinearLayout","index":0},{"index":0,"mp_id_name":"edit_first_name"}],"args":[[2,"java.lang.Integer"],[36,"java.lang.Float"]],"property":{"get":{"selector":"getTextSize","result":{"type":"java.lang.Float"},"parameters":[]},"set":{"selector":"setTextSize","parameters":[{"type":"java.lang.Integer"},{"type":"java.lang.Float"}]},"name":"fontSize","classname":"android.widget.TextView"},"name":"c131"}]}],"surveys":[],"notifications":[]}'
```
joeatwork: How do we know when this test fails?

yinfeiru: if the response is wrong, the text size won't be updated so that the following assert will fail.

yinfeiru:
I've added the sample app into this rep. To review the changes to it, it'd be better to checkout this branch, go to sample-android-mixpanel-integration/ and do a git diff (instead of reading the whole directory).

For the Android library location, should I leave
```
compile "com.mixpanel.android:mixpanel-android:4.5.4-SNAPSHOT@aar"
```
there or put a fake path there and leave a comment in README asking the user to update it manually?

acceptance/run.sh
yinfeiru: 1cffaa3

I'll remove the npm server in a later branch

acceptance/README.txt
yinfeiru: a05f2c9

acceptance/run.sh
yinfeiru: a4191a2

sorry, should have a separate branch for it

acceptance/run.sh
yinfeiru: a4191a2
yes it works

acceptance/run.sh
yinfeiru: removed a4191a2

acceptance/run.sh
yinfeiru: 326f69a

acceptance/run.sh
yinfeiru: a4191a2

acceptance/server.js
yinfeiru: 376a6a6

joeatwork:
Let's put a fake, non-working path in the andr

acceptance/sample-android-mixpanel-integration/README.md
```diff
@@ -0,0 +1,116 @@
+# Sample Android Application for Mixpanel Integration
```
joeatwork: Let's remove the readme :)

acceptance/sample-android-mixpanel-integration/build.gradle
```diff
@@ -0,0 +1,49 @@
+buildscript {
```
joeatwork: Let's change the name of the sample app to "test-application" or something

acceptance/sample-android-mixpanel-integration/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+apply plugin: 'android'
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+}
+
+dependencies {
+    // for some reason, the signed pom specifies a package type of aar.asc, adding @aar resolves the correct file
+    compile "com.mixpanel.android:mixpanel-android:4.5.4-SNAPSHOT@aar"
```
joeatwork: As discussed, let's put a non-working value for version here.

acceptance/sample-android-mixpanel-integration/README.md
yinfeiru: acb6d44

acceptance/sample-android-mixpanel-integration/build.gradle
yinfeiru: acb6d44 40c6a20

Issue #176, opened 2015-03-02
Logtags too long


joeatwork:
Log tags have a length limit of 23 characters in some parts of the Android API, but the log tags in the library are longer than that.

PR #193, opened 2015-03-26
Basic undo. We'll need to revisit text size for this to work

joeatwork:
We can (and now do) round-trip values if they can be cleanly read from and set on views as properties. We'll need to work on marshaling text size and some other properties but it works for text values of text areas and opacity.

### DIFF
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
@@ -283,5 +283,5 @@ import java.util.List;
     private static final List<Pathfinder.PathElement> NEVER_MATCH_PATH = Collections.<Pathfinder.PathElement>emptyList();
 
     @SuppressWarnings("unused")
-    private static final String LOGTAG = "MixpanelAPI.EditProtocol";
+    private static final String LOGTAG = "MixpanelAPI.EProtocol";
 } // EditProtocol
--- a/src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
@@ -264,7 +264,7 @@ public class ViewCrawler implements UpdatesFromMixpanel, TrackingDebug {
             mToken = token;
             mSnapshot = null;
 
-            mEditorChanges = new ArrayList<Pair<String, JSONObject>>();
+            mEditorChanges = new HashMap<String, Pair<String, JSONObject>>();
             mEditorEventBindings = new ArrayList<Pair<String, JSONObject>>();
             mPersistentChanges = new ArrayList<VariantChange>();
             mPersistentEventBindings = new ArrayList<Pair<String, JSONObject>>();
@@ -311,6 +311,9 @@ public class ViewCrawler implements UpdatesFromMixpanel, TrackingDebug {
                         break;
                     case MESSAGE_HANDLE_EDITOR_BINDINGS_RECEIVED:
                         handleEditorBindingsReceived((JSONObject) msg.obj);
+                        break;
+                    case MESSAGE_HANDLE_EDITOR_CHANGES_CLEARED:
+                        handleEditorBindingsCleared((JSONObject) msg.obj);
                         break;
                     case MESSAGE_HANDLE_EDITOR_CLOSED:
                         handleEditorClosed();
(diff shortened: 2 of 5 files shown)


src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -615,6 +619,24 @@ private void handleEditorChangeReceived(JSONObject changeMessage) {
...
         /**
+         * Remove a change from the connected UI.
+         */
+        private void handleEditorBindingsCleared(JSONObject clearMessage) {
+            try {
+                final JSONObject payload = clearMessage.getJSONObject("payload");
+                final JSONArray actions = payload.getJSONArray("actions");
+                for (int i = 0; i < actions.length(); i++) {
+                    final String changeId = actions.getString(i);
+                    mEditorChanges.remove(changeId);
```
yinfeiru: I'm not that familiar with JAVA ArrayList, but what if changeId doesn't exist?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -713,9 +733,7 @@ private void updateEditState() {
             }
 
             {
-                final int size = mEditorChanges.size();
-                for (int i = 0; i < size; i++) {
-                    final Pair<String, JSONObject> changeInfo = mEditorChanges.get(i);
+                for (Pair<String, JSONObject> changeInfo:mEditorChanges.values()) {
```
yinfeiru: this is nice..! I didn't know we can do this in Java.. it's python-like..

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -787,7 +805,7 @@ private SharedPreferences getSharedPreferences() {
         private final String mToken;
         private final Lock mStartLock;
 
-        private final List<Pair<String,JSONObject>> mEditorChanges;
+        private final Map<String, Pair<String,JSONObject>> mEditorChanges;
```
yinfeiru: hmmm.. I'm a bit confused.. mEditorChanges is declared as a Map but initialized as a HashMap. is this intentional?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
garycavasas: Take care of it

-----Original Message-----

joeatwork: The good news is this isn't an ArrayList, or a list anymore - it's a Map of ids to edits. If the id isn't present, remove returns null and doesn't change the underlying map.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: Yes. We always try to use the abstract interfaces rather than their concrete types. We don't really care if the Map we're given is based in a HashTable or a Tree or even something more exotic- all users need to know is that it's a map.

We try to do this wherever we can- only refer to types by their simplest usable superclass - but it's particularly idiomatic for collections.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -81,6 +101,8 @@ protected String name() {
 
         private final Caller mMutator;
         private final Caller mAccessor;
+        private final WeakHashMap<View, Object> mOriginalValues;
```
yinfeiru: Should we change this guy to Map, as you said "We always try to use the abstract interfaces "?

yinfeiru:
If I understand it correct, we use mEditorChanges to save all the changes. Every time the phone redraws the view, we overwrite the original view with items in mEditorChanges. And when we clear, we simply remove items from mEditorChanges, so that the view won't be overwritten. Am I correct?

joeatwork:
That's true. It means every new edit we clean up all of the old edits and apply new ones.

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -283,5 +283,5 @@ private Object convertArgument(Object jsonArgument, String type) throws BadInstr
     private static final List<Pathfinder.PathElement> NEVER_MATCH_PATH = Collections.<Pathfinder.PathElement>emptyList();
 
     @SuppressWarnings("unused")
-    private static final String LOGTAG = "MixpanelAPI.EditProtocol";
+    private static final String LOGTAG = "MixpanelAPI.EProtocol";
```
jbwyme: Why make it less descriptive? Too long?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -312,6 +312,9 @@ public void handleMessage(Message msg) {
                     case MESSAGE_HANDLE_EDITOR_BINDINGS_RECEIVED:
                         handleEditorBindingsReceived((JSONObject) msg.obj);
                         break;
+                    case MESSAGE_HANDLE_EDITOR_CHANGES_CLEARED:
```
jbwyme: So does this only support undoing all changes? What about a single change?

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
joeatwork: It turns out there is a 23 character limit on logtags we've been blowing past (I've actually filed this as #176) - since at this point all of our logtags are prefixed with "MixpanelAPI" (which is super useful!) that leaves us with a bunch of abbrevs. and short msgs.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: It's per change- the object's actions are a list of change ids that we use to undo (all changes get cleared on disconnect in a separate mechanism that doesn't require the editor to exist)

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: Unfortunately it's relatively inefficient (often it ends up allocating a bunch of new objects) so we prefer olde schoole C-style for loops when they're available.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
joeatwork: In this case `WeakHashMap` is special- the weakness of it is important or this class leaks a lot of memory as views disappear and are freed. Honestly, it'd be great if there was a `WeakMap` interface, because the Weak is something we care about but the Hash isn't.

PR #199, opened 2015-04-06
Set layout of view

yinfeiru:


src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -90,35 +89,43 @@ public ViewVisitor readEdit(JSONObject source) throws BadInstructionsException {
...
+                }
 
-            return new ViewVisitor.PropertySetVisitor(path, mutator, prop.accessor);
+                return new ViewVisitor.PropertySetVisitor(path, mutator, prop.accessor);
+            } else if (source.has("layout")) {
+                final JSONArray args = source.getJSONArray("args");
+                final LayoutCaller mutator = new LayoutCaller(args);
+                return new ViewVisitor.LayoutSetVisitor(path, mutator);
+            } else {
+                return null;
```
joeatwork: We should probably throw a BadInstructionsException if we can't figure out how the edit works.

src/main/java/com/mixpanel/android/viewcrawler/LayoutCaller.java
```diff
@@ -0,0 +1,40 @@
+package com.mixpanel.android.viewcrawler;
+
+import android.view.View;
+import android.widget.RelativeLayout;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class LayoutCaller {
```
joeatwork: We may want to remove the "Caller" prefix from this name, since it works very differently from PropertyCaller. (We can also use Caller in cases where we don't have a property, it'll work for any random method- we may want to roll back that rename)

src/main/java/com/mixpanel/android/viewcrawler/LayoutCaller.java
```diff
@@ -0,0 +1,40 @@
...
+    public LayoutCaller(JSONArray args) throws JSONException {
```
joeatwork: I'd prefer if this constructor didn't take JSON- it feels like this object could be independent of the protocol we use to create them.

joeatwork: Could this class be inline in LayoutSetVisitor?

src/main/java/com/mixpanel/android/viewcrawler/LayoutCaller.java
```diff
@@ -0,0 +1,40 @@
...
+        } else {
+            mArgs = new int[]{rule_id, RelativeLayout.TRUE};
+        }
+    }
+
+    public int[] getArgs() {
+        return mArgs;
+    }
+
+    public void applyMethod(View target) {
```
joeatwork: We probably don't need this to have general names like "applyMethod" - there isn't really a method to apply (We're just grabbing a LayoutParams, adding some rules, and then setting them again. I suspect we could do this inline in the visitor.

src/main/java/com/mixpanel/android/viewcrawler/LayoutCaller.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public void applyMethodWithArguments(View target, int[] args) {
+        RelativeLayout.LayoutParams params = (RelativeLayout.LayoutParams)target.getLayoutParams();
+        params.addRule(args[RULE_INDEX], args[ANCHOR_ID]);
+        target.setLayoutParams(params);
+    }
+
+    private final int[] mArgs;
+    public static final int RULE_INDEX = 0;
```
joeatwork: These could just be members rather than indexes into an array- you could just have two final int members.

src/main/java/com/mixpanel/android/viewcrawler/ViewSnapshot.java
```diff
@@ -232,6 +234,21 @@ private void addProperties(JsonWriter j, View v)
         }
     }
 
+    // we only do layouts for views with RelativeLayout for now
+    private void addLayouts(JsonWriter j, View v)
+        throws IOException {
+        if (v.getParent() instanceof RelativeLayout) {
```
joeatwork: We probably don't need this check- let's just get the layout params and return them, and let the web side decide which can be changed.

src/main/java/com/mixpanel/android/viewcrawler/ViewSnapshot.java
```diff
@@ -232,6 +234,21 @@ private void addProperties(JsonWriter j, View v)
...
+            RelativeLayout.LayoutParams layoutParams = (RelativeLayout.LayoutParams) (v.getLayoutParams());
```
joeatwork: Can this cast ever fail?

src/main/java/com/mixpanel/android/viewcrawler/ViewSnapshot.java
```diff
@@ -232,6 +234,21 @@ private void addProperties(JsonWriter j, View v)
         }
     }
 
+    // we only do layouts for views with RelativeLayout for now
+    private void addLayouts(JsonWriter j, View v)
```
joeatwork: This is simple enough (particularly if we provide them unconditionally) that we might want to inline it

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -99,12 +100,58 @@ protected String name() {
...
+            mOriginalValues = new WeakHashMap<View, int[]>();
+            mMutator = mutator;
+        }
+
+        @Override
+        public void cleanup() {
+            for (Map.Entry<View, int[]> original:mOriginalValues.entrySet()) {
+                final View changedView = original.getKey();
+                final int[] originalValue = original.getValue();
+                mMutator.applyMethodWithArguments(changedView, originalValue);
```
joeatwork: cleanup() should probably explicitly set layout rules rather than call something indirectly.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -99,12 +100,58 @@ protected String name() {
...
+        private final PropertyCaller mAccessor;
         private final WeakHashMap<View, Object> mOriginalValues;
         private final Object[] mOriginalValueHolder;
     }
 
+    public static class LayoutSetVisitor extends ViewVisitor {
+        public LayoutSetVisitor(List<Pathfinder.PathElement> path, LayoutCaller mutator) {
+            super(path);
+            mOriginalValues = new WeakHashMap<View, int[]>();
+            mMutator = mutator;
```
joeatwork: While the LayoutCaller is technically a mutator, we don't have to use such general naming here- we know it's a layout, and it'll never be anything else.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -99,12 +100,58 @@ protected String name() {
...
+
+        @Override
+        public void accumulate(View found) {
+            // currentRules is an array which has rule_index as the index and anchor_id as the value
+            // newRule and currentRule are individual rules which look like {rules_index, anchor_id}
+            final RelativeLayout.LayoutParams currentParams = (RelativeLayout.LayoutParams)found.getLayoutParams();
+            final int[] currentRules = currentParams.getRules();
+            final int[] newRule = mMutator.getArgs();
+            final int rule_index = newRule[LayoutCaller.RULE_INDEX];
+            final int[] currentRule = {rule_index, currentRules[rule_index]};
```
joeatwork: When we store these, I'd rather store them as a special object rather than an array with magic indexes. It would be more fun to read `rule.verb` and `rule.anchor`

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -99,12 +100,58 @@ protected String name() {
...
+                mMutator.applyMethodWithArguments(changedView, originalValue);
+            }
+        }
+
+        @Override
+        public void accumulate(View found) {
+            // currentRules is an array which has rule_index as the index and anchor_id as the value
+            // newRule and currentRule are individual rules which look like {rules_index, anchor_id}
+            final RelativeLayout.LayoutParams currentParams = (RelativeLayout.LayoutParams)found.getLayoutParams();
+            final int[] currentRules = currentParams.getRules();
```
joeatwork: Alternately, we could just store the whole `rules` array (which I like!)

joeatwork:
Great progress! I'm excited! I think this can be a bit simpler and less general, and I'd love to see some tests before we merge it.

Thank you so much- I'm excited to see this in action!

src/main/java/com/mixpanel/android/viewcrawler/ViewSnapshot.java
yinfeiru: no, I don't think so.

src/main/java/com/mixpanel/android/viewcrawler/ViewSnapshot.java
yinfeiru: this is needed because we do (RelativeLayout.LayoutParams) (v.getLayoutParams()) below. and if the parent is not relativelayout, this conversion will fail. also, if we don't do this conversion here, we can't call getRules on the params. so, I'd prefer to do this check here, it makes everything much easier.

joeatwork: I'd rather call getLayoutParams() and then check instanceof if we can- it's closer to what we mean.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: I actually prefer to store a JSONObject here, because there's no API which takes `currentRules`. we need to extract the verb and anchor later.

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: bd6100c

yinfeiru: note that I changed 'layout' to 'is_layout' in our analytics codebase (to make it a better looking).

src/main/java/com/mixpanel/android/viewcrawler/LayoutCaller.java
yinfeiru: 92df327

note that int[] mArgs is replaced with JSONObject mArgs in a later branch.

src/main/java/com/mixpanel/android/viewcrawler/LayoutCaller.java
yinfeiru: a6f1205

src/main/java/com/mixpanel/android/viewcrawler/LayoutCaller.java
yinfeiru: this class is removed

src/main/java/com/mixpanel/android/viewcrawler/LayoutCaller.java
yinfeiru: 76dcee1

yinfeiru: this is a nice suggestion!!

src/main/java/com/mixpanel/android/viewcrawler/ViewSnapshot.java
yinfeiru: 5dea492

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: 92df327

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: this class is removed

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: 76dcee1

src/main/java/com/mixpanel/android/viewcrawler/ViewSnapshot.java
yinfeiru: you are so right!!!  923a8dc

yinfeiru:
COOOOOOOL! and you are right, it can be much simpler!
what kind of test are you looking for? the new acceptance test or unit test?

joeatwork:
I think a unit test for the EditProtocol is probably a place to start. That will let us watch things go from JSON to Visitors, and then watch the visitors work.

yinfeiru:
For the visitor test, please check out  e704b48 and e35bb37
For the edit protocol test, please check out 8da1881

src/androidTest/java/com/mixpanel/android/viewcrawler/ViewSnapshotTest.java
```diff
@@ -32,11 +31,11 @@ public void setUp() throws NoSuchMethodException {
 
         final List<PropertyDescription> props = new ArrayList<PropertyDescription>();
 
-        final Caller textGetter = new Caller(TextView.class, "getText", new Object[0], CharSequence.class);
+        final PropertyCaller textGetter = new PropertyCaller(TextView.class, "getText", new Object[0], CharSequence.class);
```
joeatwork: Do we need this rename anymore?

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -90,35 +90,53 @@ public ViewVisitor readEdit(JSONObject source) throws BadInstructionsException {
...
-            if (null == mutator) {
-                throw new BadInstructionsException("Can't update a read-only property " + prop.name + " (add a mutator to make this work)");
-            }
+                final PropertyCaller mutator = prop.makeMutator(methodArgs);
+                if (null == mutator) {
+                    throw new BadInstructionsException("Can't update a read-only property " + prop.name + " (add a mutator to make this work)");
+                }
+
+                return new ViewVisitor.PropertySetVisitor(path, mutator, prop.accessor);
+            } else if (source.has("is_layout")) {
```
joeatwork: I'm still not sold on `is_layout` - it seems like we should have two methods here, one for reading property changes and one for reading LayoutSetVisitors

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -90,35 +90,53 @@ public ViewVisitor readEdit(JSONObject source) throws BadInstructionsException {
...
+                if (layout_info.getString("operation").equals("remove")) {
+                    params.put("anchor", 0);
+                } else if (layout_info.has("anchor_id")) {
+                    params.put("anchor", layout_info.getInt("anchor_id"));
+                } else {
+                    params.put("anchor", RelativeLayout.TRUE);
+                }
 
-            return new ViewVisitor.PropertySetVisitor(path, mutator, prop.accessor);
+                return new ViewVisitor.LayoutSetVisitor(path, params);
```
joeatwork: Consider renaming LayoutSetVisitor to LayoutChangeVisitor- "Set" can mean "assign", or "an unordered collection"

joeatwork: Would you consider a class that can describe the params? JSONObject can be a real pain, mostly because of JSONExceptions. I know that a union would be the most natural thing here (and we don't have one) but you can also use two different classes for the different sorts of arguments, or even just have anchor == null when it isn't required. It seems like we could use the type system, take advantage of fast access rather than slow hash lookups, and make the code a bit cleaner with some immutable description object.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -99,12 +103,66 @@ protected String name() {
...
+            for (Map.Entry<View, JSONObject> original:mOriginalValues.entrySet()) {
+                final View changedView = original.getKey();
+                final JSONObject originalValue = original.getValue();
+                setLayout(changedView, originalValue.optInt("verb"), originalValue.optInt("anchor"));
+            }
+        }
+
+        @Override
+        public void accumulate(View found) {
+            final int newRuleIndex = mArgs.optInt("verb");
```
joeatwork: Will zeroes work here by default? (Or can you detect what is missing because zero isn't a valid value?)

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
joeatwork: I'd also like you to reconsider because we'll end up doing lookups during the visitor's `accumulate` method, which happens a lot :)

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
joeatwork: Let's name `newRuleIndex` as `newVerb`, and use the same terminology everywhere.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -99,12 +103,66 @@ protected String name() {
...
+            setLayout(found, newRuleIndex, newAnchorId);
+        }
+
+        private void setLayout(View target, int ruleIndex, int anchorId) {
+            RelativeLayout.LayoutParams params = (RelativeLayout.LayoutParams)target.getLayoutParams();
+            params.addRule(ruleIndex, anchorId);
+            target.setLayoutParams(params);
+        }
+
+        protected String name() { return "Layout Mutator"; }
```
joeatwork: This name should change

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -99,12 +103,66 @@ protected String name() {
...
 
-        private final Caller mMutator;
-        private final Caller mAccessor;
+        private final PropertyCaller mMutator;
+        private final PropertyCaller mAccessor;
         private final WeakHashMap<View, Object> mOriginalValues;
         private final Object[] mOriginalValueHolder;
     }
 
+    public static class LayoutSetVisitor extends ViewVisitor {
```
joeatwork: I LOVE THIS CLASS! It's really simple!

joeatwork:
Great progress here!

src/androidTest/java/com/mixpanel/android/viewcrawler/ViewSnapshotTest.java
yinfeiru: I'd prefer to keep it PropertyCaller, as we have 2 types of change now. Naming it Caller makes it a bit confused, it implies Caller may get used by LayoutChangeVisitor as well. Do you prefer to call it Caller still?

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: should we have a simple class like
public class LayoutRule {
     public LayoutRule(int v, int a) {
           verb = v;
           anchor  = a;
     }
     public int getVerb() { ... }
     public int getAnchor() { ... }
}

I'm not a big fan of union. Why did you recommend union here? I think we need both of verb and anchor in the params. They are not exclusive.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: It isn't obvious, but, in EditProtocol, I did
```
params.put("verb", layout_info.getInt("rule_id"));
```
so the verb will exist for sure. Otherwise, the code won't reach this point.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: ah.. I saw this in property visitor:
```
protected String name() {
            return "Property Mutator";
        }
```

What should be the proper return value in this case?

src/androidTest/java/com/mixpanel/android/viewcrawler/ViewSnapshotTest.java
joeatwork: Caller isn't really a type of change, and it isn't limited to properties- you can use callers to call any method. It's what the property visitor uses, but it isn't a visitor itself.

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
joeatwork: Sure! In fact, we could do it even more simply:
```
class ViewVisitor {
    public static class LayoutRule {
        public LayoutRule(int aVerb, int anAnchor) {
             verb = aVerb;
             anchor = anAnchor;
        }

        public final verb;
        public final anchor;
    }
}
```

Or even more simply

```
final Pair<Integer, Integer> rule = new Pair<Integer, Integer>(...);
```

although I like the first one a bit better.

joeatwork: Or if they are always present, this is even better:

```new LayoutSetVisitor(path, verb, anchor)```

File: src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java (hunk 17 above)
Comment (joeatwork) [inline src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java]: Some human readable name for the class. "Layout Update" or "Layout Assignment" or whatever you'd like can go there- whatever describes the class.

File: src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java (hunk 15 above)
Comment (yinfeiru) [inline src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java]: I don't like 
```
new LayoutSetVisitor(path, verb, anchor)
```
The reason is that the mOriginalValue will store verb and anchor in a single object, which means we need to create an object containing verb and anchor anyway. I'd rather create an object in the very beginning and use it along the code path.

I don't like `Pair` either because we'll end up with having a magic index for verb and anchor again.

I'll vote for the class approach.

File: src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java (hunk 16 above)
Comment (yinfeiru) [inline src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java]: a84e1aa

File: src/androidTest/java/com/mixpanel/android/viewcrawler/ViewSnapshotTest.java (hunk 13 above)
Comment (yinfeiru) [inline src/androidTest/java/com/mixpanel/android/viewcrawler/ViewSnapshotTest.java]: 19de072

File: src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java (hunk 15 above)
Comment (yinfeiru) [inline src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java]: 3c35afe

File: src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java (hunk 17 above)
Comment (yinfeiru) [inline src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java]: ca6ed99

File: src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java (hunk 15 above)
Comment (yinfeiru) [inline src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java]: ca6ed99

File: src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java (hunk 14 above)
Comment (yinfeiru) [inline src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java]: 3c35afe

File: src/androidTest/java/com/mixpanel/android/viewcrawler/EditProtocolTest.java [hunk 19]
```diff
@@ -34,6 +35,8 @@ public void setUp() throws JSONException {
         mPropertyEdit = new JSONObject(
             "{\"path\":[{\"view_class\":\"com.mixpanel.android.viewcrawler.TestView\",\"index\":0},{\"view_class\":\"android.widget.LinearLayout\",\"index\":0},{\"view_class\":\"android.widget.LinearLayout\",\"index\":0},{\"view_class\":\"android.widget.Button\",\"index\":1}],\"property\":{\"classname\":\"android.widget.Button\",\"name\":\"text\",\"get\":{\"selector\":\"getText\",\"parameters\":[],\"result\":{\"type\":\"java.lang.CharSequence\"}},\"set\":{\"selector\":\"setText\",\"parameters\":[{\"type\":\"java.lang.CharSequence\"}]}},\"args\":[[\"Ground Control to Major Tom\",\"java.lang.CharSequence\"]]}"
         );
+        mLayoutEdit = new JSONObject(
+            String.format("{\"args\":[{\"rule_id\":\"11\",\"operation\":\"add\"}],\"path\": [{\"prefix\": \"shortest\", \"index\": 0, \"id\": %d }],\"is_layout\": true}", TestView.RELATIVE_LAYOUT_BUTTON_ID));
```
joeatwork: rule_id should probably be a number rather than a string

joeatwork: I thought we were moving away from "is_layout" to "edit_type"

src/androidTest/java/com/mixpanel/android/viewcrawler/ViewVisitorTest.java
```diff
@@ -400,6 +407,21 @@ public void testDuplicateTextSet() throws NoSuchMethodException {
         }
     }
 
+    public void testLayoutVisitor () throws JSONException {
+        JSONObject params = new JSONObject();
+        params.put("verb", 0);
+        params.put("anchor", RelativeLayout.TRUE);
```
joeatwork: We discussed making this a class with two final members, `verb` and `anchor`

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -90,35 +90,53 @@ public ViewVisitor readEdit(JSONObject source) throws BadInstructionsException {
...
-            final Caller mutator = prop.makeMutator(methodArgs);
-            if (null == mutator) {
-                throw new BadInstructionsException("Can't update a read-only property " + prop.name + " (add a mutator to make this work)");
-            }
+                final Caller mutator = prop.makeMutator(methodArgs);
+                if (null == mutator) {
+                    throw new BadInstructionsException("Can't update a read-only property " + prop.name + " (add a mutator to make this work)");
+                }
+
+                return new ViewVisitor.PropertySetVisitor(path, mutator, prop.accessor);
```
joeatwork: We do early returns in a lot of places, but usually for failures or "can't possibly deal with this" cases. In this case, I feel like it would be best to return once at the bottom of the method.

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -90,35 +90,53 @@ public ViewVisitor readEdit(JSONObject source) throws BadInstructionsException {
...
+                final JSONArray args = source.getJSONArray("args");
+                JSONObject layout_info = args.optJSONObject(0);
+                ViewVisitor.LayoutRule params;
+                int verb = layout_info.getInt("verb");
+                if (layout_info.getString("operation").equals("remove")) {
+                    params = new ViewVisitor.LayoutRule(verb, 0);
+                } else if (layout_info.has("anchor_id")) {
+                    params = new ViewVisitor.LayoutRule(verb, layout_info.getInt("anchor_id"));
+                } else {
+                    params = new ViewVisitor.LayoutRule(verb, RelativeLayout.TRUE);
```
joeatwork: SWEET! I love LayoutRule! You may want to update the tests, though :)

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -90,35 +90,53 @@ public ViewVisitor readEdit(JSONObject source) throws BadInstructionsException {
...
+
+                return new ViewVisitor.PropertySetVisitor(path, mutator, prop.accessor);
+            } else if (source.getString("change_type").equals("layout")) {
+                final JSONArray args = source.getJSONArray("args");
+                JSONObject layout_info = args.optJSONObject(0);
+                ViewVisitor.LayoutRule params;
+                int verb = layout_info.getInt("verb");
+                if (layout_info.getString("operation").equals("remove")) {
+                    params = new ViewVisitor.LayoutRule(verb, 0);
+                } else if (layout_info.has("anchor_id")) {
```
joeatwork: Could input ever have both anchor_id and remove? I feel like we might be able to streamline this, so that we can't create invalid combinations of properties. Perhaps just expect anchor_id == 0?

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -105,6 +109,64 @@ protected String name() {
...
```
joeatwork: I'd like to change the "Set" in LayoutSetVisitor to maybe LayoutUpdateVisitor or LayoutRulesVisitor (or just LayoutVisitor)?

joeatwork:
I LOVE THIS!

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: Ah.. Actually, RelativeLayout.TRUE is -1. So, we need to first check if it's "remove", if so, set anchor_id to 0. If not, we want to know if anchor_id is provided. If it's there (indicating it's an anchor layout update), we want to get the anchor_id. If not, we send anchor_id to -1. So, we are having 3 different values for anchor_id here. 

And, on the mixpanel website, I don't want to send "anchor_id=0" for removing, as I don't want to make any assumption of what value Android is using for disabling layouts. Same idea for setting parent layout update, where I can assume RelativeLayout.TRUE to be -1 but I'm not doing it.

src/androidTest/java/com/mixpanel/android/viewcrawler/EditProtocolTest.java
yinfeiru: TOTALLY forgot to update the tests..! will do this.. thanks for pointing out!!

yinfeiru: 9741ed0  sorry about this..

src/androidTest/java/com/mixpanel/android/viewcrawler/ViewVisitorTest.java
yinfeiru: 9741ed0

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: 9ca3060

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: ad5c0e6

joeatwork:
Let's merge this!

PR #200, opened 2015-04-08
Initial Image storage and protocol

joeatwork:
Reading, storing, and reporting on Bitmaps and Drawables in the library.


src/androidTest/java/com/mixpanel/android/mpmetrics/DecideCheckerTest.java
```diff
@@ -29,7 +33,7 @@ public void setUp() {
         mDecideMessages3.setDistinctId("DISTINCT ID 3");
     }
 
-    public void testReadEmptyLists() throws ServiceUnavailableException {
+    public void testReadEmptyLists() throws RemoteService.ServiceUnavailableException {
```
yinfeiru: Why didn't we need this `RemoteService` thing before but now we do?

src/androidTest/java/com/mixpanel/android/mpmetrics/DecideCheckerTest.java
```diff
@@ -191,18 +195,24 @@ public void testBadDecideResponses() throws ServiceUnavailableException {
...
+    public void testDecideHonorsFallbackEnabled() throws RemoteService.ServiceUnavailableException {
         mConfig.fallbackDisabled = false;
-        mPoster.response = bytes("{\"surveys\":[], \"notifications\":[]}");
+        mPoster.requestedUrls.clear();
+        mPoster.response = null;
+        mPoster.exception = new IOException("Bang!");
         mDecideChecker.addDecideCheck(mDecideMessages1);
         mDecideChecker.runDecideChecks(mPoster);
-        assertEquals(mPoster.requestedUrls.length, 2);
+        assertEquals(2, mPoster.requestedUrls.size());
```
yinfeiru: YES! This is the proper order of parameters for assertEquals. And we are having wrong orders all over the place.

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -130,8 +131,8 @@ protected MPConfig getConfig(Context context) {
         return MPConfig.getInstance(context);
     }
 
-    protected ServerMessage getPoster() {
-        return new ServerMessage();
+    protected RemoteService getPoster() {
```
yinfeiru: what's the difference between ServerMessage and RemoteService? From the lib directory, I guess RemoteService is specific to Android?

src/main/java/com/mixpanel/android/mpmetrics/InAppNotification.java
```diff
@@ -200,6 +200,6 @@ public InAppNotification createFromParcel(Parcel source) {
     private final String mCallToAction;
     private final String mCallToActionUrl;
 
-    private static final String LOGTAG = "MixpanelAPI.InAppNotification";
+    private static final String LOGTAG = "MixpanelAPI.InAppNotif";
```
yinfeiru: why?

src/main/java/com/mixpanel/android/mpmetrics/DecideChecker.java
```diff
@@ -263,6 +268,36 @@ private static int getDisplayWidth(final Display display) {
         }
     }
 
+    private static byte[] getUrls(RemoteService poster, Context context, String[] urls)
```
yinfeiru: what is the advantage you get out of moving `getUrls` from `RemoteService` to `Decidechecker`?

src/main/java/com/mixpanel/android/util/RemoteService.java
```diff
@@ -0,0 +1,36 @@
...
+
+import android.content.Context;
+
+import org.apache.http.NameValuePair;
+
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Created by joe on 4/2/15.
```
yinfeiru: Probably don't need this comment here?

src/main/java/com/mixpanel/android/util/HttpService.java
```diff
@@ -107,7 +82,7 @@ public boolean isOnline(Context context) {
                     out = null;
                 }
                 in = connection.getInputStream();
-                response = slurp(in);
+                response = Utilities.slurp(in);
```
yinfeiru: I'm not sure about this, but do you think it's an overkill of having slurp in the Utilities.java? It's only referenced once in the code base.

src/main/java/com/mixpanel/android/util/RemoteService.java
```diff
@@ -0,0 +1,36 @@
...
+                retry = 0;
+            }
+            mRetryAfter = retry;
+        }
+
+        public int getRetryAfter() {
+            return mRetryAfter;
+        }
+
+        private final int mRetryAfter;
```
yinfeiru: Just a random thought and might not correct. Do you think we can have mRetryAfter as a public member, since it's `final` and can't be changed once initialized?

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
```diff
@@ -0,0 +1,138 @@
...
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * ABSOLUTELY NOT thread, or even process safe.
+ * Writes and reads files and directories at known paths, and uses a shared instance of MessageDigest.
+ */
+/* package */ class ImageStore {
+    public ImageStore(Context context) {
+        this(
+            context.getDir(DEFAULT_DIRECTORY_NAME, Context.MODE_PRIVATE),
```
yinfeiru: Do we need the `Context.MODE_PRIVATE` here? Here's the comment from Context.java
```
    /**
     * File creation mode: the default mode, where the created file can only
     * be accessed by the calling application (or all applications sharing the
     * same user ID).
     * @see #MODE_WORLD_READABLE
     * @see #MODE_WORLD_WRITEABLE
     */
    public static final int MODE_PRIVATE = 0x0000;
```

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
```diff
@@ -0,0 +1,138 @@
...
+        final String safeName = FILE_PREFIX + Base64.encodeToString(hashed, Base64.URL_SAFE | Base64.NO_WRAP);
+        return new File(mDirectory, safeName);
+    }
+
+    private final File mDirectory;
+    private final RemoteService mPoster;
+    private final MessageDigest mDigest;
+
+    private static final String DEFAULT_DIRECTORY_NAME = "MixpanelAPI.Images";
+    private static final int MAX_BITMAP_SIZE = 10000000; // 10 MB
```
yinfeiru: Why is it 10MB? Any story behind this?

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
```diff
@@ -0,0 +1,138 @@
...
+                try {
+                    out = new FileOutputStream(file);
+                    out.write(bytes);
+                } catch (FileNotFoundException e) {
+                    Log.e(LOGTAG, "It appears that ImageStore is misconfigured, or disk storage is unavailable- can't write to bitmap directory", e);
+                } catch (IOException e) {
+                    Log.w(LOGTAG, "Can't store bitmap", e);
+                } finally {
+                    if (null != out) {
+                        try {
```
yinfeiru: Is this for the case where out.write is half way through and throws (since `out.close` can be put below `out.write(bytes)`)?

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    public void clearStorage() {
+        File[] files = mDirectory.listFiles();
+        int length = files.length;
+        for (int i = 0; i < length; i++) {
+            final File file = files[i];
+            final String filename = file.getName();
+            if (filename.startsWith(FILE_PREFIX)) {
+                final boolean ignored = file.delete();
```
yinfeiru: looks like this `ignored` variable is not used

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
```diff
@@ -0,0 +1,138 @@
...
+            if (null == bitmap) {
+                Log.w(LOGTAG, "Bitmap on disk can't be opened or is corrupt");
+                final boolean ignored = file.delete();
+            }
+        }
+
+        return bitmap;
+    }
+
+    public void clearStorage() {
```
yinfeiru: Can you please explain how this guy works? I'm confused about
1. who's calling it
2. when you remove a file, why do you match only the FILE_PREFIX? Isn't this function going to remove all the files within this directory regardless of whether or not the image will be used?
3. what if a user upload image A and then replace it with image B? Are A and B both in the memory? Is it a waste of memory? In `getImage`, you are calling `bytes = mPoster.performRequest(url, null);` without checking if the image exists in the disk. So I don't think we want to keep the old image on the disk.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
```diff
@@ -0,0 +1,138 @@
...
+            if (filename.startsWith(FILE_PREFIX)) {
+                final boolean ignored = file.delete();
+            }
+        }
+    }
+
+    public void deleteStorage(String url) {
+        final File file = storedFile(url);
+        if (null != file) {
+            final boolean ignored = file.delete();
```
yinfeiru: same here

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
yinfeiru: now, the only question I have is - do we remove the old image once it gets replaced with a new image from memory? It may potentially eat up a lot of memory.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
```diff
@@ -0,0 +1,138 @@
+package com.mixpanel.android.viewcrawler;
```
yinfeiru: I don't know how the image gets displayed on Android. But, what if a image is originally 100 x 100 and the view is only 50 x 50. Is Android going to shrink that image or cut it off?

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
```diff
@@ -0,0 +1,138 @@
...
+            } catch (IOException e) {
+                Log.i(LOGTAG, "Can't download bitmap", e);
+            } catch (RemoteService.ServiceUnavailableException e) {
+                Log.i(LOGTAG, "Couldn't download ");
+            }
+        }
+
+        final Bitmap bitmap;
+        if (null != bytes) {
+            if (null != file && bytes.length < MAX_BITMAP_SIZE) {
```
yinfeiru: A tooltip on the website is needed for such limitation.

src/androidTest/java/com/mixpanel/android/viewcrawler/ViewVisitorTest.java
```diff
@@ -305,32 +322,57 @@ public void testResetSameEventOnClick() {
...
-        paint.setColor(Color.BLUE);
+        final Caller mutateBitmap1a = new Caller(TestView.AdHocButton2.class, "setCountingProperty", new Object[] { mBitmap1a }, Void.TYPE);
+        final Caller mutateBitmap1b = new Caller(TestView.AdHocButton2.class, "setCountingProperty", new Object[] { mBitmap1b }, Void.TYPE);
+        final Caller mutateBitmap2 = new Caller(TestView.AdHocButton2.class, "setCountingProperty", new Object[] { mBitmap2 }, Void.TYPE);
+        final Caller accessor = new Caller(TestView.AdHocButton2.class, "getCountingProperty", new Object[]{}, Object.class);
 
-        final Bitmap.Config conf = Bitmap.Config.ARGB_8888;
-        final Bitmap bitmap1a = Bitmap.createBitmap(10, 10, conf); // this creates a MUTABLE bitmap
-        final Canvas canvas1 = new Canvas(bitmap1a);
-        canvas1.drawCircle(3, 3, 4, paint);
```
yinfeiru: what is this for? Are we getting rid of it?

yinfeiru:
THIS IS VERY COOL!! WE ARE GETTING THERE!!

src/androidTest/java/com/mixpanel/android/mpmetrics/DecideCheckerTest.java
joeatwork: :)

src/main/java/com/mixpanel/android/mpmetrics/DecideChecker.java
joeatwork: I wanted to use the same code path for retrieving images (which are in a different package) - it seemed like performRequest was the right place to share it. In addition, getUrls wasn't used anywhere else, so this was an opportunity to reduce the public method count on the whole system.

src/main/java/com/mixpanel/android/mpmetrics/InAppNotification.java
joeatwork: There is a size limit to LOGTAGS that we've been ignoring, so I've been shortening the tags as I see the errors. (this is issue #176 )

src/main/java/com/mixpanel/android/util/HttpService.java
joeatwork: You're right! I've moved it in f6b3925

src/main/java/com/mixpanel/android/util/RemoteService.java
joeatwork: Fixed 2e19277

src/main/java/com/mixpanel/android/util/RemoteService.java
joeatwork: In this case, I'd like to keep it private - I'm a fan of `public final` members for struct-like objects with short lifetimes, but in this case I can imagine a subclass of ServiceUnavailableException that might behave differently.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: Context.getDir doesn't have a one-argument signature, and in this case it's important that the directory be private to the app.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: Thanks!

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: It's still open- out.write could throw immediately, or even some random spot during the execution, and we want to make sure to close the stream. This is an annoying idiom, but otherwise we have to wait until GC happens to close the file if we open it and anything goes wrong.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: This is a workaround for a lint warning

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: Actually, this is only run during tests, but we should be calling `deleteStorage` when we roll back editor changes. I'll need to do that!

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: The limit is basically arbitrary- I wanted to make sure we didn't write unbounded bytes to storage.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
```diff
@@ -0,0 +1,138 @@
...
+            useDigest = MessageDigest.getInstance("SHA1");
+        } catch (NoSuchAlgorithmException e) {
+            Log.w(LOGTAG, "Images won't be stored because this platform doesn't supply a SHA1 hash function");
+            useDigest = null;
+        }
+
+        mDigest = useDigest;
+    }
+
+    public Bitmap getImage(String url) {
```
joeatwork: This should throw on failure, not return null, for consistency with EditProtocol.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
yinfeiru: oh.. sorry.. I was looking at the constructor..

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
yinfeiru: In that case, should we check if the device has enough space remaining? I don't think we want to take up too much space. (maybe same as what we did to the tracking calls, where we limit the maximum space taking to half of the available space?)

yinfeiru:
One more thing, do we want to test this feature on different devices? As you mentioned, the image may get displayed differently on devices with different screen sizes. (it can be easy to test out if we download several simulator templates in Android Studio)

joeatwork:
That's an excellent point! Right now, scaling the image is still to be done.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: Sounds good!

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: See above, this is to make the linter be quiet.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: Scaling is still in process.

src/main/java/com/mixpanel/android/viewcrawler/ImageStore.java
joeatwork: Fixed in 5bce474

PR #203, opened 2015-04-15
Delete images when editor closes

joeatwork:


src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -127,7 +137,7 @@ public ViewVisitor readEdit(JSONObject source) throws BadInstructionsException,
                     final JSONArray argPlusType = argsAndTypes.getJSONArray(i);
                     final Object jsonArg = argPlusType.get(0);
                     final String argType = argPlusType.getString(1);
-                    methodArgs[i] = convertArgument(jsonArg, argType);
+                    methodArgs[i] = convertArgument(jsonArg, argType, assetsLoaded);
```
yinfeiru: Is it a common pattern in JAVA that you pass in an array and modify it in the function, instead of assigning the returned value from the function to the array?

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -52,6 +52,15 @@ public CantGetEditAssetsException(String message, Throwable cause) {
         }
     }
 
+    public static class Edit {
+        private Edit(ViewVisitor aVisitor, List<String> someUrls) {
```
yinfeiru: I saw this kind of thing several times, but is it a common thing to use the quantity as the prefix? 
`aVisitor` `someUrls`

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -267,7 +259,18 @@ public ViewCrawlerHandler(Context context, String token, Looper looper) {
...
+            String resourcePackage = mConfig.getResourcePackageName();
+            if (null == resourcePackage) {
+                resourcePackage = context.getPackageName();
+            }
+
+            final ResourceIds resourceIds = new ResourceReader.Ids(resourcePackage, context);
+
+            mImageStore = new ImageStore(context);
+            mProtocol = new EditProtocol(resourceIds, mImageStore);
+
```
yinfeiru: Tiny but why do we need this newline?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -679,15 +682,17 @@ private void handleEditorBindingsCleared(JSONObject clearMessage) {
...
                     final String changeId = actions.getString(i);
                     mEditorChanges.remove(changeId);
                 }
-
-                updateEditState();
             } catch (final JSONException e) {
                 Log.e(LOGTAG, "Bad clear request received", e);
             }
+
+            updateEditState();
```
yinfeiru: nice!

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -878,13 +887,17 @@ private SharedPreferences getSharedPreferences() {
             return mContext.getSharedPreferences(sharedPrefsName, Context.MODE_PRIVATE);
         }
 
+
```
yinfeiru: newline?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -812,8 +820,9 @@ private void updateEditState() {
             {
                 for (Pair<String, JSONObject> changeInfo:mEditorChanges.values()) {
                     try {
-                        final ViewVisitor visitor = mProtocol.readEdit(changeInfo.second);
-                        newEdits.add(new Pair<String, ViewVisitor>(changeInfo.first, visitor));
+                        final EditProtocol.Edit edit = mProtocol.readEdit(changeInfo.second);
+                        newEdits.add(new Pair<String, ViewVisitor>(changeInfo.first, edit.visitor));
```
yinfeiru: I can understand what this logic does but it looks a bit confusing to me.. We have variables called `newEdits` and `edit`. However, when we add new elements to newEdits, we are not adding edit, but `new Pair<String, ViewVisitor>(changeInfo.first, edit.visitor)` (`edit.visitor` but not `edit`.. and ideally, it should be `newEdits.add(edit)`). Not a big deal, but the logic is different from what the variable names imply. Never mind, any change to newEdits will affect a bunch of stuff. It's ok to leave as it is.

yinfeiru:
Is EditProtocolTest.java the only file we need to change? As long as you've run the full test suite, I'm good with it.

Coolio!!

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
joeatwork: In this case, I wanted to add elements to the list as a side effect, and wanted to keep the work on the stack rather than in the state of the object. This isn't great- It would be better to gather the resources separately, but there is value to having individual methods parse individual pieces of the input, so I made a trade off.

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
joeatwork: Some Java codebases prefix their arguments with "a"- I'm doing it here (And do it elsewhere) when the natural argument name would be the same as a public member name, as the compliment to the convention of prefixing "m" to member names (since we don't do that with public members). I used "some" rather than "a" for URLS for grammar (because aUrls seems strange)

The alternative would be something like
```
private Edit(ViewVisitor visitor, List<String> urls) {
    this.visitor = visitor;
    this.urls = urls;
}
```
Which I don't like as much, since the this looks redundant (and usually is redundant)

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: Fixed in e3f22f8

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: Fixed in e3f22f8

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: I agree! Referring to Visitors as "Edits" was meaningful before we had an "Edit" object that meant something different. I've changed `newEdits` to `newVisitors` in ba93c14

PR #204, opened 2015-04-16
use buildToolsVersion 21.1.2

nemecec:
The latest buildToolsVersion in 21.1.x range is 21.1.2.


joeatwork:
Thanks for the patch!

joeatwork:
This change will be present in v4.5.4

PR #206, opened 2015-04-17
Catch layout exception

yinfeiru:
detect circular dependency exception and send error message to web ui

### DIFF
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
@@ -141,6 +141,7 @@ import java.util.List;
                 JSONObject layout_info = args.optJSONObject(0);
                 ViewVisitor.LayoutRule params;
                 int verb = layout_info.getInt("verb");
+                String name = source.getString("name");
                 if (layout_info.getString("operation").equals("remove")) {
                     params = new ViewVisitor.LayoutRule(verb, 0);
                 } else if (layout_info.has("anchor_id")) {
@@ -149,7 +150,7 @@ import java.util.List;
                     params = new ViewVisitor.LayoutRule(verb, RelativeLayout.TRUE);
                 }
 
-                visitor = new ViewVisitor.LayoutUpdateVisitor(path, params);
+                visitor = new ViewVisitor.LayoutUpdateVisitor(path, params, name);
             } else {
                 throw new BadInstructionsException("Can't figure out the edit type");
             }
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
@@ -52,9 +52,14 @@ import java.util.List;
         }
     }
 
-    public EditProtocol(ResourceIds resourceIds, ImageStore imageStore) {
+    public interface EditErrorMessage {
+        public void sendErrorMessage(ViewVisitor.CantVisitException e);
+    }
+
+    public EditProtocol(ResourceIds resourceIds, ImageStore imageStore, EditErrorMessage editErrorMessage) {
         mResourceIds = resourceIds;
         mImageStore = imageStore;
+        mEditErrorMessage = editErrorMessage;
     }
 
     public ViewVisitor readEventBinding(JSONObject source, ViewVisitor.OnEventListener listener) throws BadInstructionsException {
@@ -150,7 +155,7 @@ import java.util.List;
                     params = new ViewVisitor.LayoutRule(verb, RelativeLayout.TRUE);
                 }
 
-                visitor = new ViewVisitor.LayoutUpdateVisitor(path, params, name);
+                visitor = new ViewVisitor.LayoutUpdateVisitor(path, params, name, mEditErrorMessage);
             } else {
                 throw new BadInstructionsException("Can't figure out the edit type");
             }
@@ -369,6 +374,7 @@ import java.util.List;
 
     private final ResourceIds mResourceIds;
     private final ImageStore mImageStore;
+    private final EditErrorMessage mEditErrorMessage;
 
     private static final Class<?>[] NO_PARAMS = new Class[0];
     private static final List<Pathfinder.PathElement> NEVER_MATCH_PATH = Collections.<Pathfinder.PathElement>emptyList();
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
@@ -52,11 +52,7 @@ import java.util.List;
         }
     }
 
-    public interface OnErrorListener {
-        public void onError(ViewVisitor.CantVisitException e);
-    }
-
-    public EditProtocol(ResourceIds resourceIds, ImageStore imageStore, OnErrorListener editErrorListener) {
+    public EditProtocol(ResourceIds resourceIds, ImageStore imageStore, ViewVisitor.OnErrorListener editErrorListener) {
         mResourceIds = resourceIds;
         mImageStore = imageStore;
         mEditErrorListener = editErrorListener;
@@ -374,7 +370,7 @@ import java.util.List;
 
     private final ResourceIds mResourceIds;
     private final ImageStore mImageStore;
-    private final OnErrorListener mEditErrorListener;
+    private final ViewVisitor.OnErrorListener mEditErrorListener;
 
     private static final Class<?>[] NO_PARAMS = new Class[0];
     private static final List<Pathfinder.PathElement> NEVER_MATCH_PATH = Collections.<Pathfinder.PathElement>emptyList();
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
@@ -53,7 +53,7 @@ import java.util.List;
     }
 
     public interface OnErrorListener {
-        public void sendErrorMessage(ViewVisitor.CantVisitException e);
+        public void onError(ViewVisitor.CantVisitException e);
     }
 
     public EditProtocol(ResourceIds resourceIds, ImageStore imageStore, OnErrorListener editErrorListener) {
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
@@ -52,14 +52,14 @@ import java.util.List;
         }
     }
 
-    public interface EditErrorMessage {
+    public interface OnErrorListener {
         public void sendErrorMessage(ViewVisitor.CantVisitException e);
     }
 
-    public EditProtocol(ResourceIds resourceIds, ImageStore imageStore, EditErrorMessage editErrorMessage) {
+    public EditProtocol(ResourceIds resourceIds, ImageStore imageStore, OnErrorListener editErrorListener) {
         mResourceIds = resourceIds;
         mImageStore = imageStore;
-        mEditErrorMessage = editErrorMessage;
+        mEditErrorListener = editErrorListener;
     }
 
     public ViewVisitor readEventBinding(JSONObject source, ViewVisitor.OnEventListener listener) throws BadInstructionsException {
@@ -155,7 +155,7 @@ import java.util.List;
                     params = new ViewVisitor.LayoutRule(verb, RelativeLayout.TRUE);
                 }
 
-                visitor = new ViewVisitor.LayoutUpdateVisitor(path, params, name, mEditErrorMessage);
+                visitor = new ViewVisitor.LayoutUpdateVisitor(path, params, name, mEditErrorListener);
             } else {
                 throw new BadInstructionsException("Can't figure out the edit type");
             }
@@ -374,7 +374,7 @@ import java.util.List;
 
     private final ResourceIds mResourceIds;
     private final ImageStore mImageStore;
-    private final EditErrorMessage mEditErrorMessage;
+    private final OnErrorListener mEditErrorListener;
 
     private static final Class<?>[] NO_PARAMS = new Class[0];
     private static final List<Pathfinder.PathElement> NEVER_MATCH_PATH = Collections.<Pathfinder.PathElement>emptyList();
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditState.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditState.java
@@ -139,7 +139,7 @@ import java.util.Set;
     /* The binding between a bunch of edits and a view. Should be instantiated and live on the UI thread */
     private static class EditBinding implements ViewTreeObserver.OnGlobalLayoutListener, Runnable {
         public EditBinding(View viewRoot, ViewVisitor edit, Handler uiThreadHandler)
-        throws ViewVisitor.LayoutUpdateException {
+                throws ViewVisitor.LayoutUpdateException {
             mEdit = edit;
             mViewRoot = new WeakReference<View>(viewRoot);
             mHandler = uiThreadHandler;
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditState.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditState.java
@@ -26,7 +26,7 @@ import java.util.Set;
 /* package */ class EditState extends UIThreadSet<Activity> {
 
     public interface EditErrorMessage {
-        public void sendErrorMessage(JSONObject errorMessage);
+        public void sendErrorMessage(ViewVisitor.CantVisitException e);
     }
 
     public EditState(EditErrorMessage editerrorMessage) {
@@ -126,25 +126,21 @@ import java.util.Set;
             final int size = changes.size();
             for (int i = 0; i < size; i++) {
                 final ViewVisitor visitor = changes.get(i);
-                try {
-                    final EditBinding binding = new EditBinding(rootView, visitor, mUiThreadHandler);
-                    mCurrentEdits.add(binding);
-                } catch (ViewVisitor.LayoutUpdateException e) {
-                    mEditErrorMessage.sendErrorMessage(e.getErrorInfo());
-                }
+                final EditBinding binding = new EditBinding(rootView, visitor, mUiThreadHandler, mEditErrorMessage);
+                mCurrentEdits.add(binding);
             }
         }
     }
 
     /* The binding between a bunch of edits and a view. Should be instantiated and live on the UI thread */
     private static class EditBinding implements ViewTreeObserver.OnGlobalLayoutListener, Runnable {
-        public EditBinding(View viewRoot, ViewVisitor edit, Handler uiThreadHandler)
-                throws ViewVisitor.LayoutUpdateException {
+        public EditBinding(View viewRoot, ViewVisitor edit, Handler uiThreadHandler, EditErrorMessage editErrorMessage) {
             mEdit = edit;
             mViewRoot = new WeakReference<View>(viewRoot);
             mHandler = uiThreadHandler;
             mAlive = true;
             mDying = false;
+            mEditErrorMessage = editErrorMessage;
 
             final ViewTreeObserver observer = viewRoot.getViewTreeObserver();
             if (observer.isAlive()) {
@@ -173,9 +169,9 @@ import java.util.Set;
 
             try {
                 mEdit.visit(viewRoot);
-            } catch (ViewVisitor.LayoutUpdateException e) {
+            } catch (ViewVisitor.CantVisitException e) {
                 cleanUp();
-                throw e;
+                mEditErrorMessage.sendErrorMessage(e);
             }
             mHandler.removeCallbacks(this);
             mHandler.postDelayed(this, 1000);
@@ -206,6 +202,7 @@ import java.util.Set;
         private final WeakReference<View> mViewRoot;
         private final ViewVisitor mEdit;
         private final Handler mHandler;
+        private final EditErrorMessage mEditErrorMessage;
     }
 
     private final Handler mUiThreadHandler;
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditState.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditState.java
@@ -3,11 +3,8 @@ package com.mixpanel.android.viewcrawler;
 import android.app.Activity;
 import android.os.Handler;
 import android.os.Looper;
-import android.util.Log;
 import android.view.View;
 import android.view.ViewTreeObserver;
-
-import org.json.JSONObject;
 
 import java.lang.ref.WeakReference;
 import java.util.HashMap;
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditState.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditState.java
@@ -25,15 +25,10 @@ import java.util.Set;
  */
 /* package */ class EditState extends UIThreadSet<Activity> {
 
-    public interface EditErrorMessage {
-        public void sendErrorMessage(ViewVisitor.CantVisitException e);
-    }
-
-    public EditState(EditErrorMessage editerrorMessage) {
+    public EditState() {
         mUiThreadHandler = new Handler(Looper.getMainLooper());
         mIntendedEdits = new HashMap<String, List<ViewVisitor>>();
         mCurrentEdits = new HashSet<EditBinding>();
-        mEditErrorMessage = editerrorMessage;
     }
 
     /**
@@ -126,7 +121,7 @@ import java.util.Set;
             final int size = changes.size();
             for (int i = 0; i < size; i++) {
                 final ViewVisitor visitor = changes.get(i);
-                final EditBinding binding = new EditBinding(rootView, visitor, mUiThreadHandler, mEditErrorMessage);
+                final EditBinding binding = new EditBinding(rootView, visitor, mUiThreadHandler);
                 mCurrentEdits.add(binding);
             }
         }
@@ -134,13 +129,12 @@ import java.util.Set;
 
     /* The binding between a bunch of edits and a view. Should be instantiated and live on the UI thread */
     private static class EditBinding implements ViewTreeObserver.OnGlobalLayoutListener, Runnable {
-        public EditBinding(View viewRoot, ViewVisitor edit, Handler uiThreadHandler, EditErrorMessage editErrorMessage) {
+        public EditBinding(View viewRoot, ViewVisitor edit, Handler uiThreadHandler) {
             mEdit = edit;
             mViewRoot = new WeakReference<View>(viewRoot);
             mHandler = uiThreadHandler;
             mAlive = true;
             mDying = false;
-            mEditErrorMessage = editErrorMessage;
 
             final ViewTreeObserver observer = viewRoot.getViewTreeObserver();
             if (observer.isAlive()) {
@@ -167,12 +161,7 @@ import java.util.Set;
             }
             // ELSE View is alive and we are alive
 
-            try {
-                mEdit.visit(viewRoot);
-            } catch (ViewVisitor.CantVisitException e) {
-                cleanUp();
-                mEditErrorMessage.sendErrorMessage(e);
-            }
+            mEdit.visit(viewRoot);
             mHandler.removeCallbacks(this);
             mHandler.postDelayed(this, 1000);
         }
@@ -202,13 +191,11 @@ import java.util.Set;
         private final WeakReference<View> mViewRoot;
         private final ViewVisitor mEdit;
         private final Handler mHandler;
-        private final EditErrorMessage mEditErrorMessage;
     }
 
     private final Handler mUiThreadHandler;
     private final Map<String, List<ViewVisitor>> mIntendedEdits;
     private final Set<EditBinding> mCurrentEdits;
-    private final EditErrorMessage mEditErrorMessage;
 
     @SuppressWarnings("unused")
     private static final String LOGTAG = "MixpanelAPI.EditState";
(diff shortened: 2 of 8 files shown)


src/main/java/com/mixpanel/android/viewcrawler/EditState.java
```diff
@@ -159,7 +171,12 @@ public void run() {
             }
             // ELSE View is alive and we are alive
 
-            mEdit.visit(viewRoot);
+            try {
+                mEdit.visit(viewRoot);
+            } catch (ViewVisitor.LayoutUpdateException e) {
```
joeatwork: This seems like it may be to early to catch the error- we'll continue visiting every time the system updates the global view layout, and could throw in there, too.

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
```diff
@@ -119,15 +126,20 @@ private void applyChangesFromList(View rootView, List<ViewVisitor> changes) {
             final int size = changes.size();
             for (int i = 0; i < size; i++) {
                 final ViewVisitor visitor = changes.get(i);
-                final EditBinding binding = new EditBinding(rootView, visitor, mUiThreadHandler);
-                mCurrentEdits.add(binding);
+                try {
+                    final EditBinding binding = new EditBinding(rootView, visitor, mUiThreadHandler);
+                    mCurrentEdits.add(binding);
+                } catch (ViewVisitor.LayoutUpdateException e) {
+                    mEditErrorMessage.sendErrorMessage(e.getErrorInfo());
```
joeatwork: This seems like it may be to early to catch the error- we'll continue visiting every time the system updates the global view layout, and could throw in there, too.

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
```diff
@@ -159,7 +171,12 @@ public void run() {
             }
             // ELSE View is alive and we are alive
 
-            mEdit.visit(viewRoot);
+            try {
+                mEdit.visit(viewRoot);
+            } catch (ViewVisitor.LayoutUpdateException e) {
+                cleanUp();
+                throw e;
```
joeatwork: I'm worried that this rethrow will crash the app. Perhaps LayoutUpdateException shouldn't be a runtime error?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -651,6 +661,30 @@ private void sendReportTrackToEditor(String eventName) {
...
+        private void sendLayoutError(JSONObject errorMessage) {
+            if (mEditorConnection == null || !mEditorConnection.isValid()) {
+                return;
+            }
+
+            final OutputStream out = mEditorConnection.getBufferedOutputStream();
+            final OutputStreamWriter writer = new OutputStreamWriter(out);
+
+            try {
+                errorMessage.put("type", "layout_error");
```
joeatwork: Let's not modify existing messages in place.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -137,6 +136,14 @@ public void reportTrack(String eventName) {
         mMessageThreadHandler.sendMessage(m);
     }
 
+    @Override
+    public void sendErrorMessage(JSONObject errorMessage) {
```
joeatwork: This argument probably shouldn't be JSON - the interface could be any listener, so we can use a real object with good static types.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: We also have an "envelope" protocol for this sort of thing (With "type" and "payload" as top level keys) - in fact, we already have an error message format, it might be good to use that format if possible.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -119,11 +123,35 @@ protected String name() {
         private final Object[] mOriginalValueHolder;
     }
 
+    public static class LayoutUpdateException extends IllegalStateException {
+        public LayoutUpdateException(String message, JSONObject info) {
+            super(message);
+            mErrorInfo = info;
+        }
+
+        public JSONObject getErrorInfo() {
```
joeatwork: Let's not use JSONObjects unless we're doing I/O or serializing things to storage. If we can, it would be great to use types.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -119,11 +123,35 @@ protected String name() {
...
+
     public static class LayoutUpdateVisitor extends ViewVisitor {
-        public LayoutUpdateVisitor(List<Pathfinder.PathElement> path, LayoutRule args) {
+        public LayoutUpdateVisitor(List<Pathfinder.PathElement> path, LayoutRule args, String name) {
             super(path);
             mOriginalValues = new WeakHashMap<View, LayoutRule>();
             mArgs = args;
+            mName = name;
+
+            mHorizontalRules = new ArrayList<>(Arrays.asList(
```
joeatwork: This looks like it could be a static, immutable member (and should probably be a set?)

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +180,111 @@ public void accumulate(View found) {
...
+                }
+                throw new LayoutUpdateException("Circular dependency detected!", errorInfo);
+            }
+
             target.setLayoutParams(params);
         }
 
+        private boolean verifyLayout(View target, ArrayList<Integer> rules) {
+            ViewGroup parent = (ViewGroup) target.getParent();
+            SparseArray<View> idToChild = new SparseArray<View>();
```
joeatwork: This is where we need to be fast, so if we can get out of doing allocations it's a bit better.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +180,111 @@ public void accumulate(View found) {
...
+
+            for (int i = 0; i < parent.getChildCount(); i++) {
+                View child = parent.getChildAt(i);
+                idToChild.put(child.getId(), child);
+            }
+
+            ArrayMap<View, ArrayList<View>> dependencyGraph = new ArrayMap<View, ArrayList<View>>();
```
joeatwork: Let's call this a `Map` rather than an `ArrayMap` (and why is this an ArrayMap?)

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +180,111 @@ public void accumulate(View found) {
...
+                View currentNode = dependencyGraph.keyAt(0);
+                if (!subGraphHasCycle(dependencyGraph, currentNode, dfsStack)) {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private boolean subGraphHasCycle(ArrayMap<View, ArrayList<View>> dependencyGraph,
```
joeatwork: It might be good to break out this dependency checking into it's own class.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +180,111 @@ public void accumulate(View found) {
...
-            setLayout(found, newVerb, newAnchorId);
+
+            try {
+                setLayout(found, newVerb, newAnchorId);
+            } catch (LayoutUpdateException e) {
+                throw e;
+            }
         }
 
         private void setLayout(View target, int verb, int anchorId) {
```
joeatwork: Can setLayout be inlined into accumulate?

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +180,111 @@ public void accumulate(View found) {
...
             target.setLayoutParams(params);
         }
 
+        private boolean verifyLayout(View target, ArrayList<Integer> rules) {
+            ViewGroup parent = (ViewGroup) target.getParent();
+            SparseArray<View> idToChild = new SparseArray<View>();
+
+            for (int i = 0; i < parent.getChildCount(); i++) {
+                View child = parent.getChildAt(i);
+                idToChild.put(child.getId(), child);
```
joeatwork: We could probably skip children with id==-1

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +180,111 @@ public void accumulate(View found) {
...
+            }
+
             target.setLayoutParams(params);
         }
 
+        private boolean verifyLayout(View target, ArrayList<Integer> rules) {
+            ViewGroup parent = (ViewGroup) target.getParent();
+            SparseArray<View> idToChild = new SparseArray<View>();
+
+            for (int i = 0; i < parent.getChildCount(); i++) {
```
joeatwork: It's usually our style to cache childCount before the loop, so
```java
final int count = parent.getChildCount();
for (int i = 0; i < count; i++) {
    ...
}
```

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +180,111 @@ public void accumulate(View found) {
...
+                View child = parent.getChildAt(i);
+                idToChild.put(child.getId(), child);
+            }
+
+            ArrayMap<View, ArrayList<View>> dependencyGraph = new ArrayMap<View, ArrayList<View>>();
+            for (int i = 0; i < idToChild.size(); i++) {
```
joeatwork: Let's cache the size() before the loop, since it shouldn't change.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +180,111 @@ public void accumulate(View found) {
...
+
+                ArrayList<View> dependencies = new ArrayList<View>();
+                for (int rule : rules) {
+                    int dependencyId = layoutRules[rule];
+                    if (dependencyId > 0 && dependencyId != child.getId()) {
+                        dependencies.add(idToChild.get(dependencyId));
+                    }
+                }
+
+                dependencyGraph.put(child, dependencies);
```
joeatwork: Do we need dependencyGraph at all? Can we just check in place using the views themselves?

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
```diff
@@ -23,10 +25,15 @@
  */
 /* package */ class EditState extends UIThreadSet<Activity> {
 
-    public EditState() {
+    public interface EditErrorMessage {
```
joeatwork: Let's name this `OnErrorListener` and have the callback be named `OnError`. We don't use this convention very well in our code, but it's the platform convention for event listeners like this one.

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
yinfeiru: In my opinion, this is the best place to catch it. My argument is that if LayoutUpdateException happens, we can get rid of the edit, so that it won't exist in the queue any more. If we do it later, we'll still have the edit in the queue but having a flag to skip visit(), which is not ideal - 1. we don't need the edit anyways; 2. we are spending extra time on useless stuff

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
yinfeiru: How can this crash the app? I don't really get it.

joeatwork: Sorry, I think I was wrong about that. The exception will have to be caught above.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yinfeiru: The reason I'm doing this is because I want to have this exception for all the layout exceptions. And JSONObject is the best choice since it's flexible on its content. I'm expecting the linear layout update will have some sort of exception.

But, I'll check out the existing error message.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yinfeiru: I want to support all kinds of layout updates (linear, relative, viewgroup)

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: I didn't really know a good structure for this, so that I was following the one used in Android source code. I think the point is that ArrayMap is more memory efficient and since there won't be too many nodes in the graph, it's almost as fast as regular HashMap.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: I don't think there's any child has -1 as its -1. -1 is only used for parent layout options.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: I think -
1. the logic looks way clearer if we have dependencyGraph as a separate thing
2. it's possible that we have several unconnected graphs.. with dependencyGraph, we don't need to track if a child has been visited or not, since we can kick the visited node from the graph. Otherwise, we'll need a flag to track if a node has been visited, which looks ugly in my opinion.

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
joeatwork: Discussed, going with a different approach.

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
```diff
@@ -7,6 +7,8 @@
 import android.view.View;
 import android.view.ViewTreeObserver;
 
+import org.json.JSONObject;
```
joeatwork: You could probably get rid of this import now

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -651,6 +661,33 @@ private void sendReportTrackToEditor(String eventName) {
             }
         }
 
+        private void sendLayoutError(ViewVisitor.CantVisitException exception) {
```
joeatwork: Great!

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
joeatwork: What happens if we have a child with no explicitly assigned id? What does getId() return in that case (and can we skip those objects?)

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
joeatwork: Let's still declare the variable as a plain `Map<View, ArrayList<View>> dependencyGraph` once we construct it.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: Can we make it static? I think if you make it static, we can access it by ViewVisitor.LayoutUpdateVisitor.mHorizontalRules and this variable might not be initialized (unless we initialize the variable in the place where it's declared.

Make it a set could be a good idea.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: As we discussed, SparseArray is more memory efficient and is not much slower. And it is used in Android source code for the same purpose.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
joeatwork: We could make this static and initialize it when it's declared.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: I think the problem with Map is that this code will fail `View currentNode = dependencyGraph.keyAt(0);`. I'm grabbing the first element if the graph still has unparsed elements and Map doesn't support it.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: This guy is also called by cleanup. So, I think make it a separate function is more convenient.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
joeatwork: That is a completely good reason!

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: What is the advantage here? I think ArrayMap has all the functionality we need and this code is pretty straight forward. I don't know the advantage of having a class for this thing over doing it inline.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: 76fad9e

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: YOU ARE RIGHT!!! 5e4b721

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: 80a003b

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: 80a003b

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
yinfeiru: 80a003b

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
yinfeiru: 80a003b

src/main/java/com/mixpanel/android/viewcrawler/EditState.java
yinfeiru: I've re-done this logic. Please check out the latest commit.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -651,6 +661,33 @@ private void sendReportTrackToEditor(String eventName) {
...
+            final OutputStreamWriter writer = new OutputStreamWriter(out);
+            final JsonWriter j = new JsonWriter(writer);
+
+            try {
+                j.beginObject();
+                j.name("type").value("layout_error");
+                j.name("exception_type").value(exception.getExceptionType());
+                j.name("cid").value(exception.getName());
+                j.endObject();
+                j.flush();
```
joeatwork: I think this will flush automatically on close

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -131,7 +157,19 @@ public void cleanup() {
...
+                } catch (CantVisitException e) {
+                    ; // shouldn't reach here
+                }
+            }
+            mAlive = false;
+        }
+
+        @Override
+        public void visit(View rootView) {
+            if (mAlive) {
```
joeatwork: Great!

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -52,9 +52,14 @@ public CantGetEditAssetsException(String message, Throwable cause) {
         }
     }
 
-    public EditProtocol(ResourceIds resourceIds, ImageStore imageStore) {
+    public interface OnErrorListener {
```
joeatwork: OnErrorListener should probably be in ViewVisitor now, so we don't have a circular dependency

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -52,9 +52,14 @@ public CantGetEditAssetsException(String message, Throwable cause) {
...
+        public void sendErrorMessage(ViewVisitor.CantVisitException e);
```
joeatwork: By convention, this method should be named onError (to go with the classname, OnErrorListener)

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +190,120 @@ public void accumulate(View found) {
...
+
             target.setLayoutParams(params);
         }
 
+        private boolean verifyLayout(View target, Set<Integer> rules) {
+            ViewGroup parent = (ViewGroup) target.getParent();
+            SparseArray<View> idToChild = new SparseArray<View>();
+
+            int count = parent.getChildCount();
+            for (int i = 0; i < count; i++) {
```
joeatwork: Can there be circular dependencies on views that aren't siblings? That is, can a view refer to another view outside of it's parent when aligning?

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +190,120 @@ public void accumulate(View found) {
...
+            }
+
+            return hasCycle(dependencyGraph);
+        }
+
+        private boolean hasCycle(ArrayMap<View, ArrayList<View>> dependencyGraph) {
+            ArrayList<View> dfsStack = new ArrayList<View>();
+            while (!dependencyGraph.isEmpty()) {
+                View currentNode = dependencyGraph.keyAt(0);
+                if (!subGraphHasCycle(dependencyGraph, currentNode, dfsStack)) {
```
joeatwork: We may want to rename subgraphHasCycle, since in addition to returning a value it has a side effect. Perhaps `detectSubgraphCycle`? Something with a verb in it?

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: I don't think so. I've tried it before.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +190,120 @@ public void accumulate(View found) {
...
+            return true;
+        }
+
+        private boolean subGraphHasCycle(ArrayMap<View, ArrayList<View>> dependencyGraph,
+                                         View currentNode, ArrayList<View> dfsStack) {
+            if (dfsStack.contains(currentNode)) {
+                return false;
+            }
+
+            if (dependencyGraph.containsKey(currentNode)) {
```
joeatwork: I'm not sure I really follow this algorithm- that's why I'd like it to be in it's own file with it's own commentary. There is a lot going on here that isn't really about visiting views. In any case, a big comment describing the algorithm would probably be good.

joeatwork:
This is so cool!

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: I don't think we can do that. This is because EditProtocol is initialized in ViewCrawler and EditProtocol initializes Visitor. And because of that, to be able to pass OnErrorListener to Visitor, we need to pass an instance of ViewCrawler to EditProtocol.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
joeatwork: Great!

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
joeatwork: Yes, that's fine- but we should probably define the interface in in ViewCrawler- it's no problem for ViewCrawler to `implements ViewVisitor.OnErrorListener` - particularly since the interface revers to CantVisitException

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yinfeiru: 0122313

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: Yes, we can do that. But then EditProtocol will end up with having an instance of ViewCrawler instead of the onErrorListener interface, which gives EditProtocol too much power and looks bad.

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: b115603

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: renamed    27f7b06

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: YESSSSSSSSS. A comment is definitely necessary!!! 27f7b06

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
joeatwork: I'm not sure I understand. Let's talk about this.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -131,7 +157,19 @@ public void cleanup() {
...
+                    ; // shouldn't reach here
+                }
+            }
+            mAlive = false;
+        }
+
+        @Override
+        public void visit(View rootView) {
+            if (mAlive) {
+                super.getPathfinder().findTargetsInRoot(rootView, super.getPath(), this);
```
joeatwork: I don't think you need the `super` here, the methods will just resolve (and that way, if we ever override them it'll just work :) )

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -152,19 +190,127 @@ public void accumulate(View found) {
...
+
+            return hasCycle(dependencyGraph);
+        }
+
+        /**
+         * This function detects circular dependencies for all the views under the parent
+         * of the updated view. The basic idea is to consider the views as a directed
+         * graph and perform a DFS on all the nodes in the graph. If the current node is
+         * in the DFS stack already, there must be a circle in the graph. To speed up the
+         * search, all the parsed nodes will be removed from the graph.
```
joeatwork: Great! Thanks for the comment!

joeatwork: It still might be nice to group all of these interdependent methods together, so we can see what the comment refers to (not just hasCycle but also detectSubgraphCycle, etc) and maybe to keep those methods private to the algorithm.

joeatwork:
SO GOOD! I think we're almost there!

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: How to group them? Are you talking about having a class or what? I totally agree and that's definitely something I need to learn from you.  :)

joeatwork:
```java

{
public LayoutUpdateVisitor(...) {
      mCycleDetector = new CycleDetector()
}

mCycleDetector.hasCycle(...);

// Awesome comment here
private static class CycleDetector {
    public boolean hasCycle(....) { ... }
    private boolean detectSubgraphCycle(...) { ... }
}
}
```

yinfeiru: f38c8ac

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: 906e8a2

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -120,10 +146,60 @@ protected String name() {
     }
 
     public static class LayoutUpdateVisitor extends ViewVisitor {
-        public LayoutUpdateVisitor(List<Pathfinder.PathElement> path, LayoutRule args) {
+        private class CycleDetector {
```
joeatwork: This could be static (and we usually lay out our files)
```
public static members

public members (rare!)

public inner classes

public constructors

public methods

private methods

private classes

private members

private static members
```

joeatwork:
Looks great! I'm ok to merge this!

Awesome job on a hard problem!

yinfeiru:
Coolio! I'll merge it now!

PR #208, opened 2015-04-23
Layout exception polish

yinfeiru:
layout changes are performed on the parent


src/androidTest/java/com/mixpanel/android/viewcrawler/EditProtocolTest.java
```diff
@@ -25,8 +25,8 @@
 
 
 public class EditProtocolTest extends AndroidTestCase {
-    private static class MockOnErrorListener implements ViewVisitor.OnErrorListener {
-        public MockOnErrorListener() {
+    private static class MockOnLayoutErrorListener implements ViewVisitor.OnLayoutErrorListener {
```
joeatwork: We should probably change the name of the `onError` method to `onLayoutError` in this case.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -204,73 +204,42 @@ public LayoutUpdateVisitor(List<Pathfinder.PathElement> path, LayoutRule args,
...
+            // TODO find a way to optimize this.. remove this visitor and trigger a re-layout??
+            for (Map.Entry<View, int[]> original:mOriginalValues.entrySet()) {
                 final View changedView = original.getKey();
-                final LayoutRule originalValue = original.getValue();
-                try {
-                    setLayout(changedView, originalValue.verb, originalValue.anchor);
-                } catch (CantVisitException e) {
-                    ; // shouldn't reach here
+                final int[] originalValue = original.getValue();
+                RelativeLayout.LayoutParams originalParams = (RelativeLayout.LayoutParams) changedView.getLayoutParams();
```
joeatwork: Let's declare `originalParams` final (since it's only assigned once in the block)

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -204,73 +204,42 @@ public LayoutUpdateVisitor(List<Pathfinder.PathElement> path, LayoutRule args,
...
-            if (mOriginalValues.containsKey(found)) {
-                ; // Cache exactly one
-            } else {
-                LayoutRule originalValue = new LayoutRule(newVerb, currentRules[newVerb]);
-                mOriginalValues.put(found, originalValue);
-            }
-
             try {
-                setLayout(found, newVerb, newAnchorId);
+                setLayout(found);
```
joeatwork: This is nicely simpler!

joeatwork:
Something to be aware of - this changes the protocol for folks who already have a beta version of the library installed.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -38,7 +38,7 @@
         public void OnEvent(View host, String eventName, boolean debounce);
```
joeatwork: Can we make CycleDetector a static class in this PRQ? (It may have to move out of LayoutUpdateVisitor but it can still be private inside of the ViewVisitor class)

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -204,73 +204,42 @@ public LayoutUpdateVisitor(List<Pathfinder.PathElement> path, LayoutRule args,
 
         @Override
         public void cleanup() {
-            for (Map.Entry<View, LayoutRule> original:mOriginalValues.entrySet()) {
+            // TODO find a way to optimize this.. remove this visitor and trigger a re-layout??
```
joeatwork: You can probably remove this TODO - It looks like this cleanup is ok just as it is :)

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -204,73 +204,42 @@ public LayoutUpdateVisitor(List<Pathfinder.PathElement> path, LayoutRule args,
...
-            if (rules != null && !verifyLayout(target, rules)) {
-                throw new CantVisitException("Circular dependency detected!", "circular_dependency", mName);
-            }
-
-            target.setLayoutParams(params);
-        }
-
-        private boolean verifyLayout(View target, Set<Integer> rules) {
-            ViewGroup parent = (ViewGroup) target.getParent();
+        private void setLayout(View target) throws CantVisitException {
```
joeatwork: This could probably be inlined, since it is only called once and takes no arguments.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -282,6 +251,51 @@ private boolean verifyLayout(View target, Set<Integer> rules) {
                 }
             }
 
+            int size = mArgs.size();
+            for (int i = 0; i < size; i++) {
+                LayoutRule layoutRule = mArgs.get(i);
+                final int viewId = layoutRule.viewId;
```
joeatwork: We may not need the `viewId`, `newVerb`, and `newAnchorId` variables, since you can just refer directly to the members of `layoutRule`

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -282,6 +251,51 @@ private boolean verifyLayout(View target, Set<Integer> rules) {
...
+                final int newVerb = layoutRule.verb;
+                final int newAnchorId = layoutRule.anchor;
+                final View currentNode = idToChild.get(viewId);
+
+                // typecast ViewGroup.LayoutParams is needed for API 18 and lower
```
joeatwork: I'm not sure I understand this comment. What does it refer to?

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -282,6 +251,51 @@ private boolean verifyLayout(View target, Set<Integer> rules) {
...
+                final RelativeLayout.LayoutParams currentParams = (RelativeLayout.LayoutParams) currentNode.getLayoutParams();
+                final int[] currentRules = currentParams.getRules().clone();
+
+                if (currentRules[newVerb] == newAnchorId) {
+                    continue;
+                }
+
+                if (mOriginalValues.containsKey(currentNode)) {
+                    ; // Cache exactly one
```
joeatwork: We should probably update this comment to "Cache exactly one value per child"

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -282,6 +251,51 @@ private boolean verifyLayout(View target, Set<Integer> rules) {
                 }
             }
 
+            int size = mArgs.size();
+            for (int i = 0; i < size; i++) {
+                LayoutRule layoutRule = mArgs.get(i);
+                final int viewId = layoutRule.viewId;
+                final int newVerb = layoutRule.verb;
+                final int newAnchorId = layoutRule.anchor;
+                final View currentNode = idToChild.get(viewId);
```
joeatwork: Let's change `currentNode` to `currentChild` or `currentView`

yinfeiru:
Thanks for review it Joe. But I'm actually still working on it.. I have a really bad habit of getting used to review the changes on github.. :(

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -282,6 +251,51 @@ private boolean verifyLayout(View target, Set<Integer> rules) {
...
+                    continue;
+                }
+
+                if (mOriginalValues.containsKey(currentNode)) {
+                    ; // Cache exactly one
+                } else {
+                    mOriginalValues.put(currentNode, currentRules);
+                }
+
+                RelativeLayout.LayoutParams newParams = (RelativeLayout.LayoutParams)currentNode.getLayoutParams();
```
joeatwork: Can we use `currentParams` above for this?

joeatwork:
Terrific work! (Sorry about prematurely reviewing it!)

yinfeiru:
No, I totally appreciate it!! I'm just feeling guilty about wasting your time on the premature code!

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: I've done it. But why do you want to make it a static class in ViewVisitor? I think it's unique to LayoutUpdateVisitor.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: If I understand it correctly, we are not actually changing the attributes of the views in the app, but overwrite them. So, my question is, why do we even write the original value back if the attributes are not changed? Why don't we just trigger the window refresh in some way and leave the attributes as they are?

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: In Javascript, we always try to assign a value in a dictionary to a variable if such thing is used more than once. In Java, do we have a different pattern?

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: I'll remove this.. This is just a stupid Android thing. You have NO way to make a copy of RelativeLayout.LayoutParams.. I wanted to have RelativeLayout.LayoutParams as the original value so that I can simply call view.setParams in the cleanup to revert the changes. But you can only get a reference to it.

src/androidTest/java/com/mixpanel/android/viewcrawler/EditProtocolTest.java
yinfeiru: ebdda55

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: I'll keep this guy non-final as I get rid of `newParams`, which is a redundant thing, and use this guy for updating the view.

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: d568aca

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: 636a1ad

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: f80d1ac

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: a9e14fa

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: eaec668

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: eaec668

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
yinfeiru: db16882

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -145,143 +144,143 @@ protected String name() {
...
+                }
 
-            target.setLayoutParams(params);
-        }
+                currentParams.addRule(layoutRule.verb, layoutRule.anchor);
 
-        private boolean verifyLayout(View target, Set<Integer> rules) {
-            ViewGroup parent = (ViewGroup) target.getParent();
-            SparseArray<View> idToChild = new SparseArray<View>();
+                final Set<Integer> rules;
```
yinfeiru: 3d28ecb

It now validates the rules for every single change. This is to prevent weird bugs, like the user changes their implementation and the old layout changes don't work anymore.

joeatwork:
I'm going to merge this! Hurray!

yinfeiru:
COOL!!!

PR #209, opened 2015-04-23
Interpret Colordrawable objects

joeatwork:
A special case to accomodate background colors.

### DIFF
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
@@ -3,6 +3,7 @@ package com.mixpanel.android.viewcrawler;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.drawable.BitmapDrawable;
+import android.graphics.drawable.ColorDrawable;
 import android.graphics.drawable.Drawable;
 import android.util.Log;
 import android.util.Pair;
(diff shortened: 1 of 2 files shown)


src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -342,7 +343,13 @@ private Object convertArgument(Object jsonArgument, String type, List<String> as
             } else if ("float".equals(type) || "java.lang.Float".equals(type)) {
                 return ((Number) jsonArgument).floatValue();
             } else if ("android.graphics.drawable.Drawable".equals(type)) {
```
yinfeiru: can you do
`else if ("android.graphics.drawable.Drawable".equals(type) || "android.graphics.drawable.BitmapDrawable".equals(type)) {...}`

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -351,11 +358,11 @@ private Object convertArgument(Object jsonArgument, String type, List<String> as
         }
     }
 
-    private Drawable readDrawable(JSONObject description, List<String> assetsLoaded)
```
yinfeiru: Why do you change readDrawable to readBitmapDrawable?

joeatwork: I think choosing to interpret Drawable as a particular subclass was a mistake I made in the earlier protocol- the UI is smart enough to pick a concrete class (like BitmapDrawable or ColorDrawable) and it means that if we product NinepatchDrawables or ScaleDrawables or whatever in the future, it'll be a bit less confusing.

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
joeatwork: I wanted a good spot for the explanatory comment below- "This case is weird", as separate from the BitmapDrawable case.

yinfeiru: sounds good

yinfeiru:
Looks good to me

PR #210, opened 2015-04-24
Include the gradle wrapper in the repo since that's how it's meant to be distributed

marshallpierce:
Yes, even though it's a jar, it's really supposed to be in VCS! See the second paragraph of  http://gradle.org/docs/current/userguide/gradle_wrapper.html :)


joeatwork:
Right on! This will make things a lot easier to develop - thanks for the tip!

screen shot 2015-04-24 at 9 20 48 am

yinfeiru:
Thanks for letting us know! This is really valuable. I'll verify the whole thing and include it in the next release!

PR #213, opened 2015-04-30
Protocol id name instead

yinfeiru:
Use id name instead of id to make it more reliable.
I'm use "0" (for removal) and "-1" (for parent) because we have 2 special cases here and I don't want to have 2 more keys in the protocol.

### DIFF
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
@@ -158,14 +158,14 @@ import java.util.List;
 
                     final String view_id_name = layout_info.getString("view_id_name");
                     final String anchor_id_name = layout_info.getString("anchor_id_name");
-                    final Integer view_id = reconcileIdsInPath(-1, view_id_name, mResourceIds);
+                    final Integer view_id = reconcileIds(-1, view_id_name, mResourceIds);
                     final Integer anchor_id;
                     if (anchor_id_name.equals("0")) {
                         anchor_id = 0;
                     } else if (anchor_id_name.equals("-1")) {
                         anchor_id = RelativeLayout.TRUE;
                     } else {
-                        anchor_id = reconcileIdsInPath(-1, anchor_id_name, mResourceIds);
+                        anchor_id = reconcileIds(-1, anchor_id_name, mResourceIds);
                     }
 
                     if (view_id == null || anchor_id == null) {
@@ -271,7 +271,7 @@ import java.util.List;
 
             final int targetId;
 
-            final Integer targetIdOrNull = reconcileIdsInPath(targetExplicitId, targetIdName, idNameToId);
+            final Integer targetIdOrNull = reconcileIds(targetExplicitId, targetIdName, idNameToId);
             if (null == targetIdOrNull) {
                 return NEVER_MATCH_PATH;
             } else {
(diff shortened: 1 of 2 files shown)


src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -155,8 +155,22 @@ public Edit readEdit(JSONObject source) throws BadInstructionsException, CantGet
                 for (int i = 0; i < length; i++) {
                     JSONObject layout_info = args.optJSONObject(i);
                     ViewVisitor.LayoutRule params;
-                    params = new ViewVisitor.LayoutRule(layout_info.getInt("view_id"),
-                            layout_info.getInt("verb"), layout_info.getInt("anchor_id"));
+
+                    final Integer view_id = reconcileIdsInPath(-1, layout_info.getString("view_id"), mResourceIds);
```
joeatwork: Let's change the name of "view_id" to something like "view_id_name" if we're going to interpret it as a name. We should also probably change the name of `reconcileIdsInPath` since this call isn't in the path interpretation code.

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -155,8 +155,22 @@ public Edit readEdit(JSONObject source) throws BadInstructionsException, CantGet
...
+                    final Integer anchor_id;
+                    if (layout_info.getString("anchor_id").equals("0")) {
+                        anchor_id = 0;
+                    } else if (layout_info.getString("anchor_id").equals("-1")) {
+                        anchor_id = RelativeLayout.TRUE;
+                    } else {
+                        anchor_id = reconcileIdsInPath(-1, layout_info.getString("anchor_id"), mResourceIds);
```
joeatwork: We should also consider changing `"anchor_id"` to `"anchor_id_name"`

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -155,8 +155,22 @@ public Edit readEdit(JSONObject source) throws BadInstructionsException, CantGet
...
+                    }
+
+                    if (view_id == null || anchor_id == null) {
+                        continue;
```
joeatwork: This seems like an invalid message. While it probably isn't right to throw an exception (since we could have id names specified that we can't find for some reason, like multiple apps or versions reading the same decide messages) we should probably log when we're missing information, or we're getting directives for views that don't seem to exist.

joeatwork:
This looks *great*!

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: 5c31491 good point

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: e8fdca7

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
yinfeiru: e8fdca7 for view_id_name
4e924e8 for reconcileIds

joeatwork:
Fantastic! Merge this!

PR #214, opened 2015-05-01
New layout test suite

yinfeiru:
OMG.. Finally get a chance to take a step back and re-do the layout test. Feel much better with it now!!

1. `testLayoutEdit` in `EditProtocolTest` now only takes care of layout protocol - `"-1"` for adding parent layout, `"0"` for removing a layout, other id name for adding anchor layout, ignoring non-existing view
2. `TestView` has 2 tests now - one for testing the basic operations, the other for testing circular dependency detection


joeatwork:
Great! I'm good to go with this!

PR #216, opened 2015-05-05
Accept map for tracking api

yinfeiru:
Add the support of Map

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -680,6 +680,15 @@ public class MixpanelAPI {
         public void set(String propertyName, Object value);
 
         /**
+         * Map version of set API below.
+         *
+         * @param properties a Map containing the collection of properties you wish to apply
+         *      to the identified user.Each key in the Map will be associated with
+         *      a property name, and the value of that key will be assigned to the property.
+         */
+        public void setMap(Map<String, Object> properties);
+
+        /**
          * Set a collection of properties on the identified user all at once.
          *
          * @param properties a JSONObject containing the collection of properties you wish to apply
@@ -695,6 +704,15 @@ public class MixpanelAPI {
          * @param value The value of the Mixpanel property. For "Zip Code", this value might be the String "90210"
          */
         public void setOnce(String propertyName, Object value);
+
+        /**
+         * Map version of setOnce API below
+         *
+         * @param properties a Map containing the collection of properties you wish to apply
+         *      to the identified user. Each key in the Map will be associated with
+         *      a property name, and the value of that key will be assigned to the property.
+         */
+        public void setOnceMap(Map<String, Object> properties);
 
         /**
          * Like {@link People#set(String, Object)}, but will not set properties that already exist on a record.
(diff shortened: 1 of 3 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -330,6 +330,21 @@ public void timeEvent(final String eventName) {
     }
 
     /**
+     * Map version of the track API below.
```
joeatwork: Let's make this comment a stand-alone message, with an `{@link }` to the JSON version of the API.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -330,6 +330,21 @@ public void timeEvent(final String eventName) {
...
+     *
+     * @param eventName The name of the event to send.
+     * @param properties A Map containing the key value pairs of the properties to include in this event.
+     *                   Pass null if no extra properties exist. Cannot be null.
+     */
+    public void trackMap(String eventName, Map<String, Object> properties) {
+        if (null == properties) {
```
joeatwork: No need to limit null here?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -330,6 +330,21 @@ public void timeEvent(final String eventName) {
...
+            Log.e(LOGTAG, "trackMap does not accept null properties");
+            return;
+        }
+        track(eventName, new JSONObject(properties));
```
joeatwork: can `new JSONObject(properties)` ever throw an error?

joeatwork:
This is very exciting! Let's write some tests where we use screwy maps (like maps from strings to InputStreams, or other crazy things) and see what happens :)

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: it might throw NullPointerException, but I guess we want to crash the app in that case?

joeatwork: No, I think we should handle this - since `track(eventName, null)` is legal, I think we should call it that way.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: I do believe we need to handle null here since the JSONObject constructor doesn't handle it
image

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: oh.. and we don't need to mention `xxx API below`

yinfeiru: 3406e52

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: 3406e52

You are right!

joeatwork:
I love it! Let's merge this into v4.5.4!

PR #217, opened 2015-05-07
Acceptance test for layout

yinfeiru:
Looks like there's no good way for us to retrieve the relative layout rules directly (unless I spend several days on reading the selenium source code). As a workaround, I compare the location between different view to make sure they are properly positioned. It's the best way I can think of.


joeatwork:
THIS IS FANTASTIC!

I think comparing their positions is ok.

I'd like two more tests (but they're simple)
- Lets send an update with a crazy verb
- Lets send a variant that claims to be a layout change but is missing anchor and/or verb

This is SO COOL!

yinfeiru:
These two tests definitely provide us more coverage on the potential edge cases, but I don't think they are necessary here. What we are testing here is our layout protocol, which is private and can be used only by us. I don't think the users are capable of creating these two errors. These two edge cases are totally different from absent view test, which can be introduced by the users.

In short, I think this test is a end-to-end test which should cover special usage, but not all the edge cases which can only be generated by buggy code which is written by user.

joeatwork:
That makes sense. I proposed the the crazy tests because we may have bugs in the UI in the future (or even change the protocol, or have the platform change out from under us in future Android releases) and we rarely test how changes that come in through decide work. I want to be sure that if we make changes that will cause crashes in the wild, we catch them. Perhaps we could just write unit tests for those cases?

joeatwork:
Anyway, this is better than what we have - Let's merge it :)!

PR #218, opened 2015-05-11
Dimensions are optional because we can't always know them on the client

joeatwork:
The UI infers dimensions from the existing image - when no image is present, we won't get inbound dimensions (which is ok)

### DIFF
--- a/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
+++ b/src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
@@ -365,11 +365,23 @@ import java.util.List;
             }
 
             final String url = description.getString("url");
-            final JSONObject dimensions = description.getJSONObject("dimensions");
-            final int left = dimensions.getInt("left");
-            final int right = dimensions.getInt("right");
-            final int top = dimensions.getInt("top");
-            final int bottom = dimensions.getInt("bottom");
+
+            final boolean useBounds;
+            final int left;
+            final int right;
+            final int top;
+            final int bottom;
+            if (description.isNull("dimensions")) {
+                left = right = top = bottom = 0;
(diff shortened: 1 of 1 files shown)


yinfeiru:
It looks good. If I understand it correctly, the snapshot logic will send null as "dimensions" when it can't get such info. And that's why we need to handle the null case in EditProtocol here, right? 
Do we need to handle such case in Web UI? What is the current behavior?
This might not related to this pull request, but are we able to display the image properly on different screen sizes? Eg. for the same image, is it possible that the value read from a large screen device is larger than a smaller screen device?

joeatwork:
That's correct. The current behavior is a crash in the Web UI, and (when that is fixed) sending null dimensions back to the device, where the value is ignored with a parse error.

In general, we're ok with different sizes of images because we borrow the absolute dimensions as they're shown. In future releases, we may want to use dp or something, but I don't think it'll make it in this release.

src/main/java/com/mixpanel/android/viewcrawler/EditProtocol.java
```diff
@@ -380,7 +392,10 @@ private Drawable readBitmapDrawable(JSONObject description, List<String> assetsL
             }
 
             final Drawable ret = new BitmapDrawable(Resources.getSystem(), image);
-            ret.setBounds(left, top, right, bottom);
+            if (useBounds) {
+                ret.setBounds(left, top, right, bottom);
```
joeatwork: I agree, but I wanted to catch JS errors before I pulled and cached the image. The call to getImage does a network request, and writes to disk, so I wanted to throw a JSON error before that happened.

yinfeiru: I see. That's legit.

PR #220, opened 2015-05-19
Consider UX when applying Experiments

joeatwork:
Experiments will (by default) only be applied when an activity resumes.
Users of the library can also call joinExperimentIfAvailable to join
whenever they feel the UX is appropriate.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -990,6 +990,23 @@ public class MixpanelAPI {
         public void showNotificationIfAvailable(Activity parent);
 
         /**
+         * Applies A/B test changes, if they are present. By default, your application will attempt
+         * to join available experiments any time an activity is resumed, but you can disable this
+         * automatic behavior by adding the following tag to the &lt;application&gt; tag in your AndroidManifest.xml
+         * {@code
+         *     <meta-data android:name="com.mixpanel.android.MPConfig.AutoShowMixpanelUpdates"
+         *                android:value="false" />
+         * }
+         *
+         * If you disable AutoShowMixpanelUpdates, you'll need to call joinExperimentIfAvailable to
+         * join or clear existing experiments. If you want to display a loading screen or otherwise
+         * wait for experiments to load from the server before you apply them, you can use
+         * {@link #addOnMixpanelUpdatesReceivedListener(OnMixpanelUpdatesReceivedListener)} to
+         * be informed that new experiments are ready.
+         */
+        public void joinExperimentIfAvailable();
+
+        /**
          * Shows the given in app notification to the user. Display will occur just as if the
          * notification was shown via showNotificationIfAvailable. In most cases, it is
          * easier and more efficient to use showNotificationIfAvailable.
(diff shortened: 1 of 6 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelActivityLifecycleCallbacks.java
```diff
@@ -36,7 +35,9 @@ public void onActivityDestroyed(Activity activity) { }
     public void onActivitySaveInstanceState(Activity activity, Bundle outState) { }
 
     @Override
-    public void onActivityResumed(Activity activity) { }
+    public void onActivityResumed(Activity activity) {
```
yinfeiru: If AutoShowMixpanelUpdates is set to false, this function won't be registered. Am I right?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -990,6 +990,23 @@ public void reset() {
         public void showNotificationIfAvailable(Activity parent);
 
         /**
+         * Applies A/B test changes, if they are present. By default, your application will attempt
+         * to join available experiments any time an activity is resumed, but you can disable this
```
yinfeiru: I know I can test it by myself. But, if I have multiple experiments under the same project, will all the changes be loaded to the app? I guess so though..

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1499,6 +1518,14 @@ public void trackNotification(final String eventName, final InAppNotification no
         }
 
         @Override
+        public void joinExperimentIfAvailable() {
+            final JSONArray variants = mDecideMessages.getVariants();
```
yinfeiru: I'm wondering if we want to call it `variants` instead of `changes`. We use `variants` for describing different A/B test versions on web UI.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1499,6 +1518,14 @@ public void trackNotification(final String eventName, final InAppNotification no
...
+            if (null != variants) {
+                mUpdatesFromMixpanel.setVariants(variants);
```
yinfeiru: Event `setVariants` looks weird to me. We are not loading all the different versions to the app, are we?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1082,17 +1099,19 @@ public void reset() {
 
         /**
          * Adds a new listener that will receive a callback when new updates from Mixpanel
-         * (like surveys or in app notifications) are discovered.
+         * (like surveys, in app notifications, or A/B test experiments) are discovered. Most users of the library
```
yinfeiru: Do you mind to change it to `in-app` to make it consistent?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1082,17 +1099,19 @@ public void reset() {
...
          * <p>The given listener will be called when a new batch of updates is detected. Handlers
          * should be prepared to handle the callback on an arbitrary thread.
          *
-         * <p>Once this listener is called, you may call {@link People#getSurveyIfAvailable()}
-         * or {@link People#getNotificationIfAvailable()}
-         * to retrieve a Survey or InAppNotification object. However, if you have multiple
-         * listeners registered, one listener may have consumed the available Survey or
-         * InAppNotification, and so the other listeners may obtain null when calling
-         * {@link People#getSurveyIfAvailable()} or {@link People#getNotificationIfAvailable()}.
+         * <p>The listener will be called when new surveys, in app notifications, or experiments
```
yinfeiru: dito

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
joeatwork: That's true - the user gets one variant from every experiment, but there could be multiple experiments in progress, and this JSONArray will contain changes for all of them. (I'm using `variants` because it's used in the API)

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
joeatwork: We could be loading multiple variants, each from it's own experiment

src/main/java/com/mixpanel/android/mpmetrics/MixpanelActivityLifecycleCallbacks.java
joeatwork: That is true. This callback will never happen if AutoShowMixpanelUpdates is set to false, and the user will need to call `joinExperimentIfAvailable` themselves. The intent is to allow folks to install their own loading screens or otherwise customize the UX if they'd like.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
joeatwork: Yes, they will.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
joeatwork: Fixed in 3da3f72

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
joeatwork: Fixed in 3da3f72

PR #222, opened 2015-05-20
Fix simple/trivial lint issues

joeatwork:
Fix simple lint warnings in advance of release. Fixes logtag length issues, introduces issues with deprecated HTTP tools by upping build target version to 22.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,14 +12,14 @@ buildscript {
 }
 
 android {
-    compileSdkVersion 22
+    compileSdkVersion 21
     buildToolsVersion "22.0.1"
 
     defaultConfig {
         // If you change minSdkVersion or targetSdkVersion below, you must also change
         // the relevant attributes of the <uses-sdk> tag in AndroidManifest.xml
         minSdkVersion 9
-        targetSdkVersion 22
+        targetSdkVersion 21
         testApplicationId "com.mixpanel.android.mpmetrics"
         testInstrumentationRunner "android.test.InstrumentationTestRunner"
 
--- a/build.gradle
+++ b/build.gradle
@@ -12,20 +12,21 @@ buildscript {
 }
 
 android {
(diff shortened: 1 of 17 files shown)


src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -153,10 +153,10 @@ protected String name() {
          * in the DFS stack already, there must be a circle in the graph. To speed up the
          * search, all the parsed nodes will be removed from the graph.
          */
-        public boolean hasCycle(ArrayMap<View, ArrayList<View>> dependencyGraph) {
-            ArrayList<View> dfsStack = new ArrayList<View>();
+        public boolean hasCycle(TreeMap<View, List<View>> dependencyGraph) {
```
yinfeiru: why TreeMap?

yinfeiru:
Looks great.
One thing I'm not sure of is that if we want to bump the target version to 22. According to Android official site, only 0.7% of the users are on Lollipop 5.1 (the original `target version 21` also looks too critical to me).
http://developer.android.com/about/dashboards/index.html

Can you please give me more insights here?

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
joeatwork: ArrayMap isn't available in API 16, and I needed another map with a concept of "first element"

build.gradle
```diff
@@ -12,20 +12,21 @@ buildscript {
 }
 
 android {
-    compileSdkVersion 21
-    buildToolsVersion "21.1.2"
+    compileSdkVersion 22
```
joeatwork: Discussed, let's hold on changing the target version until after the release.

PR #223, opened 2015-05-29
return double quotes on timeout exception

yinfeiru:


src/main/java/com/mixpanel/android/viewcrawler/ViewSnapshot.java
```diff
@@ -114,6 +115,8 @@ public void snapshots(UIThreadSet<Activity> liveActivities, OutputStream out) th
                 Log.e(LOGTAG, "Exception thrown during screenshot attempt", e);
             }
         }
+        writer.write("\"\"");
+        writer.flush();
```
joeatwork: This makes me nervous. What exception are we throwing? (I'd also prefer if the control flow was to emit the quotes in an exception handler than having the return inside of the try block.

yinfeiru: maybe move it into the `catch (final TimeoutException e)` block to make it more specific?

joeatwork: That would be a good idea.

yinfeiru: 964044c

src/main/java/com/mixpanel/android/viewcrawler/ViewSnapshot.java
```diff
@@ -109,6 +109,8 @@ public void snapshots(UIThreadSet<Activity> liveActivities, OutputStream out) th
             if (MPConfig.DEBUG) {
                 Log.i(LOGTAG, "Screenshot took more than 1 second to be scheduled and executed. No screenshot will be sent.", e);
             }
+            writer.write("\"\"");
```
joeatwork: I'm not sure this is actually what we want to return in this case. What does the output look like here? Should this be the empty list rather than an empty string?

yinfeiru: technically, it doesn't matter. BUT, YOU ARE RIGHT! I should follow the same snapshot format as if the app is inactive.

yinfeiru: good? 7895a8c

joeatwork:
Do we need to do the same for InterruptedException?

yinfeiru:
I'm not sure. But, I think it'd be safer to make it specific. The current problem we are solving is the timeout issue which happens when the user pulls the app into the foreground.

Do you know what triggers InterruptedException here?

joeatwork:
In theory, it's if something interrupts the sleeping thread while we're waiting for the future to complete. It's unlikely but possible.

My thought is, if we send something on failure, we should always send the same thing on failure.

yinfeiru:
NICE SUGGESTION! I'm now much happier with it! Thanks!!!

joeatwork:
I love this! If you're happy with your tests, please merge it to v4.5.4dev

PR #225, opened 2015-06-04
NOT COMPLETE Static tweaks interface

joeatwork:
This is just for API review. We'll need to teach both sides of the conversation to transmit updates as new tweaks appear for this scheme to work, since even if the declarations are static they still get lazy loaded (for example, when an activity appears for the first time).


joeatwork:
Wrong branch

PR #226, opened 2015-06-04
Static tweaks interface

joeatwork:
This PRQ is for comment on the API, we still need to update the lib to actively inform the editor when new tweaks are declared (since even static declarations may run late if references to the enclosing class are needed late, for example if an activity isn't shown for a while)

Also, still needs tests and docs.

@jbwyme @arittr please take a look and comment


joeatwork:
This prq breaks mixpanel-android-build

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -77,164 +41,204 @@
...
+        public TweakValue updateValue(Object newValue) {
+            return new TweakValue(type, defaultValue, minimum, maximum, newValue);
+        }
+
+        public String getStringValue() {
+            String ret = null;
+
+            try {
+                ret = (String) defaultValue;
+            } catch (ClassCastException e2) {
```
yinfeiru: should be `ClassCastException e`

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -77,164 +41,204 @@
...
+                ret = (String) value;
+            } catch (ClassCastException e) {
+                ; // ok
+            }
+
+            return ret;
+        }
+
+        public Number getNumberValue() {
+            Number ret = 0;
```
yinfeiru: if both of defaultValue and value are null, we are returning 0.. Is that expected? to me, this behavior is wrong since 0 is not acting like the default value.

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -77,164 +41,204 @@
...
+                } catch (ClassCastException e) {
+                    ; // ok
+                }
+            }
+
+            return ret;
+        }
+
+        public Boolean getBooleanValue() {
+            Boolean ret = false;
```
yinfeiru: same here.

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
joeatwork: Fixed in a9d2fd0

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
joeatwork: You're right, this is a bit defensive, but we don't want to ever return null for non-string tweaks. Folks will be expecting primitive values, and won't be on the lookout for NPEs.

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
joeatwork: Same thing- no NPEs

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2043,5 +2086,10 @@ private static void checkIntentForInboundAppLink(Context context) {
     // Maps each token to a singleton MixpanelAPI instance
     private static final Map<String, Map<Context, MixpanelAPI>> sInstanceMap = new HashMap<String, Map<Context, MixpanelAPI>>();
     private static final SharedPreferencesLoader sPrefsLoader = new SharedPreferencesLoader();
+    private static final Tweaks sDefaultTweaks = new Tweaks();
```
yinfeiru: why do you call this `default tweaks`? where is the `non-default tweaks`?

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -1,68 +1,71 @@
...
+     * Returns the descriptions of all tweaks that currently exist.
+     *
+     * The Mixpanel library uses this method internally to expose tweaks and their types to the UI. Most
+     * users will not need to call this method directly.
+     */
+    public synchronized Map<String, TweakValue> getAllValues() {
+        return new HashMap<String, TweakValue>(mTweakValues);
+    }
 
     @IntDef({
```
yinfeiru: I'm pretty sure you have a good theory here. But, why do you only have 4 types in this guy? Are `int`, `short` and `byte` covered by `long`? I don't know the reason why you don't make it specific here, since we already did it in other places.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -518,38 +534,33 @@ private void sendDeviceInfo() {
...
-                        final Tweaks.TweakDescription desc = tweak.getValue();
+                    for (Map.Entry<String, Tweaks.TweakValue> tweak:tweakDescs.entrySet()) {
+                        final Tweaks.TweakValue desc = tweak.getValue();
                         final String tweakName = tweak.getKey();
                         j.beginObject();
                         j.name("name").value(tweakName);
-                        j.name("minimum").value(desc.minimum);
-                        j.name("maximum").value(desc.maximum);
+                        j.name("minimum").value((Number) null);
+                        j.name("maximum").value((Number) null);
```
yinfeiru: I'm a bit confused.. Why are `minimum` and `maximum` both `null`?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -518,38 +534,33 @@ private void sendDeviceInfo() {
...
                             case Tweaks.LONG_TYPE:
                                 j.name("type").value("number");
                                 j.name("encoding").value("l");
-                                j.name("value").value(mTweaks.getLong(tweakName));
+                                j.name("value").value(desc.getNumberValue().longValue());
                                 break;
                             case Tweaks.STRING_TYPE:
-                                j.name("type").value("string");
-                                j.name("value").value(mTweaks.getString(tweakName));
+                                j.name("type").value("unknown");
```
yinfeiru: Why is this guy `unknown`? I guess you are using it as a template and convert this `"value"` on the web UI. But, is web UI able to handle it? How can web UI tell wether it's a real string or something need to be converted?

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
```diff
@@ -284,7 +285,21 @@ public void accumulate(View found) {
         }
 
         private boolean verifyLayout(Set<Integer> rules, SparseArray<View> idToChild) {
-            final TreeMap<View, List<View>> dependencyGraph = new TreeMap<View, List<View>>();
+            // We don't really care about the order, as long as it's always the same.
+            final TreeMap<View, List<View>> dependencyGraph = new TreeMap<View, List<View>>(new Comparator<View>() {
+                @Override
+                public int compare(final View lhs, final View rhs) {
```
yinfeiru: how does this thing work? we can have an offline chat if you want.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
joeatwork: Good call! Fixed in b854bcf

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
joeatwork: The Web UI really only knows four types - integral, floating point, boolean, and text. We provide a `byte` type to the user but under the hood it doesn't really matter (and we overflow/truncate anyhow, since we get JSON values)

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: We don't support them in the api right now, so they'll always be null anyway, and I wanted to reduce the public API of `Tweaks` to the smallest thing I could before the launch.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
joeatwork: That's a bug! Fixed in 3419369

src/main/java/com/mixpanel/android/viewcrawler/ViewVisitor.java
joeatwork: The treemap needs an ordering - we use it to grab the "next" view (just like your arraymap) - it turns out that Views don't have a natural ordering, so I just gave it an arbitrary one.

PR #228, opened 2015-06-05
Send events id to decide if we don't have a people id

joeatwork:


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -223,8 +223,11 @@
 
         // TODO reading persistent identify immediately forces the lazy load of the preferences, and defeats the
         // purpose of PersistentIdentity's laziness.
-        final String peopleId = mPersistentIdentity.getPeopleDistinctId();
-        mDecideMessages.setDistinctId(peopleId);
+        String decideId = mPersistentIdentity.getPeopleDistinctId();
+        if (null == decideId) {
```
yinfeiru: I don't think I fully understand what's going on here.. Why do you use events distinct id as the decideId? Doesn't it change the behavior of this whole thing?

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelBasicTest.java
```diff
@@ -412,9 +412,9 @@ public int addJSON(JSONObject message, MPDbAdapter.Table table) {
                 final boolean isIdentified = isIdentifiedRef.get();
                 if (null == nameValuePairs) {
                     if (isIdentified) {
-                        assertEquals("DECIDE_ENDPOINT?version=1&lib=android&token=Test+Message+Queuing&distinct_id=new+person", endpointUrl);
+                        assertEquals("DECIDE_ENDPOINT?version=1&lib=android&token=Test+Message+Queuing&distinct_id=PEOPLE+ID", endpointUrl);
                     } else {
-                        assertEquals("DECIDE_ENDPOINT?version=1&lib=android&token=Test+Message+Queuing", endpointUrl);
+                        assertEquals("DECIDE_ENDPOINT?version=1&lib=android&token=Test+Message+Queuing&distinct_id=EVENTS+ID", endpointUrl);
```
yinfeiru: Is this condition covered by our test cases? I don't see it here.

joeatwork: Yes, it is run (It was failing before my change) - we call flush and observe the request.

joeatwork:
Thanks!

PR #229, opened 2015-06-08
flush at one second intervals in debug mode

jbwyme:


src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
```diff
@@ -207,8 +207,17 @@ public int getBulkUploadLimit() {
     }
 
     // Target max milliseconds between flushes. This is advisory.
-    public int getFlushInterval() {
-        return mFlushInterval;
+    public int getmFlushInterval() {
```
jbwyme: this should be `getFlushInterval`

src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
```diff
@@ -207,8 +207,17 @@ public int getBulkUploadLimit() {
...
+        return getFlushInterval(null);
+    }
+
+    public int getFlushInterval(Context context) {
+        boolean isDebuggable =  context != null && ( 0 != ( context.getApplicationInfo().flags &= ApplicationInfo.FLAG_INSTALLED ) );
+        if (isDebuggable) {
+            return 1000;
```
yinfeiru: Can we name this number by having a constant variable?

src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
```diff
@@ -207,8 +207,17 @@ public int getBulkUploadLimit() {
...
 
     // Target max milliseconds between flushes. This is advisory.
-    public int getFlushInterval() {
-        return mFlushInterval;
+    public int getmFlushInterval() {
+        return getFlushInterval(null);
+    }
+
+    public int getFlushInterval(Context context) {
+        boolean isDebuggable =  context != null && ( 0 != ( context.getApplicationInfo().flags &= ApplicationInfo.FLAG_INSTALLED ) );
```
yinfeiru: Two spaces between `=` and `context`

src/androidTest/java/com/mixpanel/android/mpmetrics/DecideFunctionalTest.java
```diff
@@ -64,7 +64,7 @@ public SharedPreferences get(final long timeout, final TimeUnit unit) throws Int
             }
         };
 
-        mMockConfig = new MPConfig(new Bundle()) {
+        mMockConfig = new MPConfig(new Bundle(), getContext()) {
```
yinfeiru: Can you put the return value of getContext() in a variable? This call appeared twice in this function (line 27 and 67).

src/androidTest/java/com/mixpanel/android/mpmetrics/HttpTest.java
```diff
@@ -88,7 +88,7 @@ public void setUp() {
             }
         };
 
-        final MPConfig config = new MPConfig(new Bundle()) {
+        final MPConfig config = new MPConfig(new Bundle(), getContext()) {
```
yinfeiru: same here

src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
jbwyme: i actually am adding a config for it

src/androidTest/java/com/mixpanel/android/mpmetrics/DecideFunctionalTest.java
jbwyme: it's just a proxy to the variable, not an issue IMO

src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
```diff
@@ -208,7 +212,16 @@ public int getBulkUploadLimit() {
 
     // Target max milliseconds between flushes. This is advisory.
     public int getFlushInterval() {
-        return mFlushInterval;
+        return getFlushInterval(null);
```
yinfeiru: This looks good but can we simply return mFlushInterval to avoid additional computation?

jbwyme: Hmm yeah that's true in this case but IMO it's best to have a single source of truth

src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
yinfeiru: This comment is hidden for some reason.. re-opening it

jbwyme: fixed

src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
yinfeiru: How often do we call this function? It's running on the mobile device. So, if it's quite often, I'd prefer to return mFlushInterval. Otherwise, I'm fine with it.

jbwyme: java will do this optimization for you so that you have literally no performance difference

yinfeiru: ok. cool.

Issue #95, opened 2014-05-21
java.security.NoSuchAlgorithmException: MessageDigest MD5 implementation not found


shimitapiro:
Hey guys, 
I'm seeing this exception in my bug report system, mostly on 4.1.X systems.
Any idea when will this be fixed ? 

0	java.lang.AssertionError: java.security.NoSuchAlgorithmException: MessageDigest MD5 implementation not found
1		at org.apache.harmony.xnet.provider.jsse.NativeCrypto.X509_NAME_hash(NativeCrypto.java:188)
2		at org.apache.harmony.xnet.provider.jsse.NativeCrypto.X509_NAME_hash_old(NativeCrypto.java:181)
3		at org.apache.harmony.xnet.provider.jsse.TrustedCertificateStore.hash(TrustedCertificateStore.java:417)
4		at org.apache.harmony.xnet.provider.jsse.TrustedCertificateStore.findCert(TrustedCertificateStore.java:377)
5		at org.apache.harmony.xnet.provider.jsse.TrustedCertificateStore.isTrustAnchor(TrustedCertificateStore.java:326)
6		at org.apache.harmony.xnet.provider.jsse.TrustManagerImpl.findTrustAnchorBySubjectAndPublicKey(TrustManagerImpl.java:307)
7		at org.apache.harmony.xnet.provider.jsse.TrustManagerImpl.cleanupCertChainAndFindTrustAnchors(TrustManagerImpl.java:237)
8		at org.apache.harmony.xnet.provider.jsse.TrustManagerImpl.checkTrusted(TrustManagerImpl.java:184)
9		at org.apache.harmony.xnet.provider.jsse.TrustManagerImpl.checkServerTrusted(TrustManagerImpl.java:163)
10		at org.apache.harmony.xnet.provider.jsse.OpenSSLSocketImpl.verifyCertificateChain(OpenSSLSocketImpl.java:597)
11		at org.apache.harmony.xnet.provider.jsse.NativeCrypto.SSL_do_handshake(Native Method)
12		at org.apache.harmony.xnet.provider.jsse.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:395)
13		at libcore.net.http.HttpConnection.setupSecureSocket(HttpConnection.java:210)
14		at libcore.net.http.HttpsURLConnectionImpl$HttpsEngine.makeSslConnection(HttpsURLConnectionImpl.java:478)
[...]

PR #231, opened 2015-06-17
Use a clean SSL context by default, allow users to override if needed

joeatwork:
Relevant to #95 - by default, this change isolates the Mixpanel library from app changes to the default SSL context, but allows users to explicitly set their own context if it is needed.

### DIFF
--- a/src/androidTest/java/com/mixpanel/android/mpmetrics/DecideCheckerTest.java
+++ b/src/androidTest/java/com/mixpanel/android/mpmetrics/DecideCheckerTest.java
@@ -16,6 +16,8 @@ import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.net.ssl.SSLSocketFactory;
+
 public class DecideCheckerTest extends AndroidTestCase {
 
     @Override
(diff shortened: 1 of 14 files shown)


yinfeiru:
hmmm.. So, looks like we don't support SSL connection at all in the current release?
Anyways, these changes look straight forward to me. Can you please run the full test suite and see what happens? I don't know if these additional SSL stuff will fail our current tests (most of our tests have a timeout and I'm expecting SSL will add in extra delay).

Also, have you tried the seledroid test? I don't know if this SSL connection work with it.

joeatwork:
We do - we use the system default SSL connection, which is usually fine. Unfortunately, we've run into issues where third party libraries or client code changes the default SSL configuration in ways that break with the api. I've run the tests - we shouldn't be emitting http requests there anyhow.

yinfeiru:
oh.. sorry.. I missed the SSL related code in ViewCrawler.java.

Does the test case cover the case where the user adds his own SSL? Are we able to test it?

joeatwork:
Right now we try not to hit the network in the unit tests - I've been testing this manually during development. We can at least test exposure to the default SSL socket settings, though.

yinfeiru:
OK. Cool. Sounds good to me.

PR #238, opened 2015-07-07
Ensure Tweak-only experiments are tracked by sharing the pipeline

joeatwork:


src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -725,24 +720,24 @@ private void handleEditorBindingsCleared(JSONObject clearMessage) {
                 Log.e(LOGTAG, "Bad clear request received", e);
             }
 
-            updateEditState();
+            applyVariantsAndEventBindings();
         }
 
         private void handleEditorTweaksReceived(JSONObject tweaksMessage) {
             try {
+                mEditorTweaks.clear();
```
yinfeiru: I'm a bit curious about this.. Why do we clear mEditorTweaks and then add all the tweaks to it, while in handleEditorChangeReceived we only do mEditorChanges.put?

joeatwork: I wanted to stay as close as I could to the old, known good editor behavior, which was to apply only the new inbound tweaks, and rely on the persistence of the mTweaks that are already applied.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -912,7 +938,38 @@ private void updateEditState() {
 
             mEditState.setEdits(editMap);
             mSeenExperiments.addAll(toTrack);
-            mVariantTracker.trackVariants(toTrack);
+
+            if (toTrack.size() > 0) {
```
yinfeiru: Why do you move this logic out from VariantTweak.trackVariants()? Do you think this is a better practice?

joeatwork: It was an opportunity to simplify a bit - VariantTracker was never used other than here. Originally, I wanted to eliminate the mutual dependency with MixpanelAPI, but it turned out not to be practical and VariantTracker was never removed.

yinfeiru:
should we ship this thing before Joe's leaving?

PR #239, opened 2015-07-16
Fix to retain A/B testing variant after app crashes

ryanseams:
Picking "experiment" instead of "experiment_id" here causes the app to encounter a key error and thus run the except block which invalidates all A/B testing variants in the shared preferences. So, for A/B tests after the app crashes when you resume using the app the test will not be applied by default and users will appear in a test they are not getting.


joeatwork:
Looks good! I'll merge this into 4.6.2

ryanseams:
This was merged into 4.6.2 release of the Android SDK

PR #242, opened 2015-07-28
added callback for tweak initialization

Syzygy2048:
I ran into the problem that I didn't know when tweaks have finished initializing - I believe this is already implemented for iOS. By implementing this I discovered that they only initialize after I call joinExperimentsIfAvailable() about 2 seconds after initializing mixpanel and my experiment, so for me this already was a huge help correctly implementing AB testing.
The implementation is similar to the one in 4.5.4 beta, even though that feature didn't make it into 4.6.+.

I also changed play-services to a specific version, which is kind of unrelated since it lead to issues on my end. Imho this should be done anyway since an update to play-services in the repo can break compilation and force changes onto the app just because someone decided to refactor that library. Feel free to ignore build.grade from the merge if you're against this decision. I should have put that in another commit, sorry.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,9 @@ android {
 // pom! If you need a transitive dependency in the library, you'll
 // have to change the bit in uploadArchives that marks all dependencies as optional.
 dependencies {
-    compile "com.google.android.gms:play-services:[3.1,)"
+
+    //compile "com.google.android.gms:play-services:[3.1,)" //bad kitty
+    compile 'com.google.android.gms:play-services-gcm:7.5.0'
 }
 
 android.libraryVariants.all { variant ->
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -18,11 +18,11 @@ import android.os.Bundle;
 import android.util.Log;
 
 import com.mixpanel.android.R;
-import com.mixpanel.android.viewcrawler.UpdatesFromMixpanel;
-import com.mixpanel.android.viewcrawler.TrackingDebug;
-import com.mixpanel.android.viewcrawler.ViewCrawler;
 import com.mixpanel.android.surveys.SurveyActivity;
 import com.mixpanel.android.util.ActivityImageUtils;
+import com.mixpanel.android.viewcrawler.TrackingDebug;
+import com.mixpanel.android.viewcrawler.UpdatesFromMixpanel;
+import com.mixpanel.android.viewcrawler.ViewCrawler;
 
 import org.json.JSONArray;
 import org.json.JSONException;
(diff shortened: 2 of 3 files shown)


yinfeiru:
Thanks for your PR and all your effort!

Actually, we do have a mechanism of updating your app once the response from the decide server is received. You can do something like
```
mMixpanel.getPeople().addOnMixpanelUpdatesReceivedListener(new OnMixpanelUpdatesReceivedListener() {
           @Override
           public void onMixpanelUpdatesReceived() {
               mMixpanel.getPeople().joinExperimentIfAvailable();
               Log.d("Experiment", "Got an experimmental thing");
           }
       });
}
```
(this code is copied from another thread about this feature)
Once you register the listener, it'll get called once the response is back.
https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/DecideChecker.java#L65

Syzygy2048:
I found that method, but was not able to get it to work the way I expected. It could be because I didn't have a correct joinExperimentsIfAvailable call though, so mixpanel started querying for experimemts with the next onResume - later than expected. (I think the documentation is lacking in that regard).
I still think the listeners are usefull as they helped me track down what was wrong with my implementation. They're also available on other platforms.

yinfeiru:
Thanks for your reply!

The lib queries for experiments every minute and also on pause and resume. Are you not seeing new variants coming in when you put the app into background?
Also, we've found a bug in the lib recently that might be related to the issue you saw. Please check out v4.6.3. Or update your gradle file to import it. Now you should see your listener, which is registered with addOnMixpanelUpdatesReceivedListener(), gets called on new variants receiving. You can find more info here https://github.com/mixpanel/mixpanel-android/pull/252/files.
One more thing, we've updated our document recently. Hopefully, it now answers the questions you had! If not, please list the questions you have and we are more than happy to update our doc!! You can find it at https://mixpanel.com/help/reference/android.

Syzygy2048:
My issue with the lib not querying was due to the fact that I initialized the lib in onCreate of my application class and needed an experiment in the very first activity. The lib doesn't seem to query for experiments at that point until the next onResume - not even if joinExperiments is called right after initialization. I managed to work around this by calling joinExperiments via a handler and a delay of 1000 ms, any lower and it would not work.

The updated documentation is very likely a direct result of my interaction with mixpanel tech support and my complaint that due to the state of the documentation I practically had to guess functionality - which is still the case except that I have a better understanding of the source code now.

Due to our iOS dev convincing our CTO, we're no longer using mixpanel for AB experiments, instead we've developed our own implementation, the results are sent to mixpanel via regular tracking. The main reason for that was, that we did not find a way to control user participation from the device itself, which is relevant if the user temporarily lacks an internet connection - if we let him default or add him to a random experiment, he'll still get assigned something from the service which leads to inconsistencies between what actually happened and what the experiment thinks that happened.

As this is developing into a tech support/complaint/rant ticket, I suggest that a decision should be made if this functionality makes less sense than it does on other platforms which have it and if it should be merged.

yinfeiru:
I'm really sorry to hear that! I do apologize for the lack of documentation! We'll definitely keep improving our documentation and make it easier for the users, like you, to integrate us!

What joinExperiments does is that it looks for experiments that haven't been applied. To make it work properly, you need to wait for the experiment response from the server to arrive. Otherwise, this API won't do anything as it can't find any new experiment. So, the proper thing to do is to register your own listener with addOnMixpanelUpdatesReceivedListener. And once a new experiment is received, you call 
joinExperiments to update your view.

For the offline problem, yes, our server has the full control. We were trying to make the Android lib really dumb when we design the A/B testing product. It only does what the server asks it to do. The reason is that we want to make our users write as less code as possible and make the server to handle most of the stuff for our users. And, clearly, this is a double-edge sword.

PR #244, opened 2015-07-28
Add app release prop

ryanseams:
This property is currently tracked in iOS and missed on Android.

This should be added for parity and allow for cross-platform comparisons between builds.


dbachelder:
:+1:

ryanseams:
closing in favor of https://github.com/mixpanel/mixpanel-android/pull/294, which resolves this for Android as well as iOS

PR #249, opened 2015-08-04
Fix addOnMixpanelUpdatesReceived listener for A/B tests

ryanseams:
Currently the ```addOnMixpanelUpdatesReceivedListener``` is broken for A/B tests. This listener is intended to run whenever an update is received from /decide and then allow the user to execute code.

Unfortunately, the current behavior does nothing for A/B tests and causes the code within the listener to never run. This fix updates the code to flip a boolean ```newContent``` when variants exist to force the listener code to run as well as adds the receipt of new experiments from /decide to the SDK debug logging.

I have tested this locally and was able to effectively use ```addOnMixpanelUpdatesReceivedListener``` to join an experiment as it comes in from /decide.

I do want to make it clear this is when A/B testing is the ONLY thing you are using. If you also receive surveys or notifications when you receive A/B tests then the code will run as expected.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -68,12 +68,17 @@ import java.util.Set;
             }
         }
 
+        if (variants.length() > 0) {
+            newContent = true;
+        }
+
         mVariants = variants;
 
         if (MPConfig.DEBUG) {
             Log.v(LOGTAG, "New Decide content has become available. " +
                     newSurveys.size() + " surveys and " +
-                    newNotifications.size() + " notifications have been added.");
+                    newNotifications.size() + " notifications and " +
+                    variants.length() + " experiments have been added.");
         }
 
         if (newContent && hasUpdatesAvailable() && null != mListener) {
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -69,10 +69,9 @@ import java.util.Set;
         }
 
         if (variants.length() > 0) {
+            mVariants = variants;
             newContent = true;
-        }
-
-        mVariants = variants;
+        } 
 
         if (MPConfig.DEBUG) {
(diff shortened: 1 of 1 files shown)


yinfeiru:
there's a problem with this patch - if there're other experiments under the same project, the decide server will return all of them, which makes variants non-empty.
I'll find a way to make sure the android lib only asks for the updates of the current app first. and then revisit your pr afterwards.

yinfeiru:
(sorry for closing your pr.. that was totally unintentional..)

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -68,12 +68,17 @@ public synchronized void reportResults(List<Survey> newSurveys, List<InAppNotifi
             }
         }
 
+        if (variants.length() > 0) {
+            newContent = true;
+        }
+
         mVariants = variants;
```
yinfeiru: can you please move this line into the if statement above? this is because we have a function checks if `mVariants` is null. `variants` will be an empty string (not null) if there's no variant, which breaks this check.
```
    public synchronized boolean hasUpdatesAvailable() {
        return (! mUnseenNotifications.isEmpty()) || (! mUnseenSurveys.isEmpty()) || mVariants != null;
    }
```

yinfeiru:
actually, I realized we have a rule of having up to one Android A/B test experiment per project. so, I think your branch is good to go as long as you address that comment. thanks for your effort!!

ryanseams:
@yinfeiru let me know if any other changes need to be made -- looks like this has everything now though based on your comments :)

yinfeiru:
yup, it looks good to me and you'll see the change in the next release.

yinfeiru:
well.. actually, there's an issue with this approach. if you look at the survey/notification logic above, it actually checks if the survey/notification have been shown. if so, newContent won't be updated. however, for the a/b variants, we don't check if it has been shown. so every time the activity resumes, the callback will be called no matter what. a better approach would be check if `variants.length() > 0` && variants and mVariants are identical. can you please make this change?

yinfeiru:
It's fixed here https://github.com/mixpanel/mixpanel-android/pull/252
The fix has been merged into v4.6.3.

PR #250, opened 2015-08-06
Guard against NPE when using certain Contexts.

kevinmost:
Using special Context objects, such as the BridgeContext provided by
Android Studio's previewer, would result in an NPE when calling
Mixpanel.getInstance() previously, because
ConfigurationChecker.checkBasicConfiguration() assumes that the context's
packageManager and packageName can't be null. This is now guarded
against, and Mixpanel.getInstance() will return null for contexts such as
this.


patedit:
@kevinmost Sorry for the late response. Can we do the same on `checkPushConfiguration`? Then I will merge the PR.

kevinmost:
Updated

patedit:
Thanks!

PR #252, opened 2015-08-11
New variants listener bug fix

yinfeiru:
Fix the bug where the user defined listener doesn't get called on new variants receive.

NOTE:
1. bump up the version
2. ~~make sure the decide change goes live before this one~~ - this is not needed as we are not going to merge in the updated_time change

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -75,7 +75,7 @@ import java.util.Set;
         for (int i = 0; i < newVariantsLength; i++) {
             try {
                 JSONObject variant = variants.getJSONObject(i);
-                if (mVariants == null || !mLoadedVariants.contains(variant.getInt("id"))) {
+                if (!mLoadedVariants.contains(variant.getInt("id"))) {
                     mVariants = variants;
                     newContent = true;
                     break;
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -5,10 +5,14 @@ import android.util.Log;
 import com.mixpanel.android.viewcrawler.UpdatesFromMixpanel;
 
 import org.json.JSONArray;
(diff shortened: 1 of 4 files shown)


src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -68,12 +70,39 @@ public synchronized void reportResults(List<Survey> newSurveys, List<InAppNotifi
...
+                    mLoadedVariants.add(variant.getInt("id"));
+                } catch(JSONException e) {
+                    Log.e(LOGTAG, "Could not convert variants[" + i + "] into a JSONObject while updating the map", e);
+                }
+            }
+        }
 
         if (MPConfig.DEBUG) {
             Log.v(LOGTAG, "New Decide content has become available. " +
                     newSurveys.size() + " surveys and " +
```
jbwyme: `surveys, ` instead of `surveys and `

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -68,12 +70,39 @@ public synchronized void reportResults(List<Survey> newSurveys, List<InAppNotifi
...
+                    mVariants = variants;
+                    newContent = true;
+                    break;
+                }
+            } catch(JSONException e) {
+                Log.e(LOGTAG, "Could not convert variants[" + i + "] into a JSONObject while comparing the new variants", e);
+            }
+        }
+
+        if (mVariants == variants) {
```
jbwyme: let's just add a `hasNewVariants` flag instead of this

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -68,12 +70,39 @@ public synchronized void reportResults(List<Survey> newSurveys, List<InAppNotifi
             }
         }
 
-        mVariants = variants;
+        int newVariantsLength = variants.length();
+
+        for (int i = 0; i < newVariantsLength; i++) {
```
jbwyme: let's add some comments explaining what we're doing with this flow

PR #271, opened 2015-10-20
Bump up the API level to 23 -

yinfeiru:
1. add a HttpClient library as it gets removed in Android M
2. replace NotificationCompat with the old Notification in API 9
3. add SuppressWarnings to eliminate the deprecation warnings

(tested on a phone with API 10)


yinfeiru:
Tested on the Android emulator with API 23.

yinfeiru:
Tested against API 21.

PR #277, opened 2015-10-27
Add support for api 23

yinfeiru:


yinfeiru:
Tested full functionality against API 23 and 21. Tested push against API 10.

src/main/java/com/mixpanel/android/util/HttpService.java
```diff
@@ -48,7 +48,7 @@ public boolean isOnline(Context context) {
     }
 
     @Override
-    public byte[] performRequest(String endpointUrl, List<NameValuePair> params, SSLSocketFactory socketFactory) throws ServiceUnavailableException, IOException {
+    public byte[] performRequest(String endpointUrl, Map<String, Object> params, SSLSocketFactory socketFactory) throws ServiceUnavailableException, IOException {
```
anlu: Why was this change necessitated by Android 23?

src/main/java/com/mixpanel/android/mpmetrics/SystemInformation.java
```diff
@@ -122,14 +122,13 @@ public String getCurrentNetworkOperator() {
...
 
-
     public Boolean isWifiConnected() {
         Boolean ret = null;
 
         if (PackageManager.PERMISSION_GRANTED == mContext.checkCallingOrSelfPermission(Manifest.permission.ACCESS_NETWORK_STATE)) {
             ConnectivityManager connManager = (ConnectivityManager) this.mContext.getSystemService(Context.CONNECTIVITY_SERVICE);
-            NetworkInfo wifiInfo = connManager.getNetworkInfo(ConnectivityManager.TYPE_WIFI);
-            ret = wifiInfo.isConnected();
+            NetworkInfo networkInfo = connManager.getActiveNetworkInfo();
```
anlu: Why did this have to change for 23? Could a user be both on 4G and WiFi, but networkInfo.getType() returns the 4G network? Seems like explicitly requesting info for TYPE_WIFI is more correct.

yinfeiru: `getNetworkInfo` is deprecated in API 23. `getActiveNetworkInfo` returns `Returns details about the currently active default data network`. If a user is both on 4G and WiFi, the default one should be Wifi.

src/main/java/com/mixpanel/android/util/HttpService.java
yinfeiru: NameValuePair has been removed in API 23. Basically, most of the stuff in `org.apache.http` has been removed in 23.

anlu:
LGTM % c

yinfeiru:
Cool. I'll do more test and ship it in a bit.

Issue #276, opened 2015-10-22
Add an option to disable ViewCrawler


yogurtearl:
On Mixpanel 4.5+, there is a ViewCrawler the periodically processes messages on the main thread. 
ViewCrawler is needed for these features to work:
* Web Configuration
* A/B Testing
* Dynamic Tweaks.

But if you are not using those features, there is no reason to incur the performance hit of having extra things run on your main thread.

Please add the ability to disable ViewCrawler ( for those that don't need Web Configuration, A/B Testing, and Dynamic Tweaks )

patedit:
Closing since it's already merged PR #313 

Thanks @yogurtearl !

aneemtalukder:
is there a way to have tweaks without the viewcrawler? or perhaps any way to run A/B tests without having to use the viewcrawler? my use case is i would like to simply get my variant for the test and update the UI myself without mixpanel's UI tools (as cool as they are!)

patedit:
Hi @aneemtalukder - sadly nope :(

aneemtalukder:
ahh got it, thanks!

PR #278, opened 2015-10-28
Allow SDK users to disable the "ViewCrawler" with a Manifest Metadata

yogurtearl:
Implements #276 

ViewCrawler will be on by default but can be turned off with: 

```
        <meta-data
            android:name="com.mixpanel.android.MPConfig.DisableViewCrawler"
            android:value="true"/>
```

### DIFF
--- a/src/androidTest/java/com/mixpanel/android/mpmetrics/MPConfigTest.java
+++ b/src/androidTest/java/com/mixpanel/android/mpmetrics/MPConfigTest.java
@@ -1,0 +1,68 @@
+package com.mixpanel.android.mpmetrics;
+
+import android.os.Build;
+import android.os.Bundle;
+import android.test.AndroidTestCase;
+import com.mixpanel.android.viewcrawler.ViewCrawler;
+
+public class MPConfigTest extends AndroidTestCase {
+
+    public static final String TOKEN                             = "TOKEN";
+    public static final String DISABLE_VIEW_CRAWLER_METADATA_KEY = "com.mixpanel.android.MPConfig.DisableViewCrawler";
+
+    public void testDisableViewCrawlerDefaultsToFalse() throws Exception {
+        final Bundle metaData = new Bundle();
+
+        // DON'T set "com.mixpanel.android.MPConfig.DisableViewCrawler" in the bundle
+
+        final MixpanelAPI mixpanelAPI = mixpanelApi(mpConfig(metaData));
+
+        if (Build.VERSION.SDK_INT >= MPConfig.UI_FEATURES_MIN_API) {
+            assertTrue("By default, we should use ViewCrawler as our Impl of UpdatesFromMixpanel",
+                       mixpanelAPI.constructUpdatesFromMixpanel(getContext(), TOKEN) instanceof ViewCrawler);
+        } else {
+            assertTrue("When API is older than MPConfig.UI_FEATURES_MIN_API, we should use NoOp",
+                       mixpanelAPI.constructUpdatesFromMixpanel(getContext(), TOKEN) instanceof MixpanelAPI.NoOpUpdatesFromMixpanel);
+        }
+    }
+
+    public void testDisableViewCrawlerTrueGetsNoOpImpl() throws Exception {
+        final Bundle metaData = new Bundle();
+
+        metaData.putBoolean(DISABLE_VIEW_CRAWLER_METADATA_KEY, true);
+
+        final MixpanelAPI mixpanelAPI = mixpanelApi(mpConfig(metaData));
+
+        if (Build.VERSION.SDK_INT >= MPConfig.UI_FEATURES_MIN_API) {
+            assertTrue("When DisableViewCrawler is true, we should use a NoOp Impl of UpdatesFromMixpanel",
+                       mixpanelAPI.constructUpdatesFromMixpanel(getContext(), TOKEN) instanceof MixpanelAPI.NoOpUpdatesFromMixpanel);
+        } else {
+            assertTrue("When API is older than MPConfig.UI_FEATURES_MIN_API, we should use NoOp",
+                       mixpanelAPI.constructUpdatesFromMixpanel(getContext(), TOKEN) instanceof MixpanelAPI.NoOpUpdatesFromMixpanel);
+        }
+    }
+
+    public void testDisableViewCrawlerFalseGetsViewCrawler() throws Exception {
+        final Bundle metaData = new Bundle();
+
+        metaData.putBoolean(DISABLE_VIEW_CRAWLER_METADATA_KEY, false);
+
+        final MixpanelAPI mixpanelAPI = mixpanelApi(mpConfig(metaData));
+
+        if (Build.VERSION.SDK_INT >= MPConfig.UI_FEATURES_MIN_API) {
+            assertTrue("When DisableViewCrawler is false, we should use ViewCrawler as our Impl of UpdatesFromMixpanel",
+                       mixpanelAPI.constructUpdatesFromMixpanel(getContext(), TOKEN) instanceof ViewCrawler);
+        } else {
+            assertTrue("When API is older than MPConfig.UI_FEATURES_MIN_API, we should use NoOp",
+                       mixpanelAPI.constructUpdatesFromMixpanel(getContext(), TOKEN) instanceof MixpanelAPI.NoOpUpdatesFromMixpanel);
+        }
+    }
+
+    private MPConfig mpConfig(final Bundle metaData) {
+        return new MPConfig(metaData, getContext());
+    }
+
+    private MixpanelAPI mixpanelApi(final MPConfig config) {
+        return new MixpanelAPI(getContext(), new TestUtils.EmptyPreferences(getContext()), TOKEN, config);
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
```diff
@@ -169,15 +169,16 @@ public synchronized void setSSLSocketFactory(SSLSocketFactory factory) {
         mBulkUploadLimit = metaData.getInt("com.mixpanel.android.MPConfig.BulkUploadLimit", 40); // 40 records default
         mFlushInterval = metaData.getInt("com.mixpanel.android.MPConfig.FlushInterval", 60 * 1000); // one minute default
         mDebugFlushInterval = metaData.getInt("com.mixpanel.android.MPConfig.DebugFlushInterval", 1 * 1000); // one second default
-        mDataExpiration = metaData.getInt("com.mixpanel.android.MPConfig.DataExpiration",  1000 * 60 * 60 * 24 * 5); // 5 days default
-        mMinimumDatabaseLimit = metaData.getInt("com.mixpanel.android.MPConfig.MinimumDatabaseLimit",  20 * 1024 * 1024); // 20 Mb
+        mDataExpiration = metaData.getInt("com.mixpanel.android.MPConfig.DataExpiration", 1000 * 60 * 60 * 24 * 5); // 5 days default
```
yinfeiru: lol  thanks for fixing this!!

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1976,8 +1982,8 @@ public synchronized void run() {
         private final Executor mExecutor = Executors.newSingleThreadExecutor();
     }
 
-    private class UnsupportedUpdatesFromMixpanel implements UpdatesFromMixpanel {
-        public UnsupportedUpdatesFromMixpanel(Tweaks tweaks) {
+    /* package */ class NoOpUpdatesFromMixpanel implements UpdatesFromMixpanel {
```
yinfeiru: I'm not exactly sure why you change the name of this function. First of all, I think the old name does indicate what's going on here. Second, we have another function called `UnsupportedUpdatesListener`. I'd like to keep the same naming approach so that people won't have to second guess what this function does.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1326,7 +1330,9 @@ public void onPrefsLoaded(SharedPreferences preferences) {
     /* package */ UpdatesFromMixpanel constructUpdatesFromMixpanel(final Context context, final String token) {
         if (Build.VERSION.SDK_INT < MPConfig.UI_FEATURES_MIN_API) {
             Log.i(LOGTAG, "Web Configuration, A/B Testing, and Dynamic Tweaks are not supported on this Android OS Version");
-            return new UnsupportedUpdatesFromMixpanel(sSharedTweaks);
+            return new NoOpUpdatesFromMixpanel(sSharedTweaks);
+        } else if (mConfig.getDisableViewCrawler()) {
```
yinfeiru: why not
```
if (Build.VERSION.SDK_INT < MPConfig.UI_FEATURES_MIN_API || mConfig.getDisableViewCrawler()) {
    return new NoOpUpdatesFromMixpanel(sSharedTweaks);
}
```

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -193,11 +193,19 @@
...
      */
     MixpanelAPI(Context context, Future<SharedPreferences> referrerPreferences, String token) {
+        this(context, referrerPreferences, token, MPConfig.getInstance(context));
+    }
+
+    /**
+     * You shouldn't instantiate MixpanelAPI objects directly.
+     * Use MixpanelAPI.getInstance to get an instance.
+     */
+    MixpanelAPI(Context context, Future<SharedPreferences> referrerPreferences, String token, final MPConfig config) {
```
yinfeiru: Hmmm.. I sort of like this change, but, on the other hand, I don't think it's necessary. As the comment says, the user shouldn't use this constructor to get an instance. The only place uses this is the `getInstance` API. And because of that, I think it's an overkill to provide an extra level of extraction in order to make it more flexible. Do you have a good reason for adding this that I'm not aware of?

src/androidTest/java/com/mixpanel/android/mpmetrics/MPConfigTest.java
```diff
@@ -0,0 +1,68 @@
+package com.mixpanel.android.mpmetrics;
```
yinfeiru: I like this test a lot!!

yinfeiru:
Thank you for this PR! Providing an option to disable the `ViewCrawler` is something we always want but didn't get a chance to get to. Thanks for all your effort!

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yogurtearl: It is needed for testing ( see MPConfigTest ). ( it is package private )

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yogurtearl: 2 separate if clauses make it clear that we are handling 2 separate cases here. 

1. API of the device it too old
1. turned off by configuration

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yogurtearl: "UnsupportedUpdatesFromMixpanel" doesn't describe what this class does. It describes one use case of this class (i.e. when the API version doesn't support viewcrawler ). 

"NoOpUpdatesFromMixpanel" describes what is does ( i.e. nothing, "NoOp" )

It may, in fact, be supported by the API of the device, but just disabled by config, so I feel "Unsupported" is confusing in that case.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: I think having them in the same clause is clear enough - the `||` explains everything. Or, you can have:
```
if (Build.VERSION.SDK_INT < MPConfig.UI_FEATURES_MIN_API ||
     mConfig.getDisableViewCrawler()) {...}
```
to make it clearer.
Having two blocks return the same thing looks weird to me.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: fair enough  :)

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: oh.. good point. I didn't notice it. Then keep it!

yinfeiru:
@yogurtearl Cool! Thank you! I'll make sure to test and merge it!

yogurtearl:
@yinfeiru when do you think this will be merged and released?

yogurtearl:
@ryanseams @patedit anything else need here, before this gets merged?

yinfeiru:
@yogurtearl sorry for the late. I'll probably test it today and make sure this will be included in the next release.

src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
```diff
@@ -169,15 +169,16 @@ public synchronized void setSSLSocketFactory(SSLSocketFactory factory) {
...
+        mMinimumDatabaseLimit = metaData.getInt("com.mixpanel.android.MPConfig.MinimumDatabaseLimit", 20 * 1024 * 1024); // 20 Mb
         mDisableFallback = metaData.getBoolean("com.mixpanel.android.MPConfig.DisableFallback", true);
         mResourcePackageName = metaData.getString("com.mixpanel.android.MPConfig.ResourcePackageName"); // default is null
         mDisableGestureBindingUI = metaData.getBoolean("com.mixpanel.android.MPConfig.DisableGestureBindingUI", false);
         mDisableEmulatorBindingUI = metaData.getBoolean("com.mixpanel.android.MPConfig.DisableEmulatorBindingUI", false);
         mDisableAppOpenEvent = metaData.getBoolean("com.mixpanel.android.MPConfig.DisableAppOpenEvent", true);
+        mDisableViewCrawler = metaData.getBoolean("com.mixpanel.android.MPConfig.DisableViewCrawler", false);
```
yinfeiru: I'd prefer to set the default value to `true` so that it won't break the code for users who were using this feature.

yinfeiru: Forget the previous comment, setting it to true will disable the view crawler..

yinfeiru:
Closing this as I've created a new one here #313

PR #283, opened 2015-11-03
use device defaults for notification features

grendell:
This includes sound, lights, and vibration settings.

Without this, `Notification.defaults` defaults to 0 while `Notification.DEFAULT_ALL` == -1, which will result in no vibration or sound for any notification, even when the user has set these to happen.

More info [here](http://developer.android.com/reference/android/app/Notification.html#defaults).

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
@@ -263,7 +263,8 @@ public class GCMReceiver extends BroadcastReceiver {
                 setWhen(System.currentTimeMillis()).
                 setContentTitle(notificationData.title).
                 setContentText(notificationData.message).
-                setContentIntent(intent);
+                setContentIntent(intent).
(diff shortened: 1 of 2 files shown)


grendell:
Seriously though, please let me know if there's anything I can do to help get this merged.  Thanks.

yinfeiru:
@grendell Thanks for the PR! Although what you suggested makes perfect sense, I'd like to add a flag in MPConfig (eg. https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java#L176) so that people will be able to turn this feature on/off. A lot of existing users have `AutoShowMixpanelUpdates` on and notifications will be shown when the app is resumed. In that case, since the app is in foreground already, there's no need to alert the user.

grendell:
@yinfeiru thanks for the feedback.  I have updated the pull request accordingly.

NickMele:
Is there a plan/timeline to get this PR merged?

patedit:
Hi guys, I've modified the PR since it had some conflicts and merge it to master. Thanks for your contribution!

PR #287, opened 2015-11-30
Add platfor for app_open event

iauglov:
Add for display platform on app_open event


yinfeiru:
@iauglov Thanks for writing in!! I believe we have a similar super property already `Operating System:Android`. I'll close this PR for now. Please re-open if you do need this property for some reason.

Issue #286, opened 2015-11-29
FadingImageView - ClassCastException


jiri-hlt:
Fatal Exception: java.lang.ClassCastException: android.widget.RelativeLayout cannot be cast to android.widget.LinearLayout
       at com.mixpanel.android.surveys.FadingImageView.onMeasure(FadingImageView.java:46)
       at android.view.View.measure(View.java:17624)
       at android.widget.RelativeLayout.measureChildHorizontal(RelativeLayout.java:728)
       at android.widget.RelativeLayout.onMeasure(RelativeLayout.java:464)
       at android.view.View.measure(View.java:17624)

jiri-hlt:
com.mixpanel.android.surveys.FadingImageView.onMeasure(FadingImageView.java:46)
should just be casted to ViewGroup:
```
ViewGroup container = (ViewGroup) getParent();
```

pavel-kurnosov:
When this will be released? I got a lot of issues on Crashlytics with same issue...

yinfeiru:
Closing this as it's been fixed in #288

yinfeiru:
@pavel-kurnosov a new release will probably be available in a week.

PR #288, opened 2015-12-02
Fix crash from ClassCastException when displaying in-app images

ryanseams:
A customer has highlighted this issue and its fix as an issue [here](https://github.com/mixpanel/mixpanel-android/issues/286).

Currently what happens is the linear layout cannot be applied when rendering an image, causing a crash to show on the device. Instead, the proposal is to grab the ViewGroup itself instead of LinearLayout -- the customer has confirmed this resolves the issues in their testing.


patedit:
Perfect. Thanks @ryanseams !

PR #290, opened 2015-12-14
make survey prompt optional via metadata flag

grendell:
Currently in the Mixpanel SDK for iOS, there is no prompt before the user is shown a survey.  We would like to replicate this on Android.  The intention of this change is to allow this to be configurable via a metadata tag in the app manifest, falling back to current behavior if this tag is not present.

Example usage:

```
    <meta-data
        android:name="com.mixpanel.android.MPConfig.ShouldPromptForSurvey"
        android:value="false"/>
```

If there is a more appropriate place for this flag, please let me know.


grendell:
Also, if there are any instructions for how to compile `mixpanel-android` locally, I have run into major issues both attempting to export an AAR and creating a submodule in our app, using Android Studio 1.5 and gradle 1.5.0.  The AAR appears to be incomplete, while compiling the submodule leads to OOMs which I can't seem to solve with `studio.vmoptions`.

grendell:
Updated based on feedback from https://github.com/mixpanel/mixpanel-android/pull/283.

PR #292, opened 2015-12-15
fix for NPE in case if File.listFiles returns null

andrey-starodubtsev:
fix for NPE found via ACRA, according http://developer.android.com/reference/java/io/File.html#listFiles(), `File.listFiles` can return `null`

### DIFF
--- a/src/main/java/com/mixpanel/android/util/ImageStore.java
+++ b/src/main/java/com/mixpanel/android/util/ImageStore.java
@@ -105,12 +105,12 @@ public class ImageStore {
 
     public void clearStorage() {
         File[] files = mDirectory.listFiles();
-        int length = files.length;
-        for (int i = 0; i < length; i++) {
-            final File file = files[i];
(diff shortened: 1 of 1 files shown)


patedit:
Hi @andrey-starodubtsev Even if this PR would solve a crash in that line, it seems the problem may be somewhere else. If `listFiles` returns `null` is only because `mDirectory` is not a directory and that should never happen. Do you have some logs I could take a look? What is the Mixpanel SDK version you are using? Thanks!

andrey-starodubtsev:
Hello, @patedit. I got crashes via ACRA while using this method in our code and added check for null everywhere, including 3-rd party libraries we use in our product, just in case.

patedit:
Any of your crashes were coming from our library?

patedit:
@andrey-starodubtsev Will close this PR but feel free to open an issue if you see a crash coming from our library. Thanks!

PR #293, opened 2015-12-15
Assign $experiments super property before $experiment_started event sends

ryanseams:
Currently, the $experiments super property is not assigned until after the event sends marking the start of the experiment ($experiment_started). Instead, the super property should be assigned before the event, allowing for parity with iOS and for users to analyze this information correctly within the UI.

Tested on Jelly Bean (API level 16) and this change was successful in fixing the behavior.


patedit:
Makes sense! We currently dispatch setting the super property on a separate queue in iOS so we can't guarantee this behavior on iOS as well. I'll note it there. Thanks!

Issue #284, opened 2015-11-10
Crash when trying to display in-app as activity is stopped


ryanseams:
In this scenario, the user is trying to display an in-app notification (specifically a mini version) and is experiencing a crash. This exception is occurring because the Mixpanel SDK attempts to render an in-app notification as the app goes to the background.

Due to the way Android manages activities, the activity is actually stopped before the in-app attempts to render, resulting in an [illegal state exception](http://www.androiddesignpatterns.com/2013/08/fragment-transaction-commit-state-loss.html).

To recreate this bug, you will need to target a user with an in-app and background the app as soon as the in-app notification attempts to render. While rare, there should be a check to prevent against the in-app rendering in this scenario.

Crash Log:
11-10 04:40:30.217 28127-28127/? V/MixpanelAPI.API: Attempting to show mini notification.
11-10 04:40:30.217 28127-28127/? D/AndroidRuntime: Shutting down VM
11-10 04:40:30.217 28127-28127/? W/dalvikvm: threadid=1: thread exiting with uncaught exception (group=0x41c19d58)
11-10 04:40:30.227 28127-28127/? E/AndroidRuntime: FATAL EXCEPTION: main
11-10 04:40:30.227 28127-28127/? E/AndroidRuntime: Process: com.ryanseams.mixpanel_playground, PID: 28127
11-10 04:40:30.227 28127-28127/? E/AndroidRuntime: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
11-10 04:40:30.227 28127-28127/? E/AndroidRuntime:     at android.app.FragmentManagerImpl.checkStateLoss(FragmentManager.java:1323)
11-10 04:40:30.227 28127-28127/? E/AndroidRuntime:     at android.app.FragmentManagerImpl.enqueueAction(FragmentManager.java:1341)
11-10 04:40:30.227 28127-28127/? E/AndroidRuntime:     at android.app.BackStackRecord.commitInternal(BackStackRecord.java:597)
11-10 04:40:30.227 28127-28127/? E/AndroidRuntime:     at android.app.BackStackRecord.commit(BackStackRecord.java:575)
11-10 04:40:30.227 28127-28127/? E/AndroidRuntime:     at com.mixpanel.android.mpmetrics.MixpanelAPI$PeopleImpl$4.run(MixpanelAPI.java:1894)
[...]

PR #300, opened 2016-01-21
transaction.commit() might raise an IllegalStateException exception 

yinfeiru:
if the current activity is no long available. When that happens, we want to add the notification back to the list

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -175,6 +175,12 @@ import java.util.Set;
         return notif;
     }
 
+    // if a notification was failed to show, add it back to the unseen list so that we
+    // won't lose it
+    public synchronized void handleNotificationDisplayFailure(InAppNotification notif) {
+        mUnseenNotifications.add(notif);
+    }
+
     public synchronized boolean hasUpdatesAvailable() {
         return (! mUnseenNotifications.isEmpty()) || (! mUnseenSurveys.isEmpty()) || mVariants != null;
     }
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -120,7 +120,7 @@ import java.util.Set;
         }
         Survey s = mUnseenSurveys.remove(0);
         if (replace) {
-            mUnseenSurveys.add(mUnseenSurveys.size(), s);
+            mUnseenSurveys.add(s);
         }
         return s;
     }
@@ -152,7 +152,7 @@ import java.util.Set;
         }
         InAppNotification n = mUnseenNotifications.remove(0);
         if (replace) {
-            mUnseenNotifications.add(mUnseenNotifications.size(), n);
+            mUnseenNotifications.add(n);
         } else {
             if (MPConfig.DEBUG) {
                 Log.v(LOGTAG, "Recording notification " + n + " as seen.");
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -177,8 +177,8 @@ import java.util.Set;
 
     // if a notification was failed to show, add it back to the unseen list so that we
     // won't lose it
-    public synchronized void markNotificationAsUnseen(InAppNotification notif, boolean isDebugMode) {
-        if (!isDebugMode) {
+    public synchronized void markNotificationAsUnseen(InAppNotification notif) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1891,7 +1891,15 @@ public void run() {
...
                                 transaction.setCustomAnimations(0, R.anim.com_mixpanel_android_slide_down);
                                 transaction.add(android.R.id.content, inapp);
-                                transaction.commit();
+
+                                try {
+                                    transaction.commit();
+                                } catch (IllegalStateException e) {
+                                    // if the app is in the background or the current activity gets killed, rendering the
+                                    // notifiction will lead to a crash
+                                    Log.v(LOGTAG, "Unable to show notification.");
```
patedit: Let's wrap this around `if (MPConfig.DEBUG) {}`

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -175,6 +175,12 @@ public synchronized InAppNotification getNotification(int id, boolean replace) {
         return notif;
     }
 
+    // if a notification was failed to show, add it back to the unseen list so that we
+    // won't lose it
+    public synchronized void handleNotificationDisplayFailure(InAppNotification notif) {
```
patedit: For now we could just change the name of this method to something like `markNotificationAsUnseen`

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -175,6 +175,12 @@ public synchronized InAppNotification getNotification(int id, boolean replace) {
...
+        mUnseenNotifications.add(notif);
```
patedit: Should we check first whether the notification was already in the `LinkedList` ?

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
yinfeiru: c77ff20

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: c77ff20

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
yinfeiru: I don't think we need to check it because there're only 2 functions capable of adding new notifications, and the other one is `reportResults` which gets called when we receive a `decide response`. Also, I don't think we ever want to compose notifications on the device. So I think it's pretty safe to add notifications back without checking.

yinfeiru:
#284

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -175,6 +175,14 @@ public synchronized InAppNotification getNotification(int id, boolean replace) {
         return notif;
     }
 
+    // if a notification was failed to show, add it back to the unseen list so that we
+    // won't lose it
+    public synchronized void markNotificationAsUnseen(InAppNotification notif, boolean isDebugMode) {
```
patedit: If isDebugMode is coming from MPConfig.DEBUG we do not need to pass that parameter always. We can just check it here.

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -175,6 +175,14 @@ public synchronized InAppNotification getNotification(int id, boolean replace) {
...
+        if (!isDebugMode) {
```
patedit: Why do we want to do this?

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
yinfeiru: Yea, we could do that. But I want to follow the same pattern as the other places.

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
yinfeiru: If it's debug mode, we add the notification back to the unseen list when `getNotification` gets called. So in this mode, we don't want to add the notification back in this function (as it's already in the unseen list).

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
yinfeiru: b409263

patedit:
This is good :)

PR #301, opened 2016-01-22
Keep prior decide response on identify calls when distinct ID does not change

ryanseams:
Currently each time you call ```identify()``` or ```getPeople().identify()```, the library clears the decide response of all surveys and in-apps as indicated [here](https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java#L41-L45). While this makes sense when the ID for a user changes, this causes problems when the ID stays the same.

Instead of resetting each time, we should do a check on the distinct ID to see if it is the same or not -- if it is the same, we should hold the response. If the distinct ID is different, we should clear the response and make a new request to decide. One additional improvement on this clearing of the response is that we should clear variants in addition to surveys and notifications.

As it stands, the current SDK behavior causes issues for many users of the library. Since each one calls identify with different behavior each user has a different level of issues with this; however, they seem to be traced to this same root problem -- the SDK says a notification exists, the user calls identify, the decide response is cleared, and then they attempt to render a notification or survey which no longer exists.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -39,7 +39,7 @@ import java.util.Set;
     // Called from other synchronized code. Do not call into other synchronized code or you'll
     // risk deadlock
     public synchronized void setDistinctId(String distinctId) {
-        if (mDistinctId == null || (mDistinctId != null && !mDistictId.equals(distinctId))){
+        if (mDistinctId == null || !mDistictId.equals(distinctId)){
             mUnseenSurveys.clear();
             mUnseenNotifications.clear();
             mVariants = null;
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -39,7 +39,7 @@ import java.util.Set;
     // Called from other synchronized code. Do not call into other synchronized code or you'll
     // risk deadlock
     public synchronized void setDistinctId(String distinctId) {
-        if (mDistinctId.equals(distinctId)) {
+        if (mDistinctId != null && mDistinctId.equals(distinctId)) {
             return; // the user is the same, so no new response is needed
         } else {
             mUnseenSurveys.clear();
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -42,7 +42,6 @@ import java.util.Set;
         if (mDistinctId == null || !mDistictId.equals(distinctId)){
             mUnseenSurveys.clear();
             mUnseenNotifications.clear();
-            mVariants = null;
         }
         mDistinctId = distinctId;
     }
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -39,7 +39,7 @@ import java.util.Set;
     // Called from other synchronized code. Do not call into other synchronized code or you'll
     // risk deadlock
     public synchronized void setDistinctId(String distinctId) {
-        if (mDistinctId == distinctId) {
+        if (mDistinctId.equals(distinctId)) {
             return; // the user is the same, so no new response is needed
         } else {
             mUnseenSurveys.clear();
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -39,14 +39,12 @@ import java.util.Set;
     // Called from other synchronized code. Do not call into other synchronized code or you'll
     // risk deadlock
     public synchronized void setDistinctId(String distinctId) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -39,9 +39,14 @@ public String getToken() {
     // Called from other synchronized code. Do not call into other synchronized code or you'll
     // risk deadlock
     public synchronized void setDistinctId(String distinctId) {
-        mUnseenSurveys.clear();
-        mUnseenNotifications.clear();
-        mDistinctId = distinctId;
+        if (mDistinctId == distinctId) {
```
spiritinlife: if (mDistinctId != null && mDistinctId.equals(distinctId)) 
I think this should be preferred.

ryanseams: agree with the equals operator... as for the nulls, it looks like this is taken care of before the other methods call this one so not sure it is needed here.

patedit: This could be also called from the client by a Mixpanel instance: `mMixpanelInstance.getPeople().setDistinctId(null)`. However, I think we should also let the user nullify the `distinctId` since it's the initial state for this variable and we do sanity checks when we use it (e.g. https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/DecideChecker.java#L200).

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -39,9 +39,14 @@ public String getToken() {
     // Called from other synchronized code. Do not call into other synchronized code or you'll
     // risk deadlock
     public synchronized void setDistinctId(String distinctId) {
-        mUnseenSurveys.clear();
-        mUnseenNotifications.clear();
-        mDistinctId = distinctId;
+        if (mDistinctId.equals(distinctId)) {
```
patedit: `mDistinctId` initial value is null. This will always crash :(

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -39,9 +39,14 @@ public String getToken() {
...
+            return; // the user is the same, so no new response is needed
```
patedit: Not a big fan of return statements at the beginning like this. I'd either do:
```
if (something)
   return

code
```

or

```
if (!something) {
    code
}
```

ryanseams:
@patedit does checking for null resolve the crash? this was suggested and tested to success by @spiritinlife

spiritinlife:
The null check is needed and i have tested that without it the app crashes, because the first time  distinctId is always going to be null.
 
I also agree with @patedit about the use of return but it is your codebase do as you wish :)
i would suggest something like , which takes  mDistinctId = distinctId; assignment out of the condition.
```
if (mDistinctId == null || (mDistinctId != null && !mDistictId.equals(distinctId))){
  mUnseenSurveys.clear();
  mUnseenNotifications.clear();
  mVariants = new JSONArray();
} 
  mDistinctId = distinctId;
```

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -39,9 +39,14 @@ public String getToken() {
...
+        } else {
+            mUnseenSurveys.clear();
+            mUnseenNotifications.clear();
+            mVariants = new JSONArray();
```
patedit: We have to nullify `mVariants` here since it'll conflict with other logic (see https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java#L179).
Alternatively, we could initialize mVariants in https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java#L192 and change line 179 to check whether it's empty or not. In here, it'd be better to empty the array instead of creating a new one.

patedit:
@ryanseams I am afraid it's not enough, sorry! Notice if you just check if `mDistinctId` is null it will never go inside the `if` statement since `mDistinctId` is initialized to null.
What @spiritinlife just proposed it's what we want.

ryanseams:
@patedit not a problem... i went ahead and changed the fix based on the comments from @spiritinlife and yourself. let me know if i need to change anything else for this to be merged!

patedit:
@ryanseams Sorry for the late response and for not paying enough attention before. We do not need to check `mDistinctId != null` since when we reach that part of the `if` statement we can already assume `mDistinctId` is not null. I'll merge after your change.

ryanseams:
@patedit i removed the check on ```mDistinctId != null``` as per your request.

one thing i also did was remove the code to set ```mVariants = null``` -- if we clear the variants and the user is in the middle of an a/b test we leave a high chance for crashes or glitchy end user experiences. we should think about how we want to handle this case for the future separate from this bug fix.

patedit:
@ryanseams thanks!

spiritinlife:
Guys there is a typo in the code **mDistictId**

patedit:
@spiritinlife ευχαριστώ :)

PR #302, opened 2016-01-25
Hide navigation bar when mini app notification appears

spiritinlife:
When navigation bar is translucent mini app notification shows bellow the navigation bar.
With this fix before the notification shows we hide the navigation bar until the notification disappears


zihejia:
Housekeeping, close PR that is no longer relevant. Thanks @spiritinlife for providing it.

PR #303, opened 2016-01-25
Backoff algorithm under failed connection

patedit:
We will increase the flush interval when a request fails.
retries = 0 --> back off time = flush interval
retries = 1 --> back off time = 2 * flush interval
retries = 2 --> back off time = 4 * flush interval
....

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
@@ -454,7 +454,7 @@ import javax.net.ssl.SSLSocketFactory;
                     } else {
                         logAboutMessageToMixpanel("Retrying this batch of events.");
                         if (!hasMessages(FLUSH_QUEUE)) {
-                            mBackOffTime = Math.min((long)Math.pow(2, mFailedRetries) * mFlushInterval, 30*60*1000); // max 30 min
+                            mBackOffTime = Math.min((long)Math.pow(2, mFailedRetries) * mFlushInterval, 30 * 60 * 1000); // max 30 min
                             sendEmptyMessageDelayed(FLUSH_QUEUE, mBackOffTime);
                             mFailedRetries++;
                         }
--- a/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
@@ -427,6 +427,10 @@ import javax.net.ssl.SSLSocketFactory;
                                 } catch (UnsupportedEncodingException e) {
                                     throw new RuntimeException("UTF not supported on this platform?", e);
                                 }
+                                if (mFailedRetries > 0) {
+                                    mFailedRetries = 0;
+                                    removeMessages(FLUSH_QUEUE);
+                                }
 
                                 logAboutMessageToMixpanel("Successfully posted to " + url + ": \n" + rawMessage);
                                 logAboutMessageToMixpanel("Response was " + parsedResponse);
@@ -450,7 +454,9 @@ import javax.net.ssl.SSLSocketFactory;
                     } else {
                         logAboutMessageToMixpanel("Retrying this batch of events.");
                         if (!hasMessages(FLUSH_QUEUE)) {
-                            sendEmptyMessageDelayed(FLUSH_QUEUE, mFlushInterval);
+                            mBackOffTime = (long)Math.pow(2, mFailedRetries) * mFlushInterval;
+                            sendEmptyMessageDelayed(FLUSH_QUEUE, mBackOffTime);
+                            mFailedRetries++;
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -450,7 +454,9 @@ private void sendData(MPDbAdapter dbAdapter, MPDbAdapter.Table table, String[] u
                     } else {
                         logAboutMessageToMixpanel("Retrying this batch of events.");
                         if (!hasMessages(FLUSH_QUEUE)) {
-                            sendEmptyMessageDelayed(FLUSH_QUEUE, mFlushInterval);
+                            mBackOffTime = Math.min((long)Math.pow(2, mFailedRetries) * mFlushInterval, 30*60*1000); // max 30 min
```
yinfeiru: `30 * 60 * 1000`

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -450,7 +454,9 @@ private void sendData(MPDbAdapter dbAdapter, MPDbAdapter.Table table, String[] u
...
+                            sendEmptyMessageDelayed(FLUSH_QUEUE, mBackOffTime);
+                            mFailedRetries++;
```
yinfeiru: Also, I think we want to limit the max value of `mFailedRetries` but not `mBackOffTime`. I don't think doing `mFailedRetries++` makes any sense if `(long)Math.pow(2, mFailedRetries) * mFlushInterval` is larger than `30 * 60 * 1000`.

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -427,6 +427,10 @@ private void sendData(MPDbAdapter dbAdapter, MPDbAdapter.Table table, String[] u
                                 } catch (UnsupportedEncodingException e) {
                                     throw new RuntimeException("UTF not supported on this platform?", e);
                                 }
+                                if (mFailedRetries > 0) {
+                                    mFailedRetries = 0;
+                                    removeMessages(FLUSH_QUEUE);
```
Quuxplusone: Does this remove all the FLUSH_QUEUE jobs from the work queue? What would go wrong if this line weren't here?
(This is probably related to my comment in the best-practices doc about how a single 503 can cause an otherwise quiet queue never to get flushed again. Looks like you're introducing a second instance of that same issue; not that there's necessarily anything wrong with that, for now.)

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -298,7 +298,7 @@ else if (msg.what == KILL_WORKER) {
                     ///////////////////////////
 
                     if ((returnCode >= mConfig.getBulkUploadLimit() || returnCode == MPDbAdapter.DB_OUT_OF_MEMORY_ERROR) &&
-                            SystemClock.elapsedRealtime() >= mRetryAfter) {
+                            SystemClock.elapsedRealtime() >= mRetryAfter && mFailedRetries <= 0) {
```
Quuxplusone: Instead of implementing backoff by inserting a delayed FLUSH_QUEUE job, why not just implement backoff by the conceptual equivalent of

    mRetryAfter *= 2;

? (Implementation-wise, you'd keep track of the `mLastFlushTime` and ask whether `SystemClock.elapsedRealtime() >= mLastFlushTime + mRetryAfter`.)

patedit: The problem with that is that we would not schedule any job. That means that we'd have to rely on the user to track a new event, for example, and when that happen is when we check for `SystemClock.elapsedRealtime()` and compare with a minimum threshold (`mRetryAfter`). Implementation wise I believe it makes more sense to schedule a task in the future than rely on the activity on the user for it to be executed.

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
patedit: Yes, it does. Let's assume we remove this line. Now, imagine that a user calls `flush()` and there is already a FLUSH_QUEUE job pending because of a previous failed network request. If this time the network request goes through, we won't flush the queue again until that previously scheduled job is executed (could be in 30 minutes!). That means we will queue events for 30 minutes, and that's too much. We do not add new FLUSH_QUEUE events if there is already one pending. The only option is if the user manually calls `flush()`.

Quuxplusone: Ah, okay. Sounds good.
I'm looking forward to the followup PR that rips out all of this delayed-job stuff and replaces it with a background thread a la the iOS library. ;)

jbwyme:
lgtm!

PR #304, opened 2016-01-25
Add spanish translation for surveys

cokeperez:


patedit:
Thanks a lot @cokeperez !

There's something that could be misinterpreted. When prompting the question "Te importaria contestar a alguna pregunta?" one of the choices is "Claro". In English "Sure" could be valid but in Spanish "Claro" means exactly the opposite for this example. I would change the question and/or "Claro". What do you think? As a spanish speaker I know this is tricky :)

papajulio:
@patedit I would change the question it's quite tricky. @cokeperez what do you think about 
'¿Quieres contestar alguna pregunta?' or something similar so the answer fits better?

cokeperez:
@patedit @papajulio thank you for your feedback.

The original question in english is '_Mind taking a quick survey?_' so maybe we can put '_¿Podrías contestar una breve encuesta?_'

What do you think?

patedit:
I like it! As soon as you change it, I'll merge your PR. Thanks @cokeperez @papajulio

cokeperez:
Done :)

Thx @patedit!

patedit:
Thanks guys!

PR #306, opened 2016-01-25
Acceptance test support

yinfeiru:
Update the acceptance test so it will work with the latest lib
1. Get rid of the layout test which is not officially supported
2. Bump the compile sdk version to 23
3. Add a listener to apply the A/B test changes


patedit:
Looks good!

PR #307, opened 2016-01-26
Add event tracking when survey is received (parity with iOS)

ryanseams:
Currently iOS tracks that a survey is received [here](https://github.com/mixpanel/mixpanel-iphone/blob/2106d67e4197b3244f202793d54d6e580152773e/Mixpanel/Mixpanel.m#L1521-L1525) whereas Android does not track any event to indicate a survey has been received.

This PR adds support for the same event and properties as sent from iOS to be sent when a survey is received on Android for parity.

### DIFF
--- a/src/main/java/com/mixpanel/android/surveys/SurveyActivity.java
+++ b/src/main/java/com/mixpanel/android/surveys/SurveyActivity.java
@@ -183,7 +183,7 @@ public class SurveyActivity extends Activity {
         });
         ctaButton.setOnTouchListener(new View.OnTouchListener() {
             @Override
-            public boolean onTouch(View v, MotionEvent event) {
+			public boolean onTouch(View v, MotionEvent event) {
                 if (event.getAction() == MotionEvent.ACTION_DOWN) {
                     v.setBackgroundResource(R.drawable.com_mixpanel_android_cta_button_highlight);
                 } else {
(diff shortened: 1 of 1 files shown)


yinfeiru:
lgtm. @patedit any comments?

@ryanseams have you tested it?

ryanseams:
@yinfeiru tested on OS version 4.4.2 -- in my testing i found an edge case to be that i was marking the survey as shown or not based on the # of answers. in theory this is wrong because i could see the survey, and then not answer any questions.

i edited the code to use the boolean value for ```mSurveyBegun``` as the value of that property. everything should be good to go now.

patedit:
Thanks @ryanseams ! It looks good. `mSurveyBegun` will work. However, just wanted to say that `onDestroySurvey` will only be called when the survey was displayed. It'd also be good to pass `true`. In iOS, a survey has a name and it could be `$ignore`. When the name is `$ignore` we do not track anything about the survey. When could it be `$ignore`? Android SDK does not even handle survey names. @yinfeiru

ryanseams:
@patedit in my testing, the ```onDestroySurvey``` method is called whether or not the survey is displayed -- as long as the prompt which says "Mind taking a quick survey?" appears, this method is called. as such, i pass false if the user says no to the prompt, and true in all other cases as this means they have seen the survey.

as for the ```$ignore``` we can implement this behavior, however, it would require modification to the ```Survey.java``` file. i have a fix which includes this if you want to ignore surveys with that title, however, as this is not explicitly mentioned as a library feature i am not sure it should be included. the android library has a config option called ```TestMode``` which iOS does not have which should suffice for allowing a user to test a survey (which it looks like is the only use case for ```$ignore``` in iOS).

patedit:
@ryanseams Thanks for the explanation! My mistake thinking that `onDestroySurvey` is also called when the survey was not displayed.

For the `$ignore`, I just wanted to use this open ticket to know why we had that in iOS and what was the reason. Thanks for sharing that info :)

Issue #240, opened 2015-07-24
Problem with rtl layout in version 4.6.0 and later


alamkanak:
Having problem with right-to-left layouts when using versions > 4.5.3. In the screenshot below, the toolbar is supposed to be right to left in both the images but it is left to right when version 4.6.0 is used.

![untitled-1](https://cloud.githubusercontent.com/assets/3194741/8880729/9ceda67c-325c-11e5-9a16-a1591bcceb1b.png)

**Reproduce**
1. Open a new android studio project.
2. [Add rtl support by changing the manifest file](http://android-developers.blogspot.com/2013/03/native-rtl-support-in-android-42.html).
3. Add mixpanel 4.6.0 as a dependency in the `build.gradle` file.
4. Initialize mixpanel in the main activity.
5. Change your device (preferably lollipop) language to, say, Arabic.
6. Run the app in that device. Notice that the toolbar is still left-to-right.
7. Change the mixpanel version to 4.5.3.
8. Run again. Notice that the toolbar is now displaying correctly.

ryanseams:
thanks for bringing this up @alamkanak! i confirmed this issue is occurring in 4.6.+ versions of the library and traced this back to [these commits](https://github.com/mixpanel/mixpanel-android/compare/d643e71630d10287d552f0ee3a1db1774ba489ec...cafd08f9856a9df6c127fb78719c4d884ee4535e)

so sorry about this -- it would appear some commits which were thought to be minor caused unintended effects on the ability to render RTL layouts

in the meantime while we find a fix for this, you can use the following version in your project to use RTL with A/B testing (note this does not contain any bug fixes since v4.6.0 was released):
```compile 'com.mixpanel.android:mixpanel-android:4.5.4-RC2'```

reemsherman:
haaaaaaaaaaaaaaaaaaaaaaaa!
I just spent at least a full day work on this. @alamkanak - thank you so much for this. You just saved my week!

rejasupotaro:
Is it fixed? I'm in trouble with a similar problem...

RTL would be disabled when I call `MixpanelAPI.getInstance`.

<img width="512" alt="screen shot 2016-01-07 at 16 54 31" src="https://cloud.githubusercontent.com/assets/883148/12165218/61e7bffa-b55f-11e5-9cce-8a76a7703e3a.png">

The library version is 4.7.0. Minimum sample project is here: https://github.com/rejasupotaro/RtlWithMixpanel

[...]

PR #310, opened 2016-01-27
Do not update the ApplicationInfo - this fixes the rtl layout issue

yinfeiru:
github issue #240


patedit:
Amazing @yinfeiru

yinfeiru:
What happens is that we accidentally changed the value of ApplicationInfo which overwrites the `FLAG_SUPPORTS_RTL` flag.
http://developer.android.com/reference/android/content/pm/ApplicationInfo.html#FLAG_SUPPORTS_RTL

PR #312, opened 2016-01-30
Scale down the image if there is not enough memory to avoid OOM error

yinfeiru:


src/main/java/com/mixpanel/android/util/ImageStore.java
```diff
@@ -93,7 +93,25 @@ public Bitmap getImage(String url) throws CantGetImageException {
                 throw new CantGetImageException("Downloaded data could not be interpreted as a bitmap");
             }
         } else {
-            bitmap = BitmapFactory.decodeFile(file.getAbsolutePath());
+            Long freeMemory = Runtime.getRuntime().freeMemory();
```
patedit: We should get the `freeMemory` right before we are going to use it.

src/main/java/com/mixpanel/android/util/ImageStore.java
```diff
@@ -93,7 +93,25 @@ public Bitmap getImage(String url) throws CantGetImageException {
...
+            BitmapFactory.Options option = new BitmapFactory.Options();
+            option.inJustDecodeBounds = true;
+            BitmapFactory.decodeFile(file.getAbsolutePath(), option);
+            Long imageSize = new Long(option.outHeight * option.outWidth);
+
+            int inSampleSize = 1;
+            while(imageSize / inSampleSize > freeMemory) {
+                // it doesn't make sense to scale the image further as the image quality will be too bad to be recognized
```
patedit: How can we guarantee this? We're not checking the image height to assure that it's good quality. If the image is very small and `inSampleSize = 2` it may also look very bad already.

yinfeiru:
branch name is wrong. new pr at #315 
closing this

PR #313, opened 2016-02-01
Allow disable viewcrawler master

yinfeiru:


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1324,9 +1328,10 @@ public void onPrefsLoaded(SharedPreferences preferences) {
     }
 
     /* package */ UpdatesFromMixpanel constructUpdatesFromMixpanel(final Context context, final String token) {
-        if (Build.VERSION.SDK_INT < MPConfig.UI_FEATURES_MIN_API) {
-            Log.i(LOGTAG, "Web Configuration, A/B Testing, and Dynamic Tweaks are not supported on this Android OS Version");
-            return new UnsupportedUpdatesFromMixpanel(sSharedTweaks);
+        if (Build.VERSION.SDK_INT < MPConfig.UI_FEATURES_MIN_API
```
patedit: Can we keep both `UnsupportedUpdatesFromMixpanel` and `NoOpUpdatesFromMixpanel`? They mean different things although the result is the same.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -193,11 +193,19 @@
...
      */
     MixpanelAPI(Context context, Future<SharedPreferences> referrerPreferences, String token) {
+        this(context, referrerPreferences, token, MPConfig.getInstance(context));
+    }
+
+    /**
+     * You shouldn't instantiate MixpanelAPI objects directly.
+     * Use MixpanelAPI.getInstance to get an instance.
+     */
+    MixpanelAPI(Context context, Future<SharedPreferences> referrerPreferences, String token, final MPConfig config) {
```
patedit: For consistency, we can just remove `final` modifier. Alternatively, we can add `final` to the rest of the constructor params.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: Though what you said makes sense, I'd like to keep only one of them. Because both of them do exactly the same thing - 1. I don't want to confuse our users by having two different classes but both have the same implement 2. we have duplicated code.

My take is that their names do have different meanings, but the end results are the same - both of them won't initialize ViewCrawler initialization. So, I'd like to keep only `NoOpUpdatesFromMixpanel`.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: ef95d74

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: Fair enough. We should have a proper log message in that case. SDK version is not supported VS manually disabled.

patedit:
Ready to merge!

yinfeiru:
Great! This is huge!

yinfeiru:
@yogurtearl fyi

jpshelley:
Was trying to follow along with issues/requests and still not quite sure how to disable the crawler.
We use the `MixpanelAPI.getInstance(context, token)` method. Is there a method to disable it via the API or do you have to pass in a MPConfig object?

ryanseams:
You can disable the ViewCrawler by adding the following permission to your ```AndroidManifest.xml``` file:

```
<meta-data android:name="com.mixpanel.android.MPConfig.DisableViewCrawler"
    android:value="true" />
```

At this time there is no means to disable the ViewCrawler outside of this permission. Will this work for your needs?

franklinharper:
Thank you for adding this option.

It would be nice if it was easy to find documentation for this configuration option. I couldn't find it anywhere aside from this PR, which seems odd.

takecare:
@ryanseams isn't there another way to do this? i'm afraid if anyone is using mixpanel in a library, doing it only through the manifest, will forcefully disable that feature for the host app as well. which is really bad.

the only solution might be to indicate users of such library that they need to explicitly enable it through:

```
<meta-data android:name="com.mixpanel.android.MPConfig.DisableViewCrawler"
    android:value="true" />
```

is this correct?

ryanseams:
hi @takecare does [this PR](https://github.com/mixpanel/mixpanel-android/pull/389) to allow disabling of the viewcrawler per project token work?

takecare:
don't think so. i'd say that's a different issue. and that solution is still based on meta-data entry.

given it's based on the manifest (as manifests get merged) you can only have one meta-data entry per key. if you define one in a library project's manifest file, it will be overridden by the host app's manifest (if it also defines that meta-data entry).

suppose someone develops a library called 'mylib' which uses mixpanel and they don't want to enable the view crawler. they go in the manifest for the library project and add the meta-data entry disabling it.

now you create an app - 'mysuperapp' - that has 'mylib' as a dependency and also uses mixpanel, but you want to enable the view crawler on the app. two things can happen:

1. you don't add the meta-data entry assuming view crawler is enabled by default, but upon the manifests being merged (the one from your app + the one from the lib) the entry from the lib gets into the final manifest thus disabling view crawler. this is wrong.

2. you're aware of 'mylib' disabling view crawler, so you add the meta-data in your manifest (which prevails over mylib's) but now view crawler will be enabled on mylib's mixpanel instance. this is also wrong.

the way i see it, the only solution would be to allow disabling the view crawler in a way that's not based on entries on the manifest.

PR #315, opened 2016-02-01
Decode image oom bug fix

yinfeiru:


src/main/java/com/mixpanel/android/util/ImageStore.java
```diff
@@ -93,7 +93,18 @@ public Bitmap getImage(String url) throws CantGetImageException {
...
+            option.inJustDecodeBounds = true;
+            BitmapFactory.decodeFile(file.getAbsolutePath(), option);
+            Long imageSize = new Long(option.outHeight * option.outWidth);
+            Long freeMemory = Runtime.getRuntime().freeMemory();
+            if (imageSize > freeMemory) {
+                throw new CantGetImageException("Do not have enough memory for the image");
+            }
+
+            option.inJustDecodeBounds = false;
+            bitmap = BitmapFactory.decodeFile(file.getAbsolutePath(), option);
```
patedit: If we are not changing the `inSampleSize` then we do not need to pass `option` and set `inJustDecodeBounds=false`.

yinfeiru: No, `inJustDecodeBounds = true` allows to get the image info without allocating memory. So, doing `inJustDecodeBounds = false` ensures the next `decodeFile` will load the image into memory.

patedit: I wrote the comment on line 106. Re-read my comment :) We do not need to change the flag and pass `option`.

yinfeiru: oh.. damn.. I totally misunderstood and YOU ARE RIGHT!

yinfeiru: 5ec8fe8

patedit:
LGTM

PR #316, opened 2016-02-01
Do not attempt to send requests to mixpanel if adblocker is enabled

patedit:


src/main/java/com/mixpanel/android/util/HttpService.java
```diff
@@ -27,8 +27,36 @@
...
+
+    private static boolean sAdBlockerEnabled;
+
+    @Override
+    public void checkAdBlockerEnabled() {
+        Thread t = new Thread(new Runnable() {
+            public void run() {
+                try {
+                    InetAddress apiMixpanelInet = InetAddress.getByName("api.mixpanel.com");
+                    InetAddress decideMixpanelInet = InetAddress.getByName("decide.mixpanel.com");
```
yinfeiru: Do we need to check `decide.mixpanel.com`? If it gets blocked, we won't get any notifications/surveys.

yinfeiru: But, yea, it's better to check both.

src/main/java/com/mixpanel/android/util/HttpService.java
```diff
@@ -27,8 +27,36 @@
  * An HTTP utility class for internal use in the Mixpanel library. Not thread-safe.
  */
 public class HttpService implements RemoteService {
+
+    private static boolean sAdBlockerEnabled;
```
yinfeiru: Do you mind renaming `sAdBlockerEnabled` to something more specific? AdBlocker being enabled doesn't mean Mixpanel gets blocked.

src/main/java/com/mixpanel/android/util/HttpService.java
```diff
@@ -27,8 +27,36 @@
...
+                    InetAddress apiMixpanelInet = InetAddress.getByName("api.mixpanel.com");
+                    InetAddress decideMixpanelInet = InetAddress.getByName("decide.mixpanel.com");
+                    sAdBlockerEnabled = apiMixpanelInet.isLoopbackAddress() ||
+                            apiMixpanelInet.isAnyLocalAddress() ||
+                            decideMixpanelInet.isLoopbackAddress() ||
+                            decideMixpanelInet.isAnyLocalAddress();
+                    if (sAdBlockerEnabled ) {
+                        Log.i(LOGTAG, "AdBlocker is enabled. Won't be able to use Mixpanel services.");
+                    }
+                } catch (UnknownHostException e) {
```
yinfeiru: When will this happen? Do you want to disable notifications/surveys in this case?

src/main/java/com/mixpanel/android/util/HttpService.java
patedit: That's the whole point of this PR. We need to check whether we are able to communicate with Mixpanel or not.

src/main/java/com/mixpanel/android/util/HttpService.java
```diff
@@ -27,8 +27,36 @@
...
+                }
+            }
+        });
+
+        t.start();
+    }
+
     @Override
     public boolean isOnline(Context context) {
+        if (sAdBlockerEnabled)
```
yinfeiru: This is picky but can you add `{}` to wrap this block, just to match how we do it in other places?

eg.

```
            if (null == bitmap) {
                throw new CantGetImageException("Downloaded data could not be interpreted as a bitmap");
            }
```

src/main/java/com/mixpanel/android/util/HttpService.java
patedit: I'm relying on `isOnline` for that. That way we do not need to keep track of things that should be enabled or not and implement that logic since we currently do not have those features.

src/main/java/com/mixpanel/android/util/HttpService.java
patedit: The only scenario where I don't include braces is when doing sanity checks at the beginning of a method. It's easier to read. Alternatively, you can also do one-line sanity checks (e.g. https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/UpdateDisplayState.java#L305). I'll go with the later to make it consistent.

yinfeiru:
Have you tried it with AdBlocker on?

patedit:
Yup

PR #317, opened 2016-02-02
Adding flags to disable notifications and surveys to avoid unneeded d…

yinfeiru:
…ecide requests

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
@@ -387,27 +387,27 @@ public class MPConfig {
         }
     }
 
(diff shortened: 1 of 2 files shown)


ryanseams:
these config options will disable a/b tests as well right? why not have one configuration option to just turn off decide rather than multiple options where the end user will not know turning off one also turns off the other features?

yinfeiru:
@ryanseams nice catch and I agree with the idea of turning decide check of.

patedit:
This is not a good solution in general. Besides what @ryanseams said, current implementation will disable everything (tweaks, notifications, surveys...) only by having notifications OR surveys off.

yinfeiru:
duplicate.. closing

PR #318, opened 2016-02-04
Remove tools:ignore gradleoverrides

patedit:
Was included to remove lint warnings. Not needed since it conflicts with existing apps.


yinfeiru:
image

Issue #320, opened 2016-02-05
Crash in ViewCrawler with Mixpanel 4.7.0 and StrictMode


jdauphant:
Hello,

I have crash on the class com.mixpanel.android.viewcrawler.ViewCrawler .

It look's like have different behavior with different device : 
- Sony Xperia Z3 with android 5.1 don't crash 
- it crash on a yotaphone with Android 5.0
- it didn't crash on a Samsung S2 on Android 4.2

The stack : 
```
E/StrictMode: null
          android.app.IntentReceiverLeaked: Activity my.package.ui.LocationActivity has leaked IntentReceiver android.hardware.SystemSensorManager$1@3e827e66 that was originally registered here. Are you missing a call to unregisterReceiver()?
          at android.app.LoadedApk$ReceiverDispatcher.<init>(LoadedApk.java:898)
          at android.app.LoadedApk.getReceiverDispatcher(LoadedApk.java:699)
          at android.app.ContextImpl.registerReceiverInternal(ContextImpl.java:1638)
          at android.app.ContextImpl.registerReceiver(ContextImpl.java:1618)
          at android.app.ContextImpl.registerReceiver(ContextImpl.java:1612)
          at android.content.ContextWrapper.registerReceiver(ContextWrapper.java:488)
          at android.hardware.SystemSensorManager.<init>(SystemSensorManager.java:81)
          at android.app.ContextImpl$31.createService(ContextImpl.java:539)
          at android.app.ContextImpl$ServiceFetcher.getService(ContextImpl.java:297)
          at android.app.ContextImpl.getSystemService(ContextImpl.java:1833)
          at android.view.ContextThemeWrapper.getSystemService(ContextThemeWrapper.java:113)
          at android.app.Activity.getSystemService(Activity.java:5067)
          at com.mixpanel.android.viewcrawler.ViewCrawler$LifecycleCallbacks.installConnectionSensor(ViewCrawler.java:208)
          at com.mixpanel.android.viewcrawler.ViewCrawler$LifecycleCallbacks.onActivityResumed(ViewCrawler.java:180)
          at android.app.Application.dispatchActivityResumed(Application.java:208)
          at android.app.Activity.onResume(Activity.java:1259)
          at android.support.v4.app.FragmentActivity.onResume(FragmentActivity.java:438)
          at my.package.BuildTypeActivity.onResume(BuildTypeActivity.kt:36)
          at my.package.BaseActivity.onResume(BaseActivity.kt:31)
[...]

PR #321, opened 2016-02-05
Use application context instead of activity sensor to remove memory leak on Android<5.1

patedit:
See issue #320


yinfeiru:
lgtm. test and merge! you are so damn good man!

PR #330, opened 2016-02-18
Exit A/B test experiments when they are turned off

ryanseams:
Currently a device will have many difficulties exiting an A/B test -- there are a multitude of conditional statements within the library which never trigger an update when a test is turned off from the Mixpanel UI. Instead, the device will continue to use the most recently cached A/B test experiment.

This PR adds support for leaving an A/B test once it is turned off. It will trigger an update to A/B test info when no new variants are received, and then process this case specially to reset all tweaks to their default values and remove all UI changes.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -30,9 +30,7 @@ import java.util.Set;
         mUnseenNotifications = new LinkedList<InAppNotification>();
         mSurveyIds = new HashSet<Integer>();
         mNotificationIds = new HashSet<Integer>();
-        if(mVariants == null) {
-            mVariants = new JSONArray();
-        }
+        mVariants = new JSONArray();
     }
 
     public String getToken() {
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -30,6 +30,9 @@ import java.util.Set;
         mUnseenNotifications = new LinkedList<InAppNotification>();
         mSurveyIds = new HashSet<Integer>();
         mNotificationIds = new HashSet<Integer>();
+        if(mVariants == null) {
+            mVariants = new JSONArray();
+        }
     }
 
     public String getToken() {
@@ -105,9 +108,9 @@ import java.util.Set;
         }
 
         // in the case we do not receive a new variant, this means the A/B test should be turned off
-        if(!hasNewVariants && newVariantsLength == 0) {
+        if(newVariantsLength == 0) {
             mLoadedVariants.clear();
-            mVariants = null;
+            mVariants = new JSONArray();
             newContent = true;
         }
 
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -77,6 +77,8 @@ import java.util.Set;
         int newVariantsLength = variants.length();
         boolean hasNewVariants = false;
 
+        Log.d("Variants", "New variants length: " + newVariantsLength);
+
         for (int i = 0; i < newVariantsLength; i++) {
             try {
                 JSONObject variant = variants.getJSONObject(i);
@@ -91,6 +93,10 @@ import java.util.Set;
             }
         }
 
+        Log.d("Variants", "New variants? " + hasNewVariants);
+        Log.d("Variants", "Currently loaded variants: " + mLoadedVariants);
+        Log.d("Variants", "Current variants: " + mVariants);
+
         if (hasNewVariants) {
             mLoadedVariants.clear();
 
@@ -104,12 +110,25 @@ import java.util.Set;
             }
         }
 
+        // in case we do not receive a new variant, this means the A/B test should be turned off
+        if(!hasNewVariants && newVariantsLength == 0) {
+            Log.e("Variants", "Attempting to clear due to disabled experiment");
+            mLoadedVariants.clear();
+            mVariants = null;
+            newContent = true;
+        }
+
+        Log.d("Variants", "Currently loaded variants: " + mLoadedVariants);
+        Log.d("Variants", "Current variants: " + mVariants);
+
         if (MPConfig.DEBUG) {
             Log.v(LOGTAG, "New Decide content has become available. " +
                     newSurveys.size() + " surveys, " +
                     newNotifications.size() + " notifications and " +
                     variants.length() + " experiments have been added.");
         }
+
+        Log.d("Variants", "Are we returning an update? " + hasUpdatesAvailable());
 
         if (newContent && hasUpdatesAvailable() && null != mListener) {
             mListener.onNewResults();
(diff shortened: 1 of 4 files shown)


src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -104,6 +104,13 @@ public synchronized void reportResults(List<Survey> newSurveys, List<InAppNotifi
             }
         }
 
+        // in the case we do not receive a new variant, this means the A/B test should be turned off
+        if(!hasNewVariants && newVariantsLength == 0) {
```
patedit: How does this solve the problem? Say we have `variantA` and `variantB` running. If we turn off `variantB`, this logic will  only work if a) we do not add a `variantC` and b) we also remove `variantA`.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1605,9 +1605,7 @@ public void trackNotification(final String eventName, final InAppNotification no
         @Override
         public void joinExperimentIfAvailable() {
             final JSONArray variants = mDecideMessages.getVariants();
-            if (null != variants) {
```
patedit: `variants` could be null. Why don't we check it here?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -754,7 +754,11 @@ private void handleEditorTweaksReceived(JSONObject tweaksMessage) {
         private void handleVariantsReceived(JSONArray variants) {
             final SharedPreferences preferences = getSharedPreferences();
             final SharedPreferences.Editor editor = preferences.edit();
-            editor.putString(SHARED_PREF_CHANGES_KEY, variants.toString());
+            if(variants != null) {
```
patedit: I am not a fan of using `null` as a meaningful value that will change the behavior/flow of the product. For instance, `variants` could be just an empty `JSONArray` if there are really no `variants`. The `SHARED_PREF_CHANGES_KEY` could always be in the device, just with empty variants.

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
ryanseams: AFAIK we cannot ever have two variants running at once as our SDK does not allow for it... If true, the first case of a new variant such as ```variantA``` or ```variantC``` being applied is handled by the original code above this section. This code is specifically for when the response goes from having variants to having nothing, in which case we would want to remove any active variants as they are no longer activated in the Mixpanel UI.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
ryanseams: We could change the code in ```DecideMessages.java``` to set ```variants``` as an empty ```JSONArray``` and then check here if the array is empty. If so, then we could also set the ```SHARED_PREF_CHANGES_KEY``` to be empty. Really it's a choice between null and empty sets in either case as the end code won't change, just the conditional logic to enter the code.

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
patedit: @ryanseams You're right, we can only have one variant. Sorry! Can we just check `if (newVariantsLength == 0)` so we get rid of having a dependency on `hasNewVariants`?

ryanseams:
@patedit Made the changes you requested -- let me know if you see anything else which needs to be changed for this to be merged!

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -30,6 +30,9 @@ public DecideMessages(String token, OnNewResultsListener listener, UpdatesFromMi
         mUnseenNotifications = new LinkedList<InAppNotification>();
         mSurveyIds = new HashSet<Integer>();
         mNotificationIds = new HashSet<Integer>();
+        if(mVariants == null) {
```
patedit: `mVariants` will always be `null` here.

ryanseams: yup yup. removed now.

patedit:
@ryanseams This is awesome. Thanks for doing it :)

fred8:
@ryanseams @patedit 
We ended up by doing the same fix in our branch. However something is missing here: the update listener must be call in order to notify the client of the SDK. When are you planning to ship this release?

patedit:
@fred8 Thanks for you comment! Are you referring to `onNewResults` callback? (https://github.com/mixpanel/mixpanel-android/pull/330/files#diff-e29a313903ebedb460fa0c710b1ae71cR123). If that is the case I think it'll always be called.

But thanks to your comment I realized we definitely have to change something @ryanseams. `hasUpdatesAvailable` will always return `true` now since `mVariants` will never be `null`. Do you think it makes sense to revert some of this (https://github.com/mixpanel/mixpanel-android/pull/330/commits/130a89b7bcb5ee77c12359f5c92b8bca618548f7)?

ryanseams:
We can change back to when I was setting ```mVariants``` as ```null``` so it would no longer call ```hasUpdatesAvailable``` OR we could change the condition to call ```hasUpdatesAvailable``` from ```mVariants != null``` to ```mVariants.length() > 0```.

You had originally shied away from using ```null```, if so, the latter probably makes the most sense... That will resolve the issue with the updates always being reported.

The question here is this if the above is desired behavior: Should the listener be called if the A/B test is being exited? If so this will need more logic and a revert as opposed to a quick fix for when ```hasNewResults``` is called.

fred8:
@ryanseams We just updated to the master version; we're not able to join an experiment anymore. The default value are always picked if the client *never* joined an experiment before (problem in initializeChanges)

For the listener it should be call regardless the changes. We're using it to update the properties and super properties of the user who join the experiment... and so by consequence unjoin...

ryanseams:
@fred8 The master version did not change any logic for joining an experiment so if you are unable to join an experiment that sounds like it would be a different issue. The current code resets to defaults when the API does not return an experiment in the response (any response with an experiment remains unchanged).

The bug right now is that the listener is always called, regardless of what the response from the API is.

fred8:
@ryanseams we just update the sdk to master; it doesn't join the experiment as it used to in 4,8,5. You can repro it very easily.

xlsior:
Hey @ryanseams, with these changes, the tweak value we setup for an experiment always returns the default value and not the value we setup in the experiment. To repro this, you start by creating 1 experiment with 2 variants (we started from blank slate i.e. no existing experiments running, if that makes a difference) and a string tweak variable - set the tweak variable for Original to "A", Variant 1 to "B" and use "C" as the default value of the tweak variable in the app. Target 100% users to the Original variant. When you turn on the experiment, Tweak.get() always returns "C" and not "A".

fred8:
@ryanseams So actually, we did an another try today. On the experiment yesterday we set a target for Operating System : Android with an old user. It seems that this property isn't set by your library or the server anymore. When does that thing change? In production we used to rely on that field. 
cc @xlsior

PR #334, opened 2016-02-23
Flush queue when app goes to background

patedit:


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1264,16 +1264,14 @@ public void logPosts() {
      * This function is automatically called when the library is initialized unless you explicitly
      * set com.mixpanel.android.MPConfig.AutoShowMixpanelUpdates to false in your AndroidManifest.xml
      */
-    @TargetApi(MPConfig.UI_FEATURES_MIN_API)
+    @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
```
yinfeiru: I think it's actually better to have `UI_FEATURES_MIN_API` as we might want to change the minimum API level down the road and we don't want to make changes all over the place.

Also, `UI_FEATURES_MIN_API` is 16 but I think  `ICE_CREAM_SANDWICH` is 14. Why do we want to make the API level lower?

yinfeiru:
lgtm

PR #335, opened 2016-03-01
Do not delete events if response is null

patedit:
Response could be null if we retried more than 3 times, for example.


yinfeiru:
lgtm

PR #336, opened 2016-03-02
Flush all the queue when there are more than 50 events queued

patedit:


src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -392,9 +392,13 @@ private void sendAllData(MPDbAdapter dbAdapter) {
 
             private void sendData(MPDbAdapter dbAdapter, MPDbAdapter.Table table, String[] urls) {
                 final RemoteService poster = getPoster();
-                final String[] eventsData = dbAdapter.generateDataString(table);
+                String[] eventsData = dbAdapter.generateDataString(table);
+                Integer queueCount = 0;
+                if (queueCount != null) {
```
yinfeiru: Why this check?

yinfeiru: Should you check `eventsData`?

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -459,6 +463,12 @@ private void sendData(MPDbAdapter dbAdapter, MPDbAdapter.Table table, String[] u
                         sendEmptyMessageDelayed(FLUSH_QUEUE, mTrackEngageRetryAfter);
                         mFailedRetries++;
                         logAboutMessageToMixpanel("Retrying this batch of events in " + mTrackEngageRetryAfter + " ms");
+                        break;
```
yinfeiru: I'm not certain but do we want to break on `OutOfMemoryError`?

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -459,6 +463,12 @@ private void sendData(MPDbAdapter dbAdapter, MPDbAdapter.Table table, String[] u
...
+                    }
+
+                    eventsData = dbAdapter.generateDataString(table);
+                    if (eventsData != null) {
+                        queueCount = Integer.valueOf(eventsData[2]);
```
yinfeiru: I know it's not your fault but can we give `2` a meaningful name? Everything single time I look at this code I have to figure out what `generateDataString` does.

yinfeiru: (we use magic index numbers all over the place in this file eg. https://github.com/mixpanel/mixpanel-android/pull/336/files#diff-7c00d8762ab328cad50eba1fecb26bc5R402    not very readable)

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -392,9 +392,13 @@ private void sendAllData(MPDbAdapter dbAdapter) {
 
             private void sendData(MPDbAdapter dbAdapter, MPDbAdapter.Table table, String[] urls) {
                 final RemoteService poster = getPoster();
-                final String[] eventsData = dbAdapter.generateDataString(table);
+                String[] eventsData = dbAdapter.generateDataString(table);
+                Integer queueCount = 0;
+                if (queueCount != null) {
+                    queueCount = Integer.valueOf(eventsData[2]);
```
yinfeiru: what if `eventsData[2]` is null

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
yinfeiru: also, please add `eventsData[2] != null` check

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
patedit: My mistake. I was doing something else before with `queueCount`.

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
patedit: `eventsData[2]` can't be `null` if `eventsData` is not `null`.

PR #340, opened 2016-03-25
Add flags to not show survey/notification on activation

SarahAnnTolsma:
This adds functionality similar to the iOS SDK, so developers can
choose not to show a survey or notification when the app activates.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -207,6 +207,9 @@ public class MixpanelAPI {
         mPeople = new PeopleImpl();
         mConfig = config;
 
+        mShowNotificationOnActive = true;
+        mShowSurveyOnActive = true;
+        
         final Map<String, String> deviceInfo = new HashMap<String, String>();
         deviceInfo.put("$android_lib_version", MPConfig.VERSION);
         deviceInfo.put("$android_os", "Android");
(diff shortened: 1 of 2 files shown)


ryanseams:
hi @SarahAnnTolsma ! looking at your PR it looks like your goal here is to not show surveys/notifications automatically when a user opens the app. is there a reason the configuration option ```AutoShowMixpanelUpdates``` (see [here](https://mixpanel.com/help/reference/android#surveys) for more) does not work for your purposes?

by turning ```AutoShowMixpanelUpdates``` to ```false``` surveys/notifications will not show automatically and you can instead use ```MixpanelAPI.getPeople().showSurveyIfAvailable()``` and ```MixpanelAPI.getPeople().showNotificationIfAvailable()``` to manually call to show a survey/notification within your app. let us know if there is a usage case we are missing here!

SarahAnnTolsma:
Hi @ryanseams -- Our usage has general steered away from configuration in the various resource files. We ship the same code-base across Android and iOS (and more recently UWP), so having config separated from code creates a lot of maintenance headache.

Separately, there is currently no way to e.g. _only_ show Survey's on launch. With this change, each experience can be individually toggled.

SarahAnnTolsma:
Along the same lines -- the iOS SDK already has these flags. This ensures the API is consistent across platforms (something we ran into a few times integrating Mixpanel).

patedit:
Thanks a lot for the PR @SarahAnnTolsma! There are a couple of things I'd like to point out with this PR:
- Notice there is no `applicationDidBecomeActive` on Android and that's why we don't have parity with iOS. In iOS if we can guarantee that the survey/notification is not shown on app activation thanks to the app lifecycle callbacks that apple provides.
- This PR relies on the dev to decide whether or not an **activity** should show a survey/notification by manually setting a boolean at the right time and changing it back later at runtime, but not whether a survey should be shown on an app activation automatically. Mixpanel provides that functionality already with `AutoShowMixpanelUpdates` as @ryanseams pointed out. 

Seems that you are concerned about having to configure the library with `meta` tags and, instead, control its behavior at runtime like this PR does. I'd be happy to review a PR that moves `AutoShowMixpanelUpdates` xml configuration to be configurable at runtime. What do you think? Thanks again!

PR #344, opened 2016-04-05
Added Client Offline Mode

juanchospotify:
Added OfflineMode configuration parameter on MPConfig to work in-sync
with client built-in offline mode. When set all the calls to
HttpService.isOnline will check the current Offline mode status of the
client and honor it.

### DIFF
--- a/src/main/java/com/mixpanel/android/util/RemoteService.java
+++ b/src/main/java/com/mixpanel/android/util/RemoteService.java
@@ -10,7 +10,7 @@ import javax.net.ssl.SSLSocketFactory;
 
 
 public interface RemoteService {
-    boolean isOnline(Context context);
+    boolean isOnline(Context context, OfflineMode offlineMode);
 
     void checkIsMixpanelBlocked();
 
--- a/src/androidTest/java/com/mixpanel/android/viewcrawler/ImageStoreTest.java
+++ b/src/androidTest/java/com/mixpanel/android/viewcrawler/ImageStoreTest.java
@@ -7,6 +7,7 @@ import android.test.AndroidTestCase;
 import android.util.Base64;
 
 import com.mixpanel.android.util.ImageStore;
+import com.mixpanel.android.util.OfflineMode;
 import com.mixpanel.android.util.RemoteService;
 
 import java.io.IOException;
@@ -71,7 +72,7 @@ public class ImageStoreTest extends AndroidTestCase {
         }
 
         @Override
-        public boolean isOnline(final Context context) {
+        public boolean isOnline(final Context context, OfflineMode offlineMode) {
             return online;
         }
 
--- a/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
@@ -378,7 +378,7 @@ import javax.net.ssl.SSLSocketFactory;
 
             private void sendAllData(MPDbAdapter dbAdapter) {
                 final RemoteService poster = getPoster();
-                if (!poster.isOnline(mContext)) {
+                if (!poster.isOnline(mContext, mConfig.getOfflineMode())) {
                     logAboutMessageToMixpanel("Not flushing data to Mixpanel because the device is not connected to the internet.");
                     return;
                 }
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideChecker.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideChecker.java
@@ -3,7 +3,6 @@ package com.mixpanel.android.mpmetrics;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.graphics.Bitmap;
-import android.graphics.BitmapFactory;
 import android.graphics.Point;
 import android.os.Build;
 import android.util.Log;
@@ -23,7 +22,6 @@ import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URLEncoder;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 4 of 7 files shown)


src/main/java/com/mixpanel/android/util/RemoteService.java
```diff
@@ -10,7 +10,7 @@
 
 
 public interface RemoteService {
-    boolean isOnline(Context context);
+    boolean isOnline(Context context, OfflineMode offlineMode);
```
patedit: Instead of modifying the interface and force to pass that object every time. Could we do `MPConfig.getInstance(context).getOfflineMode()` inside `HttpService`?

patedit:
I like the way you abstract functionality. However, for other developers this may introduce an extra complexity layer (by having to create a class that implements `OfflineMode`). Do you think it makes sense to have an in-between solution where instead of having an `OfflineMode` object in `MPConfig` we have a boolean instead?

patedit:
@juanchospotify Thanks a lot for the PR. I really like your solution and how clean it is. Thanks as well for taking care of the documentation and modifying the tests! Let me know if what you think about my comments and I'll be happy to include this functionality on the next release :)

src/main/java/com/mixpanel/android/util/RemoteService.java
juanchospotify: If we do that then 
* we hide the dependency that `RemoteService` has now with `OfflineMode`.
* this function will be inconsistent with 'performRequest(String endpointUrl, Map<String, Object> params, SSLSocketFactory socketFactory)'. On this case you also can do `MPConfig.getInstance(context).getSSLSocketFactory()`
* any implementation of `RemoteService` will need to know how to get an `OfflineMode` instance as one of his responsibilities
* We need to enforce the necessity to honor client `OfflineMode` via documentation (javadocs maybe) for implementors of `RemoteService`

Do you think that this drawbacks worth the change ?

juanchospotify:
@patedit Thanks for your comments. I'm glad to contribute to this project.

I have some concerns about changing `OfflineMode` for a `boolean` on `MPConfig`. I am thinking on "**Offline Mode**" as a feature where a client can decide when it make sense to limit the use of network for Mixpanel. Examples of this are:
* internal implementation of offline mode
* use network only on wifi (avoid my app being blamed for mobile data charges)  
* use network only when phone is charging (avoid my app being blamed for battery consumption)
* do not use network while is being used actively by the app (avoid degrade performance while streaming music or video, online gaming, etc) 

The way to manage some of this cases with a `boolean` is listen for state changes on the affected components and keep up to date the boolean value on `MPConfig`. As an addition we need to make this update calls thread safe. 

If we use `OfflineMode` interface we can use the same approach used with a boolean. But also we can use a reactive approach that check the current state only when is needed. It give more freedom to the client to decide when to listen for changes, when to cache values, check first the cheapest components to avoid expensive calls, etc.

Also using `OfflineMode` interface we decouple the setup of Mixpanel from the OfflineMode management.

On my opinion the benefits of having `OfflineMode` worth the lines of code used on `onOfflineMode(OfflineMode)` function.

src/main/java/com/mixpanel/android/util/RemoteService.java
patedit: Hi @juanchospotify ! You have great points and I considered them before writing my comment. I should have explained it better. Would love to hear what you think:

Although this feature is very practical, nobody else has requested it and that's why I wanted to be more hidden (i.e. inside a `RemoteService` implementation and not in the interface itself). Ideally, `RemoteService` assures that `isOnline` is implemented an its implementor would have its own logic. Sadly, we don't let devs give its own `RemoteService` implementation right now and you can only do this on `HttpService` which -I agree- seems a bit odd. Ideally, I would solve this problem by implementing my own `RemoteService` (e.g. `SpotifyHttpService`) that holds a dependency with a feature I want but it's not enforced by the interface.

Since you can't pass your own `RemoteService` implementation the dilemma I had yesterday was whether this should be enforced on the interface level or implement it in `HttpService` and follow up with a new PR that makes having your own `RemoteService` implementation possible.

Personally, I think that it's is a better approach for the good of the project but I am not strong on this based on the current SDK limitations. I won't push it back if you are not happy with it.

patedit:
@juanchospotify I agree with you. I wanted to delegate that functionality to the developer him/herself by making sure to keep the `boolean` up to date using its own implementation so they don't have to know what `OfflineMode` is. 
However, the drawbacks are that his/her code will probably be less clean and we have to hope they have a thread-safe implementation. 

I was not concerned about the extra lines on `onOfflineMode(OfflineMode)` (*) but about the new layer.

Let's keep `OfflineMode` then 👍 

(*) The name looks like a callback :)

src/main/java/com/mixpanel/android/util/RemoteService.java
juanchospotify: Hi @patedit thanks for the comments !  
I agree with you that adding `OfflineMode` on a call to `isOnline(Context)` is odd. Specially when you have it available on `MPConfig` that is a Singleton. But it follows the same pattern used for `SSLSocketFactory` on `byte[] performRequest(String , Map<String, Object> , SSLSocketFactory)` and was why I choose this way.

juanchospotify: On the other hand, remember that if a client can provide his own implementation of `RemoteService`, then you lose the ability to manage the communication with backend services on the most efficient way for both ends (think on bad client implementations that retry every 100ms for 1 hour). 

Also, currently `RemoteService` has 3 responsabilities:
1.  check if mixpanel is blocked
2. determine when `RemoteService` is online 
3. perform mixpanel request and return the response

It is a lot to take care as a client if the only thing that I want to decide is when to allow network access to the API. As a suggestion I would change `Remote`the interface to: 

```
public interface RemoteService {
    boolean isOnline();
    byte[] performRequest(String endpointUrl, Map<String, Object> params) throws Exception;
}
```
if `Context`, `SSLSocketFactory` and `OfflineMode` are dependencies, they should be required for the implementor avoiding enforcing this on the api.

juanchospotify: As conclusion. I do not have a strong opinion about enforce `OfflineMode` on the interface or use it directly by the implementors. But I would recommend that for consistency `SSLSocketFactory` should use the same approach: or both are present on the interface, or both should be retrieved via `MPConfig`

@patedit what do you think ?

patedit:
Hi @juanchospotify ! I went ahead and merge the PR so it's now available in the latest release (4.9.6). Thanks a lot for your high quality PR :)

juanchospotify:
Hi Sergio, Thank you for take this into consideration. We will update to
this release as soon as possible :)

On Mon, Apr 25, 2016 at 6:40 PM, Sergio Alonso Fernández <
notifications@github.com> wrote:

> Hi @juanchospotify <https://github.com/juanchospotify> ! I went ahead and
> merge the PR so it's now available in the latest release (4.9.6). Thanks a
> lot for your high quality PR :)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/mixpanel/mixpanel-android/pull/344#issuecomment-214433842>
>



-- 
Juan Manuel Rivero
Android Engineer
Revenue Alliance

PR #347, opened 2016-04-13
include  mp_campaign_id in notification intent

nguyenkims:


patedit:
Hi @nguyenkims - we will now be able to track push opens with https://github.com/mixpanel/mixpanel-android/pull/430
Thanks for your PR!

PR #356, opened 2016-04-28
new abgestures tracking

yarneo:


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2109,6 +2111,84 @@ private static void checkIntentForInboundAppLink(Context context) {
...
+                    }
+
+                    mFirstToSecondFingerDifference = -1;
+                    mSecondFingerTimeDown = System.currentTimeMillis();
+                } else if(event.getActionMasked() == MotionEvent.ACTION_UP) {
+                    resetGesture();
+                    return false;
+                }
+
+                return true;
```
patedit: We should always return `false` so the gesture is passed to other listeners.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2109,6 +2111,84 @@ private static void checkIntentForInboundAppLink(Context context) {
         }
     }
 
+    public void trackGestures(Activity parent) {
```
patedit: `trackGestures` should not be on `MixpanelAPI`. Can you come up with a different file organization?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2109,6 +2111,84 @@ private static void checkIntentForInboundAppLink(Context context) {
...
+                }
+
+                if(event.getActionMasked() == MotionEvent.ACTION_POINTER_UP && mFirstToSecondFingerDifference == 100) {
+                    mFirstToSecondFingerDifference = System.currentTimeMillis();
+                } else if (event.getActionMasked() == MotionEvent.ACTION_POINTER_UP) {
+                    resetGesture();
+                    return false;
+                }
+
+                if(event.getActionMasked() == MotionEvent.ACTION_UP && (System.currentTimeMillis() - mFirstToSecondFingerDifference) < 100 && mFirstToSecondFingerDifference > 0){
```
patedit: I don't see why you need to check `mFirstToSecondFingerDifference > 0` if you previously check `System.currentTimeMillis() - mFirstToSecondFingerDifference) < 100`.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2109,6 +2111,84 @@ private static void checkIntentForInboundAppLink(Context context) {
...
+                    return false;
+                }
+
+                if(event.getActionMasked() == MotionEvent.ACTION_DOWN) {
+                    mFirstToSecondFingerDifference = System.currentTimeMillis();
+                }
+
+                if(event.getActionMasked() == MotionEvent.ACTION_POINTER_DOWN && (System.currentTimeMillis() - mFirstToSecondFingerDifference) < 100) {
+                    mSecondFingerTimeDown = System.currentTimeMillis();
+                    mFirstToSecondFingerDifference = 100;
```
patedit: This magic number should be externalized since you rely on `100` later on several times.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2109,6 +2111,84 @@ private static void checkIntentForInboundAppLink(Context context) {
...
+                }
+
+                if(event.getActionMasked() == MotionEvent.ACTION_UP && (System.currentTimeMillis() - mFirstToSecondFingerDifference) < 100 && mFirstToSecondFingerDifference > 0){
+                    if ((System.currentTimeMillis() - mSecondFingerTimeDown) >= 2500) {
+                        if (mGestureSteps == 3) {
+                            track("$ab_gesture1");
+                            resetGesture();
+                        }
+                        mGestureSteps = 0;
+                        //long double-press action
```
patedit: Should be moved to line 2157

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yarneo: because it could be reset and become -1 and i dont want to continue the transaction if that is the case

patedit: Exactly. `System.currentMillis()-mFirstToSecondFingerDifference) < 100` already does that.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2109,6 +2111,84 @@ private static void checkIntentForInboundAppLink(Context context) {
...
+
+            @Override
+            public boolean onTouch(View v, MotionEvent event) {
+                if(event.getPointerCount() > 2) {
+                    resetGesture();
+                    return false;
+                }
+
+                if(event.getActionMasked() == MotionEvent.ACTION_DOWN) {
+                    mFirstToSecondFingerDifference = System.currentTimeMillis();
```
patedit: `mFirstToSecondFingerDifference` does not seem to have a difference.

src/main/java/com/mixpanel/android/viewcrawler/GestureTracker.java
```diff
@@ -0,0 +1,91 @@
...
+									mMixpanel.track("$ab_gesture2");
+									resetGesture();
+								} else {
+									resetGesture();
+								}
+							}
+						}
+						break;
+				}
+				return true;
```
patedit: Why are you returning `true`?

src/main/java/com/mixpanel/android/viewcrawler/GestureTracker.java
```diff
@@ -0,0 +1,91 @@
...
+			}
+
+			private long mSecondFingerTimeDown = System.currentTimeMillis();
+			private long mFirstToSecondFingerDifference = -1;
+			private int mGestureSteps = 0;
+			private long mBetweenSteps = -1;
+			private final int TimeBetweenFingersThreshold = 100;
```
patedit: Constants follow a different naming/style. Eg. `private static final int TIME_BETWEEN_FINGERS_THRESHOLD`.

src/main/java/com/mixpanel/android/viewcrawler/GestureTracker.java
```diff
@@ -0,0 +1,91 @@
...
+
+import android.app.Activity;
+import android.view.MotionEvent;
+import android.view.View;
+import com.mixpanel.android.mpmetrics.MixpanelAPI;
+
+/**
+ * Tracking ABTesting Gestures
+ */
+public class GestureTracker {
```
patedit: File is not correctly indented.

src/main/java/com/mixpanel/android/viewcrawler/GestureTracker.java
```diff
@@ -0,0 +1,91 @@
...
+								}
+							}
+						}
+						break;
+				}
+				return true;
+			}
+
+			private long mSecondFingerTimeDown = System.currentTimeMillis();
+			private long mFirstToSecondFingerDifference = -1;
```
patedit: `mFirstToSecondFingerDifference` holds both a time difference and a time. Quite confusing.

src/main/java/com/mixpanel/android/viewcrawler/GestureTracker.java
```diff
@@ -0,0 +1,91 @@
...
+			private int mGestureSteps = 0;
+			private long mBetweenSteps = -1;
```
patedit: Can we name `mBetweenSteps` in a better way? Not sure what it represents.

src/main/java/com/mixpanel/android/viewcrawler/GestureTracker.java
```diff
@@ -0,0 +1,91 @@
...
+	public GestureTracker(MixpanelAPI mMixpanel, Activity parent) {
+		trackGestures(mMixpanel, parent);
+	}
+
+	private void trackGestures(final MixpanelAPI mMixpanel, Activity parent) {
+		parent.getWindow().getDecorView().findViewById(android.R.id.content).setOnTouchListener(new View.OnTouchListener() {
+
+			private void resetGesture() {
+				mFirstToSecondFingerDifference = -1;
+				mSecondFingerTimeDown = System.currentTimeMillis();
```
patedit: Setting the second finger time down as the current time when we are resetting/initializing the variable (and not when the user actually interacts with the view) looks like a flaw in the below logic.

src/main/java/com/mixpanel/android/viewcrawler/GestureTracker.java
yarneo: because it resets the ACTION_DOWN_POINTER and all the gesture flow entirely because you don't consume the event if you return false

src/main/java/com/mixpanel/android/viewcrawler/GestureTracker.java
yarneo: not really, it doesn't really matter what it is intiialized to at this point.

patedit: Then I don't see the point of initializing a variable to the current time in ms. Let's change it :)

src/main/java/com/mixpanel/android/viewcrawler/GestureTracker.java
patedit: Right. Will other listeners be able to capture the touch? Seems we are always returning `true` if `getPointerCount` is less or equal than 2.

yarneo:
onTouch now doesn't consume touch, and using the Decor view of the window to be the last on the hierarchy to get the touch.

PR #359, opened 2016-05-05
ab testing and codeless connectivity indicator in device

yarneo:


src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -221,27 +226,14 @@ private void uninstallConnectionSensor(final Activity activity) {
...
-                return false;
-            }
-
-            if (!Build.MODEL.toLowerCase(Locale.US).contains("sdk")) {
-                return false;
-            }
-
-            return true;
+            return Build.FINGERPRINT.startsWith("generic")
+                    || Build.FINGERPRINT.startsWith("unknown")
```
patedit: `Build.FINGERPRINT` or `Build.MANUFACTURER` ?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -995,7 +986,6 @@ public JSONObject update(JSONObject in) {
                     Log.wtf(LOGTAG, "Could not build JSON for reporting experiment start", e);
                 }
```
patedit: We can remove this line as well.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -221,27 +226,14 @@ private void uninstallConnectionSensor(final Activity activity) {
...
-
-            return true;
+            return Build.FINGERPRINT.startsWith("generic")
+                    || Build.FINGERPRINT.startsWith("unknown")
+                    || Build.MODEL.contains("google_sdk")
+                    || Build.MODEL.contains("Emulator")
+                    || Build.MODEL.contains("Android SDK built for x86")
+                    || Build.MANUFACTURER.contains("Genymotion")
+                    || (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic"))
+                    || "google_sdk".equals(Build.PRODUCT);
```
patedit: Can we revert this to `Build.PRODUCT.equals("google_sdk)"` ?

src/main/java/com/mixpanel/android/viewcrawler/EditorConnection.java
```diff
@@ -42,6 +42,7 @@ public EditorConnectionException(Throwable cause) {
         public void setTweaks(JSONObject message);
         public void sendDeviceInfo();
         public void cleanup();
+        public void updateWebSocketStatus(Boolean state);
```
patedit: Why do you wrap it in a `Boolean` and not a `boolean`?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -1098,6 +1103,8 @@ public VariantTweak(JSONObject aTweak, Pair<Integer, Integer> aVariantId) {
     private final MixpanelAPI mMixpanel;
     private final DynamicEventTracker mDynamicEventTracker;
     private final EditState mEditState;
+    private Activity mCurrentActivity;
+    public static Boolean mMixpanelInEditMode = false;
```
patedit: Why do you wrap the boolean in an object?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -175,18 +175,23 @@ public void onActivityCreated(Activity activity, Bundle bundle) {
 
         @Override
         public void onActivityStarted(Activity activity) {
+            ConnectivityIndicator.start(mContext);
+            mCurrentActivity = activity;
```
patedit: Why do you we need to keep a reference to the activity here? This will leak the app. We should use `mEditState` array instead.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -175,18 +175,23 @@ public void onActivityCreated(Activity activity, Bundle bundle) {
...
         }
 
         @Override
         public void onActivityResumed(Activity activity) {
             installConnectionSensor(activity);
             mEditState.add(activity);
+            mCurrentActivity = activity;
+            ConnectivityIndicator.show(activity);
```
patedit: It's better to check for `mMixpanelInEditMode` here. `show` should show a view and it should not be it's responsibility to check whether there is a socket open or not, etc..

src/main/java/com/mixpanel/android/viewcrawler/ConnectivityIndicator.java
```diff
@@ -0,0 +1,201 @@
...
+import android.widget.FrameLayout;
+import android.widget.ImageView;
+
+import com.mixpanel.android.R;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ConnectivityIndicator
+        extends FrameLayout
```
patedit: This file needs to be reviewed in terms of the brackets use 👍

src/main/java/com/mixpanel/android/viewcrawler/ConnectivityIndicator.java
```diff
@@ -0,0 +1,201 @@
...
+{
+    private static final int MINIMIZED_DIAMETER = 100;
+    private static ConnectivityIndicator sInstance;
+    private WindowManager windowManager;
+    private WindowManager.LayoutParams params;
+    private int currentX = -1;
```
patedit: `private` variables should start with `m` and next letter is capitalized.

src/main/java/com/mixpanel/android/viewcrawler/ConnectivityIndicator.java
```diff
@@ -0,0 +1,201 @@
...
+import android.view.WindowManager;
+import android.widget.FrameLayout;
+import android.widget.ImageView;
+
+import com.mixpanel.android.R;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ConnectivityIndicator
```
patedit: Have you tried to rotate the device while the indicator is on screen and move it around?

src/main/java/com/mixpanel/android/viewcrawler/ConnectivityIndicator.java
```diff
@@ -0,0 +1,201 @@
...
+            private float initialTouchX;
+            private float initialTouchY;
+            private boolean isDragEvent=false;
+
+            public boolean onTouch(View v,MotionEvent event)
+            {
+                if((event == null) || (ConnectivityIndicator.this.params == null) || (ConnectivityIndicator.this.windowManager == null)){
+                    return false;
+                }
+                System.out.println(event.getAction() + " , " + event.getRawX() + " , " + event.getRawY());
```
patedit: Remove

src/main/java/com/mixpanel/android/viewcrawler/ConnectivityIndicator.java
```diff
@@ -0,0 +1,201 @@
...
+        iconView.setImageResource(R.drawable.com_mixpanel_android_connectivityabtest);
+        addView(iconView);
+    }
+
+    public static void show(final Activity activity)
+    {
+        if ((sInstance == null) || (!ViewCrawler.mMixpanelInEditMode) || activity == null) {
+            return;
+        }
+        if (sInstance.isShown) {
```
patedit: No need to have this variable. If this view has no parent it means is not shown :)

src/main/java/com/mixpanel/android/viewcrawler/ConnectivityIndicator.java
```diff
@@ -0,0 +1,201 @@
...
+        HashMap<String, Integer> screenMap = new HashMap<>();
+        switch (orientation)
+        {
+            case 1:
+                screenMap.put("height", metrics.heightPixels);
+                screenMap.put("width", metrics.widthPixels);
+                break;
+            case 2:
+                screenMap.put("height", metrics.widthPixels);
+                screenMap.put("width", metrics.heightPixels);
```
patedit: Let's add a `break` here.

src/main/java/com/mixpanel/android/viewcrawler/ConnectivityIndicator.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    public static Map<String, Integer> getScreenSizeMap(Context context)
+    {
+        DisplayMetrics metrics = context.getResources().getDisplayMetrics();
+        int orientation = context.getResources().getConfiguration().orientation;
+        HashMap<String, Integer> screenMap = new HashMap<>();
+        switch (orientation)
+        {
+            case 1:
```
patedit: Can we use `ORIENTATION_PORTRAIT`/`ORIENTATION_LANDSCAPE`?

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
```diff
@@ -175,18 +175,23 @@ public void onActivityCreated(Activity activity, Bundle bundle) {
 
         @Override
         public void onActivityStarted(Activity activity) {
+            ConnectivityIndicator.start(mContext);
```
patedit: I don't see why we need to initialize `ConnectivityIndicator` every time an activity starts. It should only be initialized when we need to use it. (i.e. `onActivityResumed` if we are on edit mode and want to show the indicator). `ConnectivityIndicator.getInstance.show()` when the app resumes seems like a better approach.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: FINGERPRINT

src/main/java/com/mixpanel/android/viewcrawler/ConnectivityIndicator.java
yarneo: i did!

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
patedit: Lol. I have seen people using Manufacturer instead. Hence my comment.

yarneo: https://github.com/facebook/react-native/blob/235b16d93287061a09c4624e612b5dc4f960ce47/ReactAndroid/src/main/java/com/facebook/react/modules/systeminfo/AndroidInfoHelpers.java

http://stackoverflow.com/questions/2799097/how-can-i-detect-when-an-android-application-is-running-in-the-emulator

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: yup!

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: it is initialized only once. It needs to be initialized first thing and then it is shown only when it needs to be shown. We could potentially move it to onActivityResume, but it would just be called more times and make it slower

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: mEditState is a HashSet and therefore is unordered. I need to keep track of most recent active Activity

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: it is better to have the check inside show as then in future when people reuse the method it won't crash or cause issues in the app. Because this check needs to be done everytime show is issued

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
patedit: On the stackoverflow post there are 4 people suggesting `Build.MANUFACTURER.equals("unknown")`. Also, this library does it as well: https://github.com/gingo/android-emulator-detector/blob/master/EmulatorDetectorProject/EmulatorDetector/src/main/java/net/skoumal/emulatordetector/EmulatorDetector.java#L41

I think its worth to include it.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
patedit: If we call `ConnectivityIndicator.getInstance.show()` on `onActivityResumed` if we are on editing mode we won't have to initialize it unnecessarily here.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: I think it's too much of an overkill, we should decide of a standard to use. I took mine from that stackoverflow question which seems legit. I think we shouldn't merge too many options. I don't mind what we take but let's keep it clean

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
patedit: A view should not contain any app state. Calling `show` on any view should just show the view. Plus the state of this flag won't make the app to crash. We only use this flag to know whether the `ViewCrawler` should show this view or not. Hence, this should be outside the view and delegate this task to the `ViewCrawler`.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
patedit: You already included 9 options. Not sure what the problem is here.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: It will be initialized in any case, if its onActivityResumed or onActivityStarted. The difference is that onActivityStarted will be called less times and the item will be initialized only once in any case

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: that we might be doing it wrong, im not sure what all the usecases are but im sure what i have right now consists of your check already

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: I disagree, but I'll bite

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: Want to use your example instead of mine? https://github.com/gingo/android-emulator-detector

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
patedit: Again, we should only initialize it if we have to. You created a dependency on the singleton pattern which makes it not singleton basically by having to call `start` before `show` every time.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
patedit: You could either create a sorted set, nullify the activity or create a weak reference. This will leak the app.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
patedit: That is how proper view-controller design works.

src/main/java/com/mixpanel/android/viewcrawler/ViewCrawler.java
yarneo: The dependency already exists. I don't mind moving start to mixpanelAPI.m if that makes it better, but moving it to resume instead of start doesnt make sense to me

PR #360, opened 2016-05-06
inapp fix for image going over text in certain devices

yarneo:


patedit:
Looks good to me. @ryanseams

ryanseams:
Tested on simulator with 5 devices of different screen sizes and the results look good to me.

PR #361, opened 2016-05-10
Reset A/B tests on reset() call

ryanseams:
Currently A/B test data is not refreshed on a ```reset()``` call from ```MixpanelAPI```. Since reset is supposed to be a fresh instance of MixpanelAPI, the A/B test data from the old user should also be removed.

Let me know if I should be calling to ```mDecideMessages``` directly as opposed to using the ```identify()``` method. The line that needs to trigger to reset the experiment data is [here](https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java#L406).

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -714,10 +714,11 @@ public class MixpanelAPI {
      * Will not clear referrer information.
      */
     public void reset() {
-        // Will clear distinct_ids, superProperties,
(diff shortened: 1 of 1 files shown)


ryanseams:
@patedit Anything else needed here or is this okay to go in the next release?

ryanseams:
@patedit I added a ```flush()``` to the ```reset()``` call to ensure it grabs a new /decide response once the user is reset. In addition, the description of the method has been updated to note all data which is cleared by this method.

Let me know if you see anything else needed for this or think there is a better way to do this.

patedit:
I think this is great, Ryan!

PR #378, opened 2016-07-21
Add remove people operator + Ability to remove single push device token

ryanseams:
Add support for the ```$remove``` operator which is detailed [here](https://mixpanel.com/help/reference/http#people-analytics-updates). Using this ```$remove``` operator, an additional method ```clearSinglePushRegistrationId``` has been added to allow for a single push notification device token to be removed from a user's profile.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -1728,15 +1728,8 @@ public class MixpanelAPI {
 
         @Override
         public void clearSinglePushRegistrationId(String registrationId) {
-            // Must be thread safe, will be called from a lot of different threads.
-            synchronized (mPersistentIdentity) {
-                if (mPersistentIdentity.getPeopleDistinctId() == null) {
-                    return;
-                }
-
-                mPersistentIdentity.clearPushId();
-                remove("$android_devices", registrationId);
-            }
+            remove("$android_devices", registrationId);
+            mPersistentIdentity.clearPushId();
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1691,6 +1727,12 @@ public void clearPushRegistrationId() {
         }
 
         @Override
+        public void clearSinglePushRegistrationId(String registrationId) {
+            remove("$android_devices", registrationId);
+            mPersistentIdentity.clearPushId();
```
patedit: Why would we want to clear the current push ID as well? Current one may not be `registrationId`. Can we check whether is the same one or not before clearing the shared preferences? https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java#L1686

ryanseams: Totally correct, would only want to wipe if the token passed is the same one stored on disk. Changing now.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1691,6 +1727,14 @@ public void clearPushRegistrationId() {
         }
 
         @Override
+        public void clearPushRegistrationId(String registrationId) {
+            if(mPersistentIdentity.getPushId() == registrationId) {
```
patedit: can we check for value equality and not reference? `equals`. thx!

PR #381, opened 2016-08-16
add android related properties to decide requests

patedit:


entombedvirus:
that was fast! is it possible to wrap all the extra properties as a JSON object? That would make the implementation on par with what the iOS SDK is doing.

i.e.

`&properties={"$android_lib_version": "X.Y.Z", ...}`

instead of

`&$android_lib_version=X.Y.Z&...`

patedit:
Let me know what you think!

PR #383, opened 2016-08-24
Master

crysxd:
- Made sSharedTweaks publicly available
- Added OnTweakUpdatedListener
- Auto declaring unknown tweaks instead of discarding them

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -189,6 +189,15 @@ public class MixpanelAPI {
     }
 
     /**
+     * Returns the globally shared {@link Tweaks}.
+     * @return the globally shared {@link Tweaks}
+     */
+    public static Tweaks getTweaks() {
+        return sSharedTweaks;
+
+    }
+
+    /**
      * You shouldn't instantiate MixpanelAPI objects directly.
      * Use MixpanelAPI.getInstance to get an instance.
      */
--- a/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
@@ -43,6 +43,8 @@ public class Tweaks {
             this.declareTweak(tweakName, value, this.detectTweakType(value));
         }
 
+        Log.i(LOGTAG, "Updated tweak with name \"" + tweakName + "\": " + value);
+
         final TweakValue container = mTweakValues.get(tweakName);
         final TweakValue updated = container.updateValue(value);
         mTweakValues.put(tweakName, updated);
--- a/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
@@ -1,5 +1,6 @@
 package com.mixpanel.android.mpmetrics;
 
+import android.support.annotation.BoolRes;
 import android.support.annotation.IntDef;
 import android.util.Log;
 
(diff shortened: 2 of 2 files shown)


airbone42:
:+1:

crysxd:
Discarding unkown tweaks when receiving an update can cause misbehaving when the tweak was not locally created before the server sends an update. If the tweak is requested first when it is actually needed, the default value is always returned. By auto-declaring unknown tweaks they are already available when first needed.

OnTweakUpdatedListener allows custom handling of tweak changes, e.g. caching them in a persistent store so they are available immediately after restarting the app / reboot.

airbone42:
Mixpanel are you considering to merge this and release it in an upcoming update?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -189,6 +189,15 @@
     }
 
     /**
+     * Returns the globally shared {@link Tweaks}.
+     * @return the globally shared {@link Tweaks}
+     */
+    public static Tweaks getTweaks() {
```
yinfeiru: Why do you want to expose `sSharedTweaks`?

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -33,18 +34,88 @@ public synchronized void addOnTweakDeclaredListener(OnTweakDeclaredListener list
...
+        Object oldValue = container.value;
         final TweakValue updated = container.updateValue(value);
+        Object newValue = updated.value;
         mTweakValues.put(tweakName, updated);
+
+        if(autoDeclared || !oldValue.equals(newValue)) {
+
+            Log.i(LOGTAG, "Updated tweak with name \"" + tweakName + "\": " + updated.getStringValue());
+
+            for (OnTweakUpdatedListener listener : mTweakUpdatedListeners) {
```
yinfeiru: Why do you want to call the listener here? If you have multiple `Tweak`s, the listener will get called when you update the first `Tweak`, while all the other `Tweak`s are having the default value.

yinfeiru:
@crysxd Thanks for the pr! I actually didn't see this until I finished mine.. https://github.com/mixpanel/mixpanel-android/pull/390

Can you please reply to my comments when you get a chance? I'd like to know why you did them. :)

patedit:
Friendly ping @crysxd  :)

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
crysxd: Sure, but the listener is called every time any value was change. In our case the Activity is listening for the desired Tweak to arrive (it checks for tweak name). We only need one tweak for our Login A/B test, so it's fine for the other Tweaks to have the default value

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
crysxd: This is necessary to attach the listener

crysxd:
Sorry for the late response. At first, we don't use Mixpanel anymore to assign a device to A/B/C groups anymore. We had too many issues with the configuration of the A/B tests and fetching the A/B/C value from the server.

We already found a bug with these changes. When configuring two independed A/B tests, all developer tweaks of the first test will appear in the second test as the device is auto declaring them. Mixpanel seems to show all tweaks used by the device in the session for configuring the test.

patedit:
Closing this PR in favor of https://github.com/mixpanel/mixpanel-android/pull/390. Let us know if you fixed the issues and you want to expose more functionality with the existing implementation :) Thanks guys!

PR #386, opened 2016-08-30
Push notification enhacements

patedit:
The following keys can now be used as part of a push notification payload:

- `mp_icnm_l` Large icon to be used on a push notification.
- `mp_icnm_w` White icon to be used on Lollipop and above.
- `mp_color` Color to be used on Lollipop and above (#argb).

Example of a payload:

`{"mp_icnm":"an_icon", "mp_icnm_l":"big_icon", "mp_icnm_w":"white_icon","mp_color":"#FFAA0000"}`

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
@@ -2,6 +2,8 @@ package com.mixpanel.android.mpmetrics;
 
 import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
+import android.graphics.BitmapFactory;
+import android.graphics.Color;
 import android.support.v4.app.NotificationCompat;
 import android.app.Notification;
 import android.app.NotificationManager;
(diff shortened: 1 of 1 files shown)


src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
```diff
@@ -256,22 +290,27 @@ private void handleNotificationIntent(Context context, Intent intent) {
 
     @SuppressWarnings("deprecation")
     @TargetApi(9)
-    private Notification makeNotificationSDKLessThan11(Context context, PendingIntent intent, NotificationData notificationData) {
+    protected Notification makeNotificationSDKLessThan11(Context context, PendingIntent intent, NotificationData notificationData) {
```
yinfeiru: why do we want to make it `protected`?

patedit: To let developers override this class and create their own notification!

Issue #357, opened 2016-05-05
Tweak.get() always returns the default value


xlsior:
This happens when the app uses multiple instances of the MixpanelAPI, pointing to separate MixPanel projects and only has an experiment running on one of them. This is easily reproducible in the test-application (https://github.com/mixpanel/mixpanel-android/tree/master/acceptance/test-application). Here are the repro steps:

1. You will need to setup two MixPanel projects - Project 1 and Project 2.

2. Make these changes to the sample app:

    Declare a tweak variable and an API key for Project 2
    ```
public static Tweak<String> sTweakTest = MixpanelAPI.stringTweak(
        "Test",
        "default"
    );

public static final String MIXPANEL_API_TOKEN2 = "project 2 token here";
    ```
    Instantiate MixpanelAPI in onCreate()
    ```
mMixpanel = MixpanelAPI.getInstance(this, MIXPANEL_API_TOKEN); 
mMixpanel2 = MixpanelAPI.getInstance(this, MIXPANEL_API_TOKEN2); // add this line for Project #2
    ```
    flush in onDestroy()
    ```
mMixpanel.flush();
mMixpanel2.flush(); // add this line
    ```

3. Create an experiment in Project 1. Set the "Test" tweak in Original to "A", and in Variant 1 to "B". Target Original to 100% of users.

4. sTweakTest.get() always returns "default".

ryanseams:
Whenever you call to grab a Tweak value from ```MixpanelAPI```, it is grabbing that value from the most recently called instance of Mixpanel. As such, the order of the ```getInstance()``` calls matters in terms of applying the right Tweak values.

Is it possible to reverse the order of ```getInstance()``` to initialize ```mMixpanel2``` before ```mMixpanel```? This should result in the behavior you are looking for, where you have multiple Mixpanel instances but the Tweaks are applied based off the of the ```mMixpanel``` values.

devorchik:
The order we are initializing in is correct (anonymous, production) with the tweak variables hanging off production.  I think the underlying bug is when the anonymous result arrives after the production response (from the server), and then becomes data used for the tweaks.

ryanseams:
You are correct -- the issue here is the most recent response from the Mixpanel /decide server is being applied for the Tweak values, regardless of the project. This is made worse by the fact there is no way to specify a Tweak value on a per project basis due to the way Tweaks are defined as global objects of ```MixpanelAPI``` and not tied to any specific project.

[...]

PR #389, opened 2016-09-01
A feature that allows the user to disable view crawler for certain projects

yinfeiru:
https://github.com/mixpanel/mixpanel-android/issues/357

When a user has multiple instances of MixpanelAPI, the A/B changes might get messed up as the lib always applies the change in the last decide response.

This feature is for disabling A/B and codeless for certain projects that are specified by the user.

To use this feature, the user will need to define
```
<resources>
    <string-array name="my_project_list">
        <item>project token 1</item>
        <item>project token 2</item>
    </string-array>
</resources>
```
in `array.xml`.

Then reference it in the AndroidManifest
```
        <meta-data android:name="com.mixpanel.android.MPConfig.DisableViewCrawlerForProjects"
            android:resource="@array/my_project_list" />
```

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
@@ -258,6 +258,13 @@ public class MPConfig {
         }
         mEditorUrl = editorUrl;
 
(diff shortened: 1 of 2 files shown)


patedit:
LGTM

PR #390, opened 2016-09-02
Tweak updated listener

yinfeiru:


src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -185,7 +200,8 @@ public Boolean getBooleanValue() {
 
         public final @TweakType int type;
 
-        private final Object value;
+        protected final Object value;
+
```
patedit: remove

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -121,6 +131,11 @@ public TweakValue updateValue(Object newValue) {
             return new TweakValue(type, defaultValue, minimum, maximum, newValue);
         }
 
+        @Override
+        public boolean equals(Object value) {
```
patedit: Not needed

PR #405, opened 2016-11-17
Add flag to ignore invisible views on mixpanel visual editors

patedit:
With `com.mixpanel.android.MPConfig.IgnoreInvisibleViewsVisualEditor` set to true on your `AndroidManifest.xml` the `ViewCrawler` will filter out invisible views.


yinfeiru:
nice!

PR #410, opened 2016-12-14
Deprecate MP Surveys

patedit:


src/main/java/com/mixpanel/android/surveys/SurveyActivity.java
```diff
@@ -57,9 +57,12 @@
  * You should not send Intent's directly to display this activity. Instead use
  * {@link com.mixpanel.android.mpmetrics.MixpanelAPI.People#showSurveyIfAvailable(Activity)} and
  * {@link com.mixpanel.android.mpmetrics.MixpanelAPI.People#showNotificationIfAvailable(Activity)}
+ *
+ * @drepecate Surveys are deprecated and will be removed from the SDK soon.
  */
 @TargetApi(MPConfig.UI_FEATURES_MIN_API)
 @SuppressLint("ClickableViewAccessibility")
+@Deprecated
```
yinfeiru: Do we want to mark this class as deprecated? It's used by in-app as well.

Issue #408, opened 2016-11-25
ConcurrentModificationException on Tweaks.getDefaultValues()


RomainPiel:
We've had a few crashes with this stacktrace since we've updated Mixpanel to version 4.9.2. We have a few tweaks in place but non that should involve the view crawler.

```
java.util.ConcurrentModificationException
       at java.util.HashMap$HashIterator.nextEntry(HashMap.java:788)
       at java.util.HashMap$EntryIterator.next(HashMap.java:825)
       at java.util.HashMap$EntryIterator.next(HashMap.java:823)
       at java.util.HashMap.constructorPutAll(HashMap.java:208)
       at java.util.HashMap.<init>(HashMap.java:196)
       at com.mixpanel.android.mpmetrics.Tweaks.getDefaultValues(Tweaks.java:71)
       at com.mixpanel.android.viewcrawler.ViewCrawler$ViewCrawlerHandler.applyVariantsAndEventBindings(ViewCrawler.java:903)
       at com.mixpanel.android.viewcrawler.ViewCrawler$ViewCrawlerHandler.initializeChanges(ViewCrawler.java:447)
       at com.mixpanel.android.viewcrawler.ViewCrawler$ViewCrawlerHandler.handleEventBindingsReceived(ViewCrawler.java:789)
       at com.mixpanel.android.viewcrawler.ViewCrawler$ViewCrawlerHandler.handleMessage(ViewCrawler.java:333)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:158)
```

1jpgodara:
Same here, After updating mix panel library to version 4.9.2 we are getting so many crashes with same stack trace.

abhi-practo:
@patedit when can we expect this released?

patedit:
@abhi-practo This fix was included in the latest release. Thanks!

PR #411, opened 2016-12-15
use concurrentmap for tweaks

patedit:
Should fix https://github.com/mixpanel/mixpanel-android/issues/408


1jpgodara:
When can we expect the release for this?

1jpgodara:
@patedit ^ We are getting lots of crashes because of this, Can you please update on the release for this?

PR #432, opened 2017-02-24
Fix NoSuchMethodError mini inapp crash

joeymalysz:
Fragment.getContext() is only supported at API level 23 - changing to getActivity() will resolve.

Issue not caught initially due to build tools error.


patedit:
Thanks @joeymalysz :)

Issue #424, opened 2017-02-02
BLUETOOTH permission is listed as optional in docs but seems to be required


agramian:
The [Android setup docs](https://mixpanel.com/help/reference/android#installing-as) under *Step 2 - Add permissions to your AndroidManifest.xml:* list the BLUETOOTH permission as optional but I am seeing a consistent crash for at least a Samsung SM-925T.  I have not tested enough to determine if this is device specific but either way it should not happen.  It causes our app to go into a crash/restart cycle of sometimes more than 10 times.

```
02-02 10:09:11.688 13384-13405/? E/BluetoothAdapter: Application does not have bluetooth permission, registering is failed
02-02 10:09:11.688 13384-13405/? E/BluetoothAdapter: java.lang.SecurityException: Need BLUETOOTH permission: Neither user 10348 nor current process has android.permission.BLUETOOTH.
                                                         at android.os.Parcel.readException(Parcel.java:1620)
                                                         at android.os.Parcel.readException(Parcel.java:1573)
                                                         at android.bluetooth.IBluetoothManager$Stub$Proxy.registerStateChangeCallback(IBluetoothManager.java:350)
                                                         at android.bluetooth.BluetoothAdapter.<init>(BluetoothAdapter.java:594)
                                                         at android.bluetooth.BluetoothAdapter.getDefaultAdapter(BluetoothAdapter.java:576)
                                                         at com.mixpanel.android.mpmetrics.SystemInformation.isBluetoothEnabled(SystemInformation.java:141)
                                                         at com.mixpanel.android.mpmetrics.AnalyticsMessages$Worker$AnalyticsMessageHandler.getDefaultEventProperties(AnalyticsMessages.java:559)
                                                         at com.mixpanel.android.mpmetrics.AnalyticsMessages$Worker$AnalyticsMessageHandler.prepareEventObject(AnalyticsMessages.java:573)
                                                         at com.mixpanel.android.mpmetrics.AnalyticsMessages$Worker$AnalyticsMessageHandler.handleMessage(AnalyticsMessages.java:251)
                                                         at android.os.Handler.dispatchMessage(Handler.java:102)
                                                         at android.os.Looper.loop(Looper.java:158)
                                                         at android.os.HandlerThread.run(HandlerThread.java:61)
```

patedit:
Hi @agramian ! Thanks for reporting this. What's the SDK Version that you are using? Also, have you modified the library somehow?

Thanks!

[...]

PR #435, opened 2017-03-02
Fix crash with lint

patedit:


edenman:
Was it intentional that this change makes the `BLUETOOTH` permission required?

patedit:
Hi @edenman! I gave it some thought and since BLUETOOTH permission is considered as [normal](https://developer.android.com/reference/android/Manifest.permission.html#BLUETOOTH), the system will grant access automatically and won't prompt the user. Also, this would fix this crash indirectly https://github.com/mixpanel/mixpanel-android/issues/424 - Please, let me know what you think and if you have a case where you don't want that permission to be merged on your manifest file.

edenman:
Hm.  I upgraded to mixpanel-android 4.9.7 and I get this warning in the Play store when I go to publish it:
screen shot 2017-03-02 at 2 28 49 pm

Maybe the warning doesn't take into account protection levels?

patedit:
Thanks for the info! That warning is pretty annoying and I'm sure that many other developers will have the same concern as you. You can read the second paragraph of [this link](https://developer.android.com/guide/topics/permissions/normal-permissions.html) and see that the system grants access automatically to normal permissions. 

In your case I think users won't have to accept the BLUETOOTH permission. If it's a real concern for you right now I have no problem in removing the permission and create a release. Let me know what you prefer.

edenman:
I think the warning is actually correct. https://support.google.com/googleplay/answer/6014972 seems to indicate that adding Bluetooth permission will keep auto-update from applying.

edenman:
(confirmed this with somebody at Google.  i'll just remove the permission in my manifest for now)

patedit:
Ah, interesting. I did not consider that. The best thing will be to remove it from the SDK as well. Thanks a lot for your useful feedback :)

edenman:
👍

smithaaron:
I also inexplicably received this warning when uploading a new build to the store. It was puzzling at first since I hadn't made any manifest or dependency changes. Eventually I realised it was due to mixpanel, as I had my dependency using a '+' as per your docs
`compile "com.mixpanel.android:mixpanel-android:4.+`

I now have 4.9.6 hardcoded as I don't need nor want to add the Bluetooth permission. Please apply a proper fix instead of forcing this on people.

ming13:
Patiently awaiting for a next release for now...

patedit:
@edenman @ming13 @smithaaron I removed this permission in the latest release 4.9.8. sorry for the inconvenience guys :)

PR #441, opened 2017-04-03
Add support for minimum and maximum values for Tweak variables

ryanseams:
Add support for minimum and maximum values for Tweak variables. In the Mixpanel A/B testing builder this restricts numerical Tweak values to be within a specified range: 
image

The syntax of a Tweak with min and max looks like:
`Tweak<Integer> testInt = MixpanelAPI.intTweak("Test Int", 2, 0, 10);`

This PR brings the Android library up to parity with iOS which allows for minimum and maximum values on Tweak variables.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
@@ -122,16 +122,46 @@ public class Tweaks {
      * and will likely not be directly useful to code that imports the Mixpanel library.
      */
     public static class TweakValue {
-        private TweakValue(@TweakType int aType, Object aDefaultValue, Number aMin, Number aMax, Object value) {
+        private TweakValue(@TweakType int aType, Object aDefaultValue, Number aMin, Number aMax, Object value, String aName) {
             type = aType;
-            defaultValue = aDefaultValue;
+            name = aName;
             minimum = aMin;
             maximum = aMax;
+            if(minimum != null && maximum != null) {
+                try {
+                    if (((Number) aDefaultValue).longValue() < minimum.longValue()) {
+                        MPLog.w(LOGTAG, "Attempt to define a tweak \"" + name + "\" with default value " + aDefaultValue + " which is below the minimum. " +
+                                "Tweak \"" + name + "\" will be set to the minimum value " + minimum + ".");
+                        aDefaultValue = minimum;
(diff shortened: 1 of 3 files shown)


patedit:
@ryanseams Thanks for the PR! Setting minimum and maximum values for Tweaks could be very useful and informative. I can see that we are now sending its minimum and maximum value as part of the json payload. However, we should also use those numbers to restrict a new updated tweak value probably here https://github.com/mixpanel/mixpanel-android/pull/441/files#diff-12156f10a71140dc852b0573774a2484R133

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -122,16 +122,46 @@ public synchronized boolean isNewValue(String tweakName, Object value) {
...
+                                "Tweak \"" + name + "\" will be set to the maximum value " + maximum);
+                        value = maximum;
+                    }
+                } catch(ClassCastException e) {
+                    ; // ok
+                }
+            }
+            defaultValue = aDefaultValue;
             this.value = value;
+
```
patedit: extra line

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -122,16 +122,46 @@ public synchronized boolean isNewValue(String tweakName, Object value) {
...
     public static class TweakValue {
-        private TweakValue(@TweakType int aType, Object aDefaultValue, Number aMin, Number aMax, Object value) {
+        private TweakValue(@TweakType int aType, Object aDefaultValue, Number aMin, Number aMax, Object value, String aName) {
             type = aType;
-            defaultValue = aDefaultValue;
+            name = aName;
             minimum = aMin;
             maximum = aMax;
+            if(minimum != null && maximum != null) {
+                try {
```
patedit: Some notes in this piece of code:
- You don't need to cast that to a `Number` object and use primitives instead (you won't need a try/catch then)
- This could be a utility in this class or somewhere. Given a value, a min and a max, return the appropriate number so it's not out of bounds.
- This logic seems the same as doing `Math.min(Math.max(value, min), max);`

patedit:
Thanks @ryanseams ! added some tests (https://github.com/mixpanel/mixpanel-android/commit/4cf892abbd2788f3574a80516c8f57be917779ff) and it works great :)

PR #443, opened 2017-04-07
update log output to be consistent with meta-data naming

gmale:
It took me a few minutes to realize why my attempts to change these meta-data properties were failing. After, I spotted the inconsistent naming, I figured I'd submit a PR for a low-risk fix that changes only the log output rather than the base variable names so that ALL the logged values can be appended to `com.mixpanel.android.MPConfig` consistently.


patedit:
Sorry for the confusion. Happy to merge this!

Issue #444, opened 2017-04-10
Data not sent to MixPanel in Kiosk environment


gmale:
We are using MixPanel on locked-down Kiosk devices in a retail store with custom Ethernet hardware. After a lot of investigation, we've found this information in the logs:

> 04-07 23:39:06.111	Flushing queue due to scheduled or forced flush (Thread 224)
04-07 23:39:06.113	ConnectivityManager says we are not online
04-07 23:39:06.114	Not flushing data to Mixpanel because the device is not connected to the internet. (Thread 224)

After inspecting the SDK source code, these checks happen in `HttpService` but there is a tiny bug that doesn't account for when `NetworkInfo` can be null. According to the documentation for [ConnectivityManager#getActiveNetworkInfo](https://developer.android.com/reference/android/net/ConnectivityManager.html#getActiveNetworkInfo()):

>This may return null when there is no default network.

Accordingly, [we're submitting the following patch, which corrected the in-store problem for us](https://github.com/mixpanel/mixpanel-android/pull/445). After weeks of investigating every networking layer and IT restriction known to man, it turned out this one line of code was the culprit. Now, MixPanel works in the store–albeit with a patched SDK.

patedit:
Thanks again @gmale  👍

PR #445, opened 2017-04-10
bugfix: netInfo can be null even when there is connectivity.

gmale:
This bug resulted in MixPanel requests never being sent in our locked-down Ethernet Kiosk environment, as referenced in Issue #444 

From our investigation, it is worth noting that:

- We have connectivity via Ethernet
- We have the `android.permission.ACCESS_NETWORK_STATE` permission enabled
- `connectivityManager.getActiveNetworkInfo()` always returns `null`
- `connectivityManager.getAllNetworks()` always returns either `null` or an empty array (on API 22)

Also, according to the documentation for [ConnectivityManager#getActiveNetworkInfo](https://developer.android.com/reference/android/net/ConnectivityManager.html#getActiveNetworkInfo()):

>This may return null when there is no default network.

The code in `HttpService` was assuming that a null `NetworkInfo` response signifies no connectivity but that is not always the case. In our Ethernet environment, we have connectivity but there is no default network.


patedit:
@gmale This is great. Thanks for the investigation and for implementing a fix. Will merge now and release it later today.

PR #451, opened 2017-04-24
Update ConfigurationChecker.java

argenisf:
Returning the value for mTakeoverActivityAvailable after it has been deemed false (otherwise it is always deemed true on line 203)


patedit:
thanks @argenisf !

PR #460, opened 2017-05-12
Pre track events

patedit:


src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
```diff
@@ -179,13 +243,19 @@ public int addJSON(JSONObject j, Table table) {
...
 
         try {
             final SQLiteDatabase db = mDb.getWritableDatabase();
-            db.delete(tableName, "_id <= " + last_id, null);
+            StringBuffer deleteQuery = new StringBuffer("_id <= " + last_id + " AND " + KEY_TOKEN + " = '" + token + "'");
+
+            if (!includeAutomaticEvents) {
+                deleteQuery.append(" AND " + KEY_AUTOMATIC_DATA + "=0");
+            }
+            db.delete(tableName, deleteQuery.toString(), null);
```
yarneo: hey, not totally competent in this part of the code so have a question! I noticed that if includeAutomaticEvents is false (which could be the case if decide is null or automaticEventsEnabled is null) then we delete the automatic events. Are we still saving the events of automaticevents that we collect before we get a decide response (as an example, first app open usually happens before we have a decide response). Just making sure!

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -238,8 +239,20 @@
             decideId = mPersistentIdentity.getEventsDistinctId();
         }
         mDecideMessages.setDistinctId(decideId);
+
         mMessages = getAnalyticsMessages();
 
+        if (mPersistentIdentity.isFirstLaunch(MPDbAdapter.getInstance(mContext).getDatabaseFile().exists())) {
```
yarneo: when automaticevents is toggled to off (in the UI) , do you have any concerns in which we might want to not even execute the automaticevents code? I have it that the code isn't even executed, but initially it was for safety, so i should remove that safety if so.

src/androidTest/java/com/mixpanel/android/mpmetrics/TestUtils.java
```diff
@@ -30,7 +30,15 @@ public CleanMixpanelAPI(final Context context, final Future<SharedPreferences> r
...
+
+            final String timeEventsPrefsName = "com.mixpanel.android.mpmetrics.MixpanelAPI.TimeEvents_" + token;
+            final SharedPreferences timeSharedPrefs = context.getSharedPreferences(timeEventsPrefsName, Context.MODE_PRIVATE);
+            timeSharedPrefs.edit().clear().commit();
+
+            final String mixpanelPrefsName = "com.mixpanel.android.mpmetrics.Mixpanel";
+            final SharedPreferences mpSharedPrefs = context.getSharedPreferences(mixpanelPrefsName, Context.MODE_PRIVATE);
+            mpSharedPrefs.edit().clear().putBoolean(token, true).putBoolean("has_launched", true).commit();
 
             return super.getPersistentIdentity(context, referrerPreferences, token);
```
yarneo: i couldn't find any specific unit tests for automaticEvents, do you feel good about the overall coverage?

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -233,31 +281,35 @@ public void handleMessage(Message msg) {
...
+                            returnCode = mDbAdapter.addJSON(message, token, MPDbAdapter.Table.EVENTS, eventDescription.isAutomatic());
                         } catch (final JSONException e) {
                             MPLog.e(LOGTAG, "Exception tracking event " + eventDescription.getEventName(), e);
                         }
                     } else if (msg.what == FLUSH_QUEUE) {
                         logAboutMessageToMixpanel("Flushing queue due to scheduled or forced flush");
                         updateFlushFrequency();
-                        sendAllData(mDbAdapter);
-                        if (SystemClock.elapsedRealtime() >= mDecideRetryAfter) {
+                        token = (String) msg.obj;
```
patedit: that's a good point. the problem is that we use methods like `hasMessages` and `removeMessage` in other parts of the code so we'd need to keep a reference to the object that we pass on this message. Since the token (an string) is passed on every message (PeopleDescription, EventDescription, etc..) it makes it easier to keep the string as the object so we can remove those messages in the queue.

src/main/java/com/mixpanel/android/mpmetrics/AutomaticEvents.java
```diff
@@ -0,0 +1,14 @@
...
+    public static final String FIRST_OPEN = "$ae_first_open";
+    public static final String FIRST_OPEN_DATE = "$ae_first_app_open_date";
+    public static final String SESSION = "$ae_session";
+    public static final String SESSION_LENGTH = "$ae_session_length";
+    public static final String TOTAL_SESSIONS = "$ae_total_app_sessions";
+    public static final String TOTAL_SESSIONS_LENGTH = "$ae_total_app_session_length";
+    public static final String APP_UPDATED = "$ae_updated";
+    public static final String VERSION_UPDATED = "$ae_updated_version";
+    public static final String APP_CRASHED = "$ae_crashed";
+    public static final String APP_CRASHED_REASON = "$ae_crashed_reason";
```
patedit: i can see why it's useful in js but can't see the use case in java since you can follow references. i prefer not to bring this methodology in the sdk :)

src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java
patedit: if `includeAutomaticEvents` is false then we don't delete the automatic events! (`=0`)!

yarneo: awesome!

PR #465, opened 2017-05-31
add a method to check elapsed time for a timed event. this allows use…

pchien:
…rs to choose different behavior based on the elapsed time, e.g. if they want a minimum duration they can choose to not track

@patedit i couldn't figure out if there was a demo app or how to run one so i didn't actually test it...

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -461,6 +461,20 @@ public class MixpanelAPI {
     }
 
     /**
+     * Retrieves the time elapsed for the named event since timeEvent() was called.
+     *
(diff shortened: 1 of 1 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -461,6 +461,20 @@ public void timeEvent(final String eventName) {
...
     /**
+     * Retrieves the time elapsed for the named event since timeEvent() was called.
+     *
+     * @param eventName the name of the event to be tracked that was previously called with timeEvent()
+     */
+    public double eventElapsedTime(final String eventName) {
+        final long currentTime = System.currentTimeMillis();
+        long startTime;
+        synchronized (mEventTimings) {
+            startTime = mEventTimings.get(eventName);
```
patedit: We probably want to check if there is a timed event for the given name and return 0 in that case 👍

PR #469, opened 2017-06-21
Persist seen in-app campaign ids accross sessions

patedit:
So in case there is no connectivity and the user opens up the app again later, the SDK will make sure we don't show the same in-app twice (i.e. `decide` won't know that the notification was seen until we flush again / connectivity is restored).

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
@@ -21,7 +21,7 @@ import java.util.Set;
         void onNewResults();
     }
 
-    public DecideMessages(Context context, String token, OnNewResultsListener listener, UpdatesFromMixpanel updatesFromMixpanel) {
+    public DecideMessages(Context context, String token, OnNewResultsListener listener, UpdatesFromMixpanel updatesFromMixpanel, HashSet<Integer> notificationIds) {
         mContext = context;
         mToken = token;
         mListener = listener;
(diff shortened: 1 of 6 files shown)


src/main/java/com/mixpanel/android/mpmetrics/PersistentIdentity.java
```diff
@@ -430,6 +432,37 @@ public synchronized void setHasLaunched() {
...
+        }
+        return campaignIds;
+    }
+
+    public synchronized void saveCampaignAsSeen(Integer notificationId) {
+        try {
+            final SharedPreferences prefs = mLoadStoredPreferences.get();
+            final SharedPreferences.Editor editor = prefs.edit();
+            String campaignIds = prefs.getString("seen_campaign_ids", "");
+            editor.putString("seen_campaign_ids", campaignIds + notificationId + DELIMITER);
```
pchien: lame there's no default serialization/deserialization

src/main/java/com/mixpanel/android/mpmetrics/DecideMessages.java
```diff
@@ -21,15 +21,15 @@
...
+    public DecideMessages(Context context, String token, OnNewResultsListener listener, UpdatesFromMixpanel updatesFromMixpanel, HashSet<Integer> notificationIds) {
         mContext = context;
         mToken = token;
         mListener = listener;
         mUpdatesFromMixpanel = updatesFromMixpanel;
 
         mDistinctId = null;
         mUnseenNotifications = new LinkedList<InAppNotification>();
-        mNotificationIds = new HashSet<Integer>();
+        mNotificationIds = new HashSet<Integer>(notificationIds);
```
pchien: is there a reason to make a copy here?

patedit: for safety mainly

Issue #473, opened 2017-07-14
Icons are never loaded in mipmap folder


lazybee27102:
I use `ic_launcher` for notification icon which was stored in `mipmap` folder. But it never loaded in the status bar. When I try to read your codes, then I see you never load it from mipmap folder.

This is the code of loading `drawable` resource 
```
public static class Drawables extends ResourceReader {
        protected Drawables(String resourcePackageName, Context context) {
            super(context);
            mResourcePackageName = resourcePackageName;
            initialize();
        }

        @Override
        protected Class<?> getSystemClass() {
            return android.R.drawable.class;
        }

        @Override
        protected String getLocalClassName(Context context) {
            return mResourcePackageName + ".R$drawable";
        }

        private final String mResourcePackageName;
    }
```

How about creating new class `Mipmaps` for `.R$mipmap` when resource in drawable isn't available? 

PS: I think this use-case is common, people want to see their `ic_launcher` as the notification icon. The same as `smallIcon` and `whiteIcon`.

Please take a look and give me your response. Thanks

patedit:
Hi @lazybee27102 ! Thanks for the time you spent researching this issue. It totally makes sense and I love your idea. Thanks for sharing! Would it be possible for you to create a PR for this feature? Otherwise, I can take care of it in the coming weeks.

lazybee27102:
Hi @patedit, I just create a pull request. It would be nice if you could take a look. 
Thank you

patedit:
Thanks @lazybee27102 - will close this ticket on favor of the pr

PR #474, opened 2017-07-25
[Func] Add mipmap as new ResourceReader

phatnhse:
## This is the pull request for issue #473 
First I want to implement if/else condition for a class named `Drawables`. But there is no method that can help me to detect if that resource really exists. 

Then I decide to create new class name `Mipmaps` that can return the the class name of `android.R.mipmap.class` if `SDK_INT >= HONEYCOMB` (Or return `Drawable` as well)

```
public static class Mipmaps extends ResourceReader {
        protected Mipmaps(String resourcePackageName, Context context) {
            super(context);
            mResourcePackageName = resourcePackageName;
            initialize();
        }

        @Override
        protected Class<?> getSystemClass() {
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
                return android.R.mipmap.class;
            }

            return android.R.drawable.class;
        }

        @Override
        protected String getLocalClassName(Context context) {
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
                return mResourcePackageName + ".R$mipmap";
            }

            return mResourcePackageName + ".R$drawable";
        }

        private final String mResourcePackageName;
    }
```

This could be better if we can combine `Drawables` and `Mipmaps` as one class. But I think It's still ok, at least it works fine for your codebase.

Hope you take a look. It would be nice if I could get your response 

P/S: Hey, can you help me to write test for this functionality?

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
@@ -2,9 +2,6 @@ package com.mixpanel.android.mpmetrics;
 
 import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
-import android.graphics.BitmapFactory;
-import android.graphics.Color;
-import android.support.v4.app.NotificationCompat;
 import android.app.Notification;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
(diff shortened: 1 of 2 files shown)


src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
```diff
@@ -236,8 +243,8 @@ private Intent buildNotificationIntent(Context context, String uriString, String
         return ret;
     }
 
-    private Notification buildNotification(Context context, Intent inboundIntent, ResourceIds iconIds) {
-        final NotificationData notificationData = readInboundIntent(context, inboundIntent, iconIds);
+    private Notification buildNotification(Context context, Intent inboundIntent, ResourceIds mipmapIds, ResourceIds iconIds) {
```
patedit: could we have a `List<ResourceIds>` instead?

zihejia:
On Jan 1, 2022, we’ll remove the Messages & Experiments feature from Mixpanel([more details](https://mixpanel.com/blog/why-were-sunsetting-messaging-and-experiments/)). 

This is housekeeping to close PR that is no longer relevant. Thanks @phatnhse for providing it.

PR #479, opened 2017-08-08
Use GAID as Distinct ID for new users

ryanseams:
Use GAID as Distinct ID for new users, also adding this as a super property and a people property. This PR uses delay to wait for the asynch GAID to return, and only then starts to process events and people updates requested on the main thread. Although this might not be a final solution due to the asycnh issues with getting the GAID, hopefully it can serve as inspiration for one.


zihejia:
Thanks for providing the PR. We will not use `GAID` on the SDK unfortunately.

PR #481, opened 2017-08-21
properties for $campaign_open for the cta url and if the button was p…

pchien:
…rimary or secondary. also now always track $campaign_open if they tap on either button or mini, instead of only if there was a ctal url

this is currently untested. i don't know how to use gradle to point the test app at my local changes. will ask in person later this week

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/InAppFragment.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/InAppFragment.java
@@ -35,6 +35,9 @@ import com.mixpanel.android.R;
 import com.mixpanel.android.util.MPLog;
 import com.mixpanel.android.util.ViewUtils;
 
+import org.json.JSONException;
+import org.json.JSONObject;
+
 /**
  * Attached to an Activity when you display a mini in-app notification.
  *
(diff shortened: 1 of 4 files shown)


pchien:
okay i tested now. it seems to work!

src/main/java/com/mixpanel/android/mpmetrics/InAppFragment.java
```diff
@@ -148,11 +152,18 @@ public boolean onSingleTapUp(MotionEvent event) {
...
                     }
+
+                    try {
+                        trackingProperties = new JSONObject();
+                        trackingProperties.put("url", uriString);
+                    } catch (final JSONException e) {
+                        MPLog.e(LOGTAG, "Can't put url into json properties");
+                    }
                 }
+                mMixpanel.getPeople().trackNotification("$campaign_open", inApp, trackingProperties);
```
patedit: do we still want to track campaign open even if the mini in-app does nothing?

src/main/java/com/mixpanel/android/takeoverinapp/TakeoverInAppActivity.java
```diff
@@ -187,11 +191,31 @@ public void onClick(View v) {
...
+                    }
+                    try {
+                        if (trackingProperties == null) {
+                            trackingProperties = new JSONObject();
+                        }
+                        trackingProperties.put("button", whichButton);
+                    } catch (final JSONException e) {
+                        MPLog.e(LOGTAG, "Can't put button type into json properties");
                     }
+                    mMixpanel.getPeople().trackNotification("$campaign_open", inApp, trackingProperties);
```
patedit: same as before

src/main/java/com/mixpanel/android/mpmetrics/InAppFragment.java
pchien: i think so. it tracks taps, same as normal inapps

PR #495, opened 2017-09-20
Android O support

patedit:
- Add notification channel for android O notifications
- New emulator support
- Show in-app on every new activity
- Travis CI

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
@@ -2,6 +2,7 @@ package com.mixpanel.android.mpmetrics;
 
 import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
+import android.app.NotificationChannel;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
 import android.support.v4.app.NotificationCompat;
@@ -251,7 +252,9 @@ public class GCMReceiver extends BroadcastReceiver {
         );
 
         final Notification notification;
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            notification = makeNotificationSDK26OrHigher(context, contentIntent, notificationData);
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
             notification = makeNotificationSDK21OrHigher(context, contentIntent, notificationData);
         } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
             notification = makeNotificationSDK16OrHigher(context, contentIntent, notificationData);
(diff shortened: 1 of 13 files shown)


src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java
```diff
@@ -399,6 +402,56 @@ protected Notification makeNotificationSDK21OrHigher(Context context, PendingInt
         return n;
     }
 
+    @SuppressLint("NewApi")
+    @TargetApi(26)
+    protected Notification makeNotificationSDK26OrHigher(Context context, PendingIntent intent, NotificationData notificationData) {
+        NotificationManager mNotificationManager =
+                (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+        String channelId = "mp";
```
fred8: I think it will be better if the host app can provide the channel id, name, importance

patedit: Thanks for your feedback!

darwind:
Hi - I know this is an old issue and it's closed, but I seem to have hit it now for some reason.

When targeting anything higher than SDK 25, I don't receive any notifications. If I change my `targetSdk` to 25 I'm receiving the notifications.

Looking at the errors that I get when the notification is not shown it all boils down to an issue with a missing notification channel. 🤔 

I found out by searching on Google for a log message that was printed when trying to send notifications to our app and I stumbled upon this SO answer: https://stackoverflow.com/a/48513181/882251

The message I'm seeing is similar to the one OP describes in his question:

```
NotificationService: No Channel found for pkg=com.google.android.apps.messaging, channelId=miscellaneous, id=5, tag=null, opPkg=com.google.android.apps.messaging, callingUid=10130, userId=0, incomingUserId=0, notificationUid=10130, notification=Notification(channel=miscellaneous pri=0 contentView=null vibrate=null sound=null defaults=0x0 flags=0x8 color=0xff2a56c6 vis=PRIVATE)
```

Just with our packageId instead of `com.google.android.apps.messaging`.

Is this something you are aware of?

I also find it quite weird, that you haven't updated your notification receiver to use Firebase directly - you're using a normal `BroadcastReceiver` which means we have to manually specify and request the `WAKELOCK` permission which shouldn't be necessary using the updated Firebase setup.

You're also requesting another unnecessary permision `GET_ACCOUNTS` which I don't really see how we can include and have the user accept when it comes to push notifications?

I'm a bit confused by your implementation I have to say...

I'm looking at your sample app: https://github.com/mixpanel/sample-android-mixpanel-integration
and the instructions here: https://mixpanel.com/help/reference/android-push-notifications
which is describing out-dated procedures it seems.

Am I missing something here? 🤔

argenisf:
Hey @darwind could you check your version of the Android SDK? Support for Oreo and channels was added back [in version 5.2.0](https://github.com/mixpanel/mixpanel-android/releases/tag/v5.2.0) so perhaps the version you are using is older? If you look at the default receiver on the current version of the library, you [will see support there](https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/GCMReceiver.java#L406).

We have not updated to use Firebase just yet, but it's definitely in the roadmap. Give updating a try, and if it does not work, make sure to write into support@mixpanel.com with more details and we'll help you get set up.

Issue #527, opened 2018-01-22
Switch endpoint at runtime


drauly:
Hello,

The library allows us to change mixpanel url in configuration. We would like to be able to do it at runtime (for AB testing purpose).

The idea is to be able to do :

```java
MixpanelAPI mixpanel = MixpanelAPI.getInstance(this, projectToken, "http://my.mixpanel.proxy");
```

Would you accept any PR to achieve that ? 

Thank you

drauly:
Hello @patedit !

PR submitted

patedit:
Issue resolved https://github.com/mixpanel/mixpanel-android/pull/530

PR #529, opened 2018-01-23
switch url at runtime

drauly:
Should resolve issue #527 

This allow to change url at runtime:

```java
MixpanelAPI mixpanel = MixpanelAPI.getInstance(this, projectToken, "http://my.mixpanel.proxy");
```
Provided url is optional

### DIFF
--- a/src/androidTest/java/com/mixpanel/android/mpmetrics/AutomaticEventsTest.java
+++ b/src/androidTest/java/com/mixpanel/android/mpmetrics/AutomaticEventsTest.java
@@ -91,11 +91,11 @@ public class AutomaticEventsTest extends AndroidTestCase {
             }
 
             @Override
-            public int addJSON(JSONObject j, String token, Table table, boolean isAutomaticRecord) {
+            public int addJSON(JSONObject j, String token, String endpoint, Table table, boolean isAutomaticRecord) {
                 if (token.equalsIgnoreCase(TOKEN)) {
                     mTrackedEvents++;
                     mLatch.countDown();
-                    return super.addJSON(j, token, table, isAutomaticRecord);
+                    return super.addJSON(j, token, endpoint, table, isAutomaticRecord);
                 }
 
                 return 1;
@@ -252,10 +252,10 @@ public class AutomaticEventsTest extends AndroidTestCase {
             }
 
             @Override
-            public int addJSON(JSONObject j, String token, Table table, boolean isAutomaticRecord) {
+            public int addJSON(JSONObject j, String token, String endpoint, Table table, boolean isAutomaticRecord) {
                 if (token.equalsIgnoreCase(SECOND_TOKEN)) {
                     secondLatch.countDown();
-                    return super.addJSON(j, token, table, isAutomaticRecord);
+                    return super.addJSON(j, token, endpoint, table, isAutomaticRecord);
                 }
 
                 return 1;
--- a/src/androidTest/java/com/mixpanel/android/mpmetrics/AutomaticEventsTest.java
+++ b/src/androidTest/java/com/mixpanel/android/mpmetrics/AutomaticEventsTest.java
@@ -143,12 +143,12 @@ public class AutomaticEventsTest extends AndroidTestCase {
         mCleanMixpanelAPI = new MixpanelAPI(getContext(), mMockReferrerPreferences, TOKEN) {
 
             @Override
-        /* package */ PersistentIdentity getPersistentIdentity(final Context context, final Future<SharedPreferences> referrerPreferences, final String token) {
-                final String prefsName = "com.mixpanel.android.mpmetrics.MixpanelAPI_" + token;
+        /* package */ PersistentIdentity getPersistentIdentity(final Context context, final Future<SharedPreferences> referrerPreferences, final String token, final String endpoint) {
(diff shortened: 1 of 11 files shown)


patedit:
Thanks a lot for the effort @drauly! Sadly, I'm afraid that this PR can't be merged as it's right now. Let's work together and figure out a way to get this out 👍 

Local storage is coupled with the token string and this creates a problem with the proposed solution. Some examples: 
- We create two ([1](https://github.com/mixpanel/mixpanel-android/blob/v5.2.4/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java#L1362), [2](https://github.com/mixpanel/mixpanel-android/blob/v5.2.4/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java#L1365)) shared preferences files based on the token. If we create two instances with the same token, those preferences will be shared (e.g. `instanceA.timeEvent("An Event")` followed by `instanceB.track("An Event")` will send a timed event through `instanceB` and `instanceA` will never receive that event!). There are many more use cases.
- [The events data](https://github.com/mixpanel/mixpanel-android/blob/v5.2.4/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java#L455) that we send is also based on [the token](https://github.com/mixpanel/mixpanel-android/blob/v5.2.4/src/main/java/com/mixpanel/android/mpmetrics/MPDbAdapter.java#L350). So if `instanceA` flushes its queue first, `instanceB` (using the same token) won't have anything to send.

Let me better understand your use case and find a good solution: Are you planning on sending all the events to both your proxy server and https://api.mixpanel.com/track or just a subset of them? If it's the former, I think an elegant solution is to have an array of endpoints in `MPConfig` (https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java#L355) and be able to add/remove urls. That way, [this method](https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java#L448) would accept an array of strings as the last param. We used to do something similar before (see [this](https://github.com/mixpanel/mixpanel-android/blob/v5.1.1/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java#L450) and [this](https://github.com/mixpanel/mixpanel-android/blob/v5.1.1/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java#L476)).

If it's the latter, things will get more complicated and will need to think about something else. Thanks for your PR again!

drauly:
Thanks for your feedbacks.

My use case is to only use 1 instance at a time. 
All events got by instanceX goes to endpointX

Sorry I missed the sharedPreferenced and local storage issue.
If I understand well, we only have to index those not only by token, but by token+endpoint (as for the mxpApi instances).

I'll do a commit soon to achieve that. Tell me if this solution sounds good to you, and if I missed something else

drauly:
Commits pushed.

Note that I didn't change tables structure for backward compatibility (changing it would involves data migration)

let me know your feedback,

Thanks !

patedit:
Hi @drauly ! Thanks for clarifying your use case. In that case, what do you think about this PR https://github.com/mixpanel/mixpanel-android/pull/530?

I'd rather have customizable params only in `MPConfig` 👍 Let me know what you think!

drauly:
Hello @patedit ,

Thanks for the PR. I would be a much simpler solution indeed. But even if I change the url through MPConfig, i can not get a new instance of mxpApi. 

instanceA = MixpanelAPI.getInstance(this, token)
MPConfig.setEndpoint(...)
instanceB = MixpanelAPI.getInstance(this, token)
// here instanceB == instanceA => no url change

I'm aware that my PR involve many modification for a simple usecase. If you have simpler solution let me know.
The problem is that we really need it :(

Waiting for your feedback,

Thanks again

patedit:
Sorry, I miss understood your comment: "My use case is to only use 1 instance at a time.".

You need to have two different instances created at the same time and send some of your tracking calls to your proxy and some of your tracking calls to Mixpanel?

drauly:
Here is an example :
App ask to AB testing server : "should I track on proxy or on direct mxp url ?" 
AB respond "direct"
I build instanceA and use it (mxpApiSingleton = instanceA)
Later I ask again the same question to AB server and it says "proxy"
I then build instanceB and replace my instance (mxpApiSingleton = instanceB)
And so on ...

patedit:
Do you want to keep different super properties and/or people properties in each instance?
I might be missing something, but if the only difference is the url, why wouldn't something like this work?

```
if (shouldTrackProxy()) {
     mxpApiSingleton.flush(); // send all the tracked events so the go the the right upstream server
     // wait for flush to finish
     MPConfig.getInstance(this).setEventsEndpoint("https://myapp.company.com/proxy/")
} else {
     mxpApiSingleton.flush();
     // wait for flush to finish
     MPConfig.getInstance(this).setEventsEndpoint("https://api.mixpanel.com/track?ip=1")
}
```

drauly:
Yes, only url should change.
Your solution seems ok and much simpler, I was missing something :) 
I will test it soon and give you a feedback as soon as possible.
Thanks

patedit:
Fixed on https://github.com/mixpanel/mixpanel-android/pull/530

PR #532, opened 2018-02-01
Prevent NullPointerException crash

tomasz-m:
The crash is happening "randomly" (at least for us now) and asynchronously so we cannot try-catch it:
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.Object.equals(java.lang.Object)' on a null object reference
       at com.mixpanel.android.mpmetrics.Tweaks.isNewValue(Tweaks.java:63)
       at com.mixpanel.android.viewcrawler.ViewCrawler$ViewCrawlerHandler.applyVariantsAndEventBindings(ViewCrawler.java:931)
       at com.mixpanel.android.viewcrawler.ViewCrawler$ViewCrawlerHandler.handleVariantsReceived(ViewCrawler.java:791)
       at com.mixpanel.android.viewcrawler.ViewCrawler$ViewCrawlerHandler.handleMessage(ViewCrawler.java:348)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:154)
       at android.os.HandlerThread.run(HandlerThread.java:61)

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
@@ -58,10 +58,9 @@ public class Tweaks {
             MPLog.w(LOGTAG, "Attempt to reference a tweak \"" + tweakName + "\" which has never been defined.");
             return false;
         }
-        if(container == null || container.value == null)
-            return value == null;
-        
         final TweakValue container = mTweakValues.get(tweakName);
+        if (container == null || container.value == null)
+            return value == null;
         return !container.value.equals(value);
     }
 
--- a/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
@@ -58,7 +58,9 @@ public class Tweaks {
             MPLog.w(LOGTAG, "Attempt to reference a tweak \"" + tweakName + "\" which has never been defined.");
             return false;
         }
-
+        if(container == null || container.value == null)
+            return value == null;
+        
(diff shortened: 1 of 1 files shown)


src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -58,8 +58,9 @@ public synchronized boolean isNewValue(String tweakName, Object value) {
             MPLog.w(LOGTAG, "Attempt to reference a tweak \"" + tweakName + "\" which has never been defined.");
             return false;
         }
-
         final TweakValue container = mTweakValues.get(tweakName);
+        if (container == null || container.value == null)
```
patedit: We check for `containsKey()` in line 57 so `container == null` is not needed here.
Also, we avoid one-liner if statements without `{}`

src/main/java/com/mixpanel/android/mpmetrics/Tweaks.java
```diff
@@ -58,8 +58,9 @@ public synchronized boolean isNewValue(String tweakName, Object value) {
...
+            return value == null;
```
patedit: this means that we will consider `value` as a new value when it's `null` and the container's value is also `null`.
can we change this line to `value != null`?

zihejia:
On Jan 1, 2022, we’ll remove the Messages & Experiments feature from Mixpanel([more details](https://mixpanel.com/blog/why-were-sunsetting-messaging-and-experiments/)). 

This is housekeeping to close PR that is no longer relevant. Thanks @tomasz-m for providing it.

PR #537, opened 2018-03-03
Allow for longer database expiration time

ryanseams:
This PR adds support for longer database expiration by using a `float` as opposed to an `int`. Previously the max database expiration was limited to the max size of an `int` in Java.


patedit:
i'm afraid this will throw an exception if you use a number below 2^32

patedit:
Created a PR that will address this issue: https://github.com/mixpanel/mixpanel-android/pull/538/files

Thanks @ryanseams !!

PR #540, opened 2018-03-27
GDPR support

patedit:
Allow devs to opt-in / opt-out an end user from tracking.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -268,8 +268,6 @@ public class MixpanelAPI {
 
         if (optOutTrackingDefault) {
             optOutTracking();
-        } else {
-            mPersistentIdentity.setOptOutTracking(false);
         }
 
         mUpdatesListener = constructUpdatesListener();
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -228,15 +228,15 @@ public class MixpanelAPI {
      * You shouldn't instantiate MixpanelAPI objects directly.
      * Use MixpanelAPI.getInstance to get an instance.
      */
-    MixpanelAPI(Context context, Future<SharedPreferences> referrerPreferences, String token) {
-        this(context, referrerPreferences, token, MPConfig.getInstance(context));
+    MixpanelAPI(Context context, Future<SharedPreferences> referrerPreferences, String token, boolean optOutTrackingDefault) {
+        this(context, referrerPreferences, token, MPConfig.getInstance(context), optOutTrackingDefault);
     }
 
     /**
      * You shouldn't instantiate MixpanelAPI objects directly.
      * Use MixpanelAPI.getInstance to get an instance.
      */
-    MixpanelAPI(Context context, Future<SharedPreferences> referrerPreferences, String token, MPConfig config) {
+    MixpanelAPI(Context context, Future<SharedPreferences> referrerPreferences, String token, MPConfig config, boolean optOutTrackingDefault) {
         mContext = context;
         mToken = token;
         mPeople = new PeopleImpl();
@@ -264,6 +264,14 @@ public class MixpanelAPI {
         mTrackingDebug = constructTrackingDebug();
         mPersistentIdentity = getPersistentIdentity(context, referrerPreferences, token);
         mEventTimings = mPersistentIdentity.getTimeEvents();
+        mMessages = getAnalyticsMessages();
+
+        if (optOutTrackingDefault) {
+            optOutTracking();
+        } else {
+            mPersistentIdentity.setOptOutTracking(false);
+        }
+
         mUpdatesListener = constructUpdatesListener();
         mDecideMessages = constructDecideUpdates(token, mUpdatesListener, mUpdatesFromMixpanel);
         mConnectIntegrations = new ConnectIntegrations(this);
@@ -275,8 +283,6 @@ public class MixpanelAPI {
             decideId = mPersistentIdentity.getEventsDistinctId();
         }
         mDecideMessages.setDistinctId(decideId);
-
-        mMessages = getAnalyticsMessages();
 
         if (mPersistentIdentity.isFirstLaunch(MPDbAdapter.getInstance(mContext).getDatabaseFile().exists())) {
             track(AutomaticEvents.FIRST_OPEN, null, true);
@@ -301,6 +307,7 @@ public class MixpanelAPI {
                 messageProps.put("mp_lib", "Android");
                 messageProps.put("lib", "Android");
                 messageProps.put("distinct_id", token);
+                messageProps.put("$lib_version", MPConfig.VERSION);
 
                 final AnalyticsMessages.EventDescription eventDescription =
                         new AnalyticsMessages.EventDescription(
@@ -356,6 +363,38 @@ public class MixpanelAPI {
      * @return an instance of MixpanelAPI associated with your project
      */
     public static MixpanelAPI getInstance(Context context, String token) {
+        return getInstance(context, token, false);
+    }
+
+    /**
+     * Get the instance of MixpanelAPI associated with your Mixpanel project token.
+     *
+     * <p>Use getInstance to get a reference to a shared
+     * instance of MixpanelAPI you can use to send events
+     * and People Analytics updates to Mixpanel.</p>
+     * <p>getInstance is thread safe, but the returned instance is not,
+     * and may be shared with other callers of getInstance.
+     * The best practice is to call getInstance, and use the returned MixpanelAPI,
+     * object from a single thread (probably the main UI thread of your application).</p>
+     * <p>If you do choose to track events from multiple threads in your application,
+     * you should synchronize your calls on the instance itself, like so:</p>
+     * <pre>
+     * {@code
+     * MixpanelAPI instance = MixpanelAPI.getInstance(context, token);
+     * synchronized(instance) { // Only necessary if the instance will be used in multiple threads.
+     *     instance.track(...)
+     * }
+     * }
+     * </pre>
+     *
+     * @param context The application context you are tracking
+     * @param token Your Mixpanel project token. You can get your project token on the Mixpanel web site,
+     *     in the settings dialog.
+     * @param optOutTrackingDefault Whether or not Mixpanel can start tracking by default. See
+     *     {@link #optOutTracking()}.
+     * @return an instance of MixpanelAPI associated with your project
+     */
+    public static MixpanelAPI getInstance(Context context, String token, boolean optOutTrackingDefault) {
         if (null == token || null == context) {
             return null;
         }
@@ -374,7 +413,7 @@ public class MixpanelAPI {
 
             MixpanelAPI instance = instances.get(appContext);
             if (null == instance && ConfigurationChecker.checkBasicConfiguration(appContext)) {
-                instance = new MixpanelAPI(appContext, sReferrerPrefs, token);
+                instance = new MixpanelAPI(appContext, sReferrerPrefs, token, optOutTrackingDefault);
                 registerAppLinksListeners(context, instance);
                 instances.put(appContext, instance);
             }
@@ -413,6 +452,7 @@ public class MixpanelAPI {
      * @param original the old distinct_id that alias will be mapped to.
      */
     public void alias(String alias, String original) {
+        if (hasOptedOutTracking()) return;
         if (original == null) {
             original = getDistinctId();
         }
@@ -458,6 +498,7 @@ public class MixpanelAPI {
      * @see People#identify(String)
      */
     public void identify(String distinctId) {
+        if (hasOptedOutTracking()) return;
         synchronized (mPersistentIdentity) {
             mPersistentIdentity.setEventsDistinctId(distinctId);
             String decideId = mPersistentIdentity.getPeopleDistinctId();
@@ -476,6 +517,7 @@ public class MixpanelAPI {
      * @param eventName the name of the event to track with timing.
      */
     public void timeEvent(final String eventName) {
+        if (hasOptedOutTracking()) return;
         final long writeTime = System.currentTimeMillis();
         synchronized (mEventTimings) {
             mEventTimings.put(eventName, writeTime);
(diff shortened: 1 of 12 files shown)


jbwyme:
you just love to see this

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -263,6 +263,14 @@
         mTrackingDebug = constructTrackingDebug();
         mPersistentIdentity = getPersistentIdentity(context, referrerPreferences, token);
         mEventTimings = mPersistentIdentity.getTimeEvents();
+        mMessages = getAnalyticsMessages();
+
+        if (optOutTrackingDefault) {
+            optOutTracking();
+        } else {
+            mPersistentIdentity.setOptOutTracking(false);
```
zihejia: If the app restart, will that overwrite the previous optOut setting?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -300,6 +306,7 @@
                 messageProps.put("mp_lib", "Android");
                 messageProps.put("lib", "Android");
                 messageProps.put("distinct_id", token);
+                messageProps.put("$lib_version", MPConfig.VERSION);
```
yinfeiru: are we doing the same with iOS?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -352,6 +359,38 @@
      * @return an instance of MixpanelAPI associated with your project
      */
     public static MixpanelAPI getInstance(Context context, String token) {
+        return getInstance(context, token, false);
```
yinfeiru: have we made corresponding documentation changes?

yinfeiru: if the end user opted out, shutdown the app and restart. now when `getInstance(Context context, String token)` gets called, looks like we'll set `optOutTrackingDefault` to `false`? how does it work in that case?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -508,6 +550,7 @@ public double eventElapsedTime(final String eventName) {
      * See also {@link #track(String, org.json.JSONObject)}
      */
     public void trackMap(String eventName, Map<String, Object> properties) {
+        if (hasOptedOutTracking()) return;
```
yinfeiru: it's fine but you don't need this (`track(String eventName, JSONObject properties)` will do it)

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -545,6 +589,7 @@ public void track(String eventName, JSONObject properties) {
      * @param eventName the name of the event to send
      */
     public void track(String eventName) {
+        if (hasOptedOutTracking()) return;
```
yinfeiru: it's fine but you don't need this (track(String eventName, JSONObject properties) will do it)

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -757,6 +809,89 @@ public void reset() {
...
+            getPeople().deleteUser();
+            getPeople().clearCharges();
+        }
+        mPersistentIdentity.clearPreferences();
+        synchronized (mEventTimings) {
+            mEventTimings.clear();
+            mPersistentIdentity.clearTimeEvents();
+        }
+        mPersistentIdentity.clearReferrerProperties();
+        mPersistentIdentity.setOptOutTracking(true);
```
yinfeiru: do we want to call `unregisterSuperProperty` as well?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -757,6 +809,89 @@ public void reset() {
...
+     *                   opt-in event that is sent to Mixpanel.
+     *
+     * See also {@link #optInTracking()} and {@link #optOutTracking()}.
+     */
+    public void optInTracking(String distinctId, JSONObject properties) {
+        mPersistentIdentity.setOptOutTracking(false);
+        if (distinctId != null) {
+            identify(distinctId);
+        }
+        track("$opt_in", properties);
```
yinfeiru: you guys probably have discussed this before but please make sure we send the same event in all the SDKs.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1452,6 +1597,7 @@ public void set(JSONObject properties) {
 
         @Override
         public void set(String property, Object value) {
+            if (hasOptedOutTracking()) return;
```
yinfeiru: redundant

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1461,6 +1607,7 @@ public void set(String property, Object value) {
 
         @Override
         public void setOnceMap(Map<String, Object> properties) {
+            if (hasOptedOutTracking()) return;
```
yinfeiru: redundant

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1484,6 +1632,7 @@ public void setOnce(JSONObject properties) {
 
         @Override
         public void setOnce(String property, Object value) {
+            if (hasOptedOutTracking()) return;
```
yinfeiru: redundant

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1577,9 +1733,10 @@ public InAppNotification getNotificationIfAvailable() {
 
         @Override
         public void trackNotificationSeen(InAppNotification notif) {
-            if(notif == null) return;
-
+            if (notif == null) return;
             mPersistentIdentity.saveCampaignAsSeen(notif.getId());
+            if (hasOptedOutTracking()) return;
```
yinfeiru: we need to be careful about this since decide uses `$campaign_delivery` to determine if a notification should be included in the decide response or not.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1621,6 +1778,7 @@ public void showGivenNotification(final InAppNotification notif, final Activity
 
         @Override
         public void trackNotification(final String eventName, final InAppNotification notif, final JSONObject properties) {
+            if (hasOptedOutTracking()) return;
```
yinfeiru: same here. be careful!

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2112,6 +2278,7 @@ private void recordPeopleMessage(JSONObject message) {
     }
 
     private void pushWaitingPeopleRecord() {
+        if (hasOptedOutTracking()) return;
```
yinfeiru: it's fine but do you need to do this?
1. `sendAllPeopleRecords` does the check already
2. it only gets called by `identify` which does the check too

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2104,6 +2269,7 @@ protected void track(String eventName, JSONObject properties, boolean isAutomati
     }
 
     private void recordPeopleMessage(JSONObject message) {
+        if (hasOptedOutTracking()) return;
```
yinfeiru: since `deleteUser` calls `recordPeopleMessage`, have you tested and made sure that by calling `optOutTracking` the profile will actually be removed?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2041,11 +2205,12 @@ public void removeOnMixpanelTweaksUpdatedListener(OnMixpanelTweaksUpdatedListene
 
     ////////////////////////////////////////////////////
     protected void flushNoDecideCheck() {
```
yinfeiru: i've never used this guy. what does it do exactly?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -2121,6 +2288,7 @@ private void pushWaitingPeopleRecord() {
     // MUST BE THREAD SAFE. Called from crazy places. mPersistentIdentity may not exist
     // when this is called (from its crazy thread)
     private void sendAllPeopleRecords(JSONArray records) {
+        if (hasOptedOutTracking()) return;
```
yinfeiru: same as above, idk if `deleteUser` will work as expected due to this change.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: since you call `mPersistentIdentity.setOptOutTracking(true);` at the end of `optOutTracking`, it might work. but do you think there's a race condition? if the people queue gets flushed after `mPersistentIdentity.setOptOutTracking(true);` gets called, it won't send out the `deleteUser` event.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -757,6 +809,89 @@ public void reset() {
         return mDeviceInfo;
     }
 
+    /**
+     * Use this method to opt-out a user from tracking. Events and people updates that haven't been
+     * flushed yet will be deleted. Use {@link #flush()} before calling this method if you want
+     * to send all the queues to Mixpanel before.
```
yinfeiru: is it allowed to flush the events on the device after the user opt out in gdpr?

src/main/java/com/mixpanel/android/mpmetrics/PersistentIdentity.java
```diff
@@ -574,6 +624,35 @@ private void readIdentities() {
...
+        SharedPreferences prefs = null;
+        try {
+            prefs = mMixpanelPreferences.get();
+        } catch (final ExecutionException e) {
+            MPLog.e(LOGTAG, "Cannot read opt out flag from sharedPreferences.", e.getCause());
+        } catch (final InterruptedException e) {
+            MPLog.e(LOGTAG, "Cannot read opt out flag from sharedPreferences.", e);
+        }
+
+        if (null == prefs) {
```
yinfeiru: well.. i might be too picky but this is not how to do this kind of check at other places.
eg. `if (mIsUserOptOut == null)`

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: Yes

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: that's just the default flag - it won't override user's action

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: yeah - defensive programming since i can't guarantee this method won't change in the future

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: same as above

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: `mPersistentIdentity.clearPreferences` removes all the super properties

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: see comment below

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: see comment above

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: see comment above

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: flushes events/people queues without hitting decide

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: if the user wants to send everything that has been tracked - yes.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: nope, there is not such case. the analytics worker executes one message at a time.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: yup, it will!

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
patedit: yup! again, in someone adds other functionality it's hard to keep track of things. defensive programming will assure in this case that nothing breaks in a different release

src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
yinfeiru: i'm not sure if i follow. i think `deleteUser` enqueues a people message but doesn't flush the queue. if the queue gets flushed after `mPersistentIdentity.setOptOutTracking(true)` getting called, the message won't be sent.

patedit: We don't have any checks on the `flush` method, so the message will be sent :)

PR #556, opened 2018-06-11
Remove get_accounts requirement

argenisf:
Remove the check of the get_accounts permission unless the version if below jelly bean


patedit:
Thanks @argenisf !

PR #601, opened 2019-06-05
Remove unused resources.

ming13:
I’ve been researching our APK and noticed a large number of PNG files we cannot have since we use vector images almost everywhere. After a further analysis Mixpanel SDK proved to be a part of an issue. Since it targets API 14 it is impossible to ship only vector images.

At the same time, the `UnusedResources` Lint option is disabled in `build.gradle`. Even worse — enabling it doesn’t produce any Lint warnings, most likely because Lint thinks about library resources as potentially used by applications. That’s why I used IJ to find unused resources by hand and removed them.

* AAR size before — 555 KB
* AAR size after — 427 KB

It would be great to ship this ASAP since it affects all applications using the SDK. Multiply this by a combined number of users of these application and you can imagine network bandwich and storage savings.


patedit:
@ming13 thanks a lot for taking a look at this. By quickly looking into the PR, I think we might be removing too many things. In-app notifications rely on some resources to be part of the app to properly render. I'll take a look at your PR and find out what can't be removed. Thanks a ton

ming13:
@patedit, indeed it uses some resources, but those I left intact. I used the Safe Delete option, so the IDE searched references before removing. Also I’ve run `ag` on a number of them just to be sure there are no weird reflection tricks. Also, it wouldn’t compile with missing resources 😉

ming13:
@patedit, sorry to bother you with this, but have you got a chance to take a look at this PR? We would like to benefit a reduced APK size 😉

patedit:
Hi @ming13 - thanks for submitting this PR. Reason I had some concerns is because we used to load those assets dynamically for in-apps. So the name of the resource would come from our backend and we would load that file to display the in-app. It's no longer the case and we use URLs instead. Everything looks great. Will merge this and will be available on the next release! Thanks again!

ming13:
@patedit, thank you!

Issue #608, opened 2019-07-15
Trying to initialize FCM although not using Firebase


ori-6over6:
After upgrading mixpanel from version 4+ to 5.6.4 my app crashes on launch.
The exception I'm getting is `IllegalStateException: Default FirebaseApp is not initialized in this process com.sixoversix.copyglass. Make sure to call FirebaseApp.initializeApp(Context) first` when calling `MixpanelAPI.getInstance(context, token);`
I'm not using firebase in this app so I didn't expect Mixpanel SDK to do firebase related initializations.

When debugging the code I saw that the firebase initialization is executed because the function `checkPushNotificationConfiguration` returns true.
When further debugging I saw it returns true because I'm using a different sdk (intercom) which internally uses firebase.

See screenshot for clarifying what I found in debug:
![Screen Shot 2019-07-15 at 2 25 37 PM](https://user-images.githubusercontent.com/38836284/61213460-c8653480-a70d-11e9-9644-5e3f1fa92d89.png)

Please let me know what can I do to overcome this.
Thanks!

patedit:
Hi @ori-6over6 - wow, I was not expecting this at all. I assume that intercom includes that entry in the SDK automatically without you having to add that entry on your `AndroidManifest.xml` ?

ori-6over6:
Hi @patedit, yes you are correct, this service is not defined in our Manifest.

patedit:
@ori-6over6 I just open a PR to address this issue. will be available in the next release. Thanks a lot for bringing this up!

ori-6over6:
@patedit Thanks a lot for the quick fix!
Glad I could help :)

ori-6over6:
Hey @patedit , I see a new version 5.6.5 was released but I'm still experiencing the issue...
Was the fix included the new version?

patedit:
hi @ori-6over6 - sorry, it was not included. The PR I created won't be compatible with our react native sdk so decided to not merge it until a better fix is applied. is this affecting your production build and/or did you come up with a workaround?

ori-6over6:
Hey @patedit, we didn't find any workaround, so this issue is preventing us from upgrading our mixpanel dependency - we are still using version 4.9.4

patedit:
@ori-6over6 please, update to v5.6.7. This should be fixed on that release. Thanks for your patience :)

ori-6over6:
@patedit updated your sdk version and the issue is resolved. Thank you :)

PR #609, opened 2019-07-18
Better check before enabling fcm

patedit:
Fixes https://github.com/mixpanel/mixpanel-android/issues/608


patedit:
won't work with react native. review.

PR #614, opened 2019-08-05
setShowWhen to true

jose96043:
For apps targeting Build.VERSION_CODES.N and above, time is not shown anymore by default by using setWhen and must be opted into by using setShowWhen(boolean)


patedit:
Thanks @jose96043 !!

PR #619, opened 2019-09-09
Add button support for push notifications

jbwyme:
adds support for a new push data key `mp_buttons` to specify action buttons in a push notification

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java
@@ -140,7 +140,11 @@ public class MixpanelFCMMessagingService extends FirebaseMessagingService {
         Notification notification = mixpanelPushNotification.createNotification(messageIntent);
         if (null != notification) {
             final NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
-            notificationManager.notify(0, notification);
+            if (mixpanelPushNotification.data.tag != null) {
+                notificationManager.notify(mixpanelPushNotification.data.tag, 0, notification);
+            } else {
+                notificationManager.notify(0, notification);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java
```diff
@@ -424,6 +456,35 @@ protected static Notification makeNotificationSDK21OrHigher(Context context, Pen
...
+        if (null != buttonsJsonStr) {
+            try {
+                JSONArray buttonsArr = new JSONArray(buttonsJsonStr);
+                for (int i = 0; i < buttonsArr.length(); i++) {
+                    JSONObject buttonObj = buttonsArr.getJSONObject(i);
+
+                    // get button icon from name if one sent
+                    final String btnIconName = buttonObj.optString("icnm");
+                    int btnIcon = MixpanelFCMMessagingService.NotificationData.NOT_SET;
+                    if (null != btnIconName) {
```
patedit: `btnIconName` will be empty if it doesn't exist

src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java
```diff
@@ -260,6 +279,17 @@ protected static Notification makeNotificationSDK26OrHigher(Context context, Pen
                 setStyle(new Notification.BigTextStyle().bigText(notificationData.message)).
                 setChannelId(channelId);
 
+        for (int i = 0; i < notificationData.buttons.size(); i++) {
```
patedit: why not adding this feature to API 20 and above instead of 26 and above?

jbwyme: idk i was waiting for you to tell me what to do, so should I refactor it to a utility function and call it in both appropriate `makeNotification` functions?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java
jbwyme: ah optString returns `""` eh?

PR #620, opened 2019-09-10
Push notification refactor and feature additions

jbwyme:
This refactor:
- makes it easier to opt-in new features for push notifications by conditionally calling `Notification.Builder.*` apis when appropriate vs having separate methods for each Android SDK version range.
- uses a class instance for better testability and mocking
- has more test coverage than before

This PR also adds a number of new features for push notifications such as the ability to add action buttons, expanded images, etc.


src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java
```diff
@@ -170,8 +136,8 @@ public static void showPushNotification(Context context, Intent messageIntent) {
 
         final ResourceIds drawableIds = new ResourceReader.Drawables(resourcePackage, context);
         final Context applicationContext = context.getApplicationContext();
-        final Notification notification = buildNotification(applicationContext, messageIntent, drawableIds);
-
+        MixpanelPushNotification mixpanelPushNotification = new MixpanelPushNotification(applicationContext, drawableIds);
```
patedit: we should keep the implementation so developers can override how the notification is built if they want to add new features or modify settings (that's why `makeNotificationIf...` methods were protected if devs are overriding `MixpanelFCMMessagingService`).

We need a way to let developers override the notification before its posted to the notificationmanager

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -0,0 +1,505 @@
...
+        final String groupKey = inboundIntent.getStringExtra("mp_groupkey");
+        final String ticker = inboundIntent.getStringExtra("mp_ticker");
+        final String stickyString = inboundIntent.getStringExtra("mp_sticky");
+        final String timeString = inboundIntent.getStringExtra("mp_time");
+        final int visibility = inboundIntent.getIntExtra("mp_visibility", Notification.VISIBILITY_PRIVATE);
+        final String silent = inboundIntent.getStringExtra("mp_silent");
+
+        trackCampaignReceived(campaignId, messageId, extraLogData);
+
+        if (message == null) {
```
patedit: a push notification with no body is still shown - are we sending empty strings? we need to think how we are handling silent push with no body

src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java
```diff
@@ -195,289 +136,47 @@ public void process(MixpanelAPI api) {
...
+    public static void showPushNotification(Context context, Intent messageIntent, MixpanelPushNotification mixpanelPushNotification) {
+        Notification notification = mixpanelPushNotification.createNotification(messageIntent);
 
-        int largeNotificationIcon = MixpanelFCMMessagingService.NotificationData.NOT_SET;
-        if (null != largeIconName) {
-            if (iconIds.knownIdName(largeIconName)) {
-                largeNotificationIcon = iconIds.idFromName(largeIconName);
-            }
-        }
+        MPLog.d(LOGTAG, "MP FCM notification received: " + mixpanelPushNotification.getData().getMessage());
```
patedit: since `mData` could be null at this point, let's check for npe

PR #625, opened 2019-09-17
push notifications fields update

aecast:
add support for various fields in android push notifs


jbwyme:
Can you redo the commit history so that it looks like:
- add thumbnail and expanded image support (with tests)
- add badge number support (with tests)

PR #628, opened 2019-09-18
add push builder fields support

aecast:
added support for subtitles, expandable images, and notification badges along with tests


src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -85,6 +88,14 @@ public void testNoTitle() {
...
     }
 
+    public void testSubTitle() {
+        final Intent intent = new Intent();
+        intent.putExtra("mp_message", "MESSAGE");
+        intent.putExtra("mp_stitle", "TITLE");
+        mpPushSpy.createNotification(intent);
+        verify(builderSpy).setSubText("TITLE");
+    }
+
```
jbwyme: add test for when no `mp_stitle` is set please

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -108,6 +119,32 @@ public void testInvalidIcon() {
         verify(builderSpy).setSmallIcon(DEFAULT_ICON_ID);
     }
 
+    public void testExpandedImageUsingValidUrl() throws ImageStore.CantGetImageException{
```
jbwyme: space before trailing `{`

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -19,6 +21,7 @@
 
     private final String VALID_RESOURCE_NAME = "com_mixpanel_android_logo";
     private final int VALID_RESOURCE_ID = R.drawable.com_mixpanel_android_logo;
+    private final String VALID_URL = "https://dev.images.mxpnl.com/1939595/fc767ba09b1a5420bdbcee71c7ae9904.png";
```
jbwyme: `s/VALID_URL/VALID_IMAGE_URL/`

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -108,6 +119,32 @@ public void testInvalidIcon() {
...
+        verify(mpPushSpy).getBitmapFromUrl(VALID_URL);
+        verify(builderSpy).setStyle(new Notification.BigPictureStyle().bigPicture(fakeBitmap));
+    }
+
+    public void testExpandedImageUsingInvalidUrl() throws ImageStore.CantGetImageException{
+        final Intent intent = new Intent();
+        intent.putExtra("mp_message", "MESSAGE");
+        intent.putExtra("mp_imgurl", "");
+
+        when(mpPushSpy.getBitmapFromUrl(VALID_URL)).thenReturn(null);
```
jbwyme: `mpPushSpy.getBitmapFromUrl(VALID_URL)` will never be called because you're sending an empty string to `mp_imgurl`. I don't think you need to mock this method because it shouldn't actually be called at all.

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -108,6 +119,32 @@ public void testInvalidIcon() {
...
+
+        mpPushSpy.createNotification(intent);
+        verify(mpPushSpy).getBitmapFromUrl(VALID_URL);
```
jbwyme: the only reason this passes is because it gets called on line 141. I'd remove that line and modify this to verify `getBitmapFromUrl` is never called.

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -108,6 +119,32 @@ public void testInvalidIcon() {
...
+        verify(builderSpy).setStyle(new Notification.BigPictureStyle().bigPicture(fakeBitmap));
+    }
+
+    public void testExpandedImageUsingInvalidUrl() throws ImageStore.CantGetImageException{
+        final Intent intent = new Intent();
+        intent.putExtra("mp_message", "MESSAGE");
+        intent.putExtra("mp_imgurl", "");
```
jbwyme: nit: empty strings sometimes have special meaning (generally due to a bad design decision but nevertheless). for this test case let's use something a little more explicit like `intent.putExtra("mp_imgurl", "this is a bad url")`

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -200,6 +237,22 @@ public void testNoActionButtons() {
...
+        intent.putExtra("mp_message", "MESSAGE");
+        intent.putExtra("mp_badgecount", "2");
+        mpPushSpy.createNotification(intent);
+        verify(builderSpy).setNumber(2);
+    }
+
+    public void testInvalidNotificationBadge() {
+        final Intent intent = new Intent();
+        intent.putExtra("mp_message", "MESSAGE");
+        intent.putExtra("mp_badgecount", "0");
```
jbwyme: what if the user actually wants to set the count to 0 (e.g. remove the badge)? What does the Android API do if you pass 0 to `setBadgeCount`?

More generally, if there are a number of different ways something can be invalid, let's have a test case for each one. E.g. `testNotificationBadgeCountLessThan0`, `testNotificationBadgeCountNotANumber` etc. This provides better test coverage and perhaps even more importantly, documents the edge cases we've identified and are explicitly handling.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -60,19 +61,27 @@ public Notification createNotification(Intent inboundIntent) {
...
         maybeSetChannel();
+        maybeSetNotificationBadge(data);
 
         final Notification n = buildNotification();
         n.flags |= Notification.FLAG_AUTO_CANCEL;
 
         return n;
     }
 
+    protected void maybeSetSubTitle(NotificationData data) {
```
jbwyme: should this be `maybeSetSubText`? Does calling `builder.setSubText` result in that text looking like a subtitle? My impression was it goes beneath the message body.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -91,7 +100,34 @@ protected void maybeSetLargeIcon(NotificationData data) {
 
     protected void maybeSetExpandableNotification(NotificationData data) {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
-            builder.setStyle(new Notification.BigTextStyle().bigText(data.message));
+            if (null != data.expandableImageUrl && data.expandableImageUrl.startsWith("http")) {
+                try {
+                    Bitmap imageBitmap = getBitmapFromUrl(data.expandableImageUrl);
+
+                    if (imageBitmap == null) {
+                        setBigTextStyle(data.message);
```
jbwyme: Let's add some debug logging for every possible flow to help developers debug. Here's an example:
```java
MPLog.d(LOGTAG, "getBitmapFromUrl returned null, setting BigTextStyle instead. Image URL: " + data.expandableImageUrl);
```

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -153,25 +189,39 @@ protected void maybeSetChannel() {
         }
     }
 
+    protected void maybeSetNotificationBadge(NotificationData data) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.BASE && data.badgeCount > 0) {
```
jbwyme: just to repeat the question I had in the test code, what happens if you set badgeCount to 0?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -153,25 +189,39 @@ protected void maybeSetChannel() {
...
+            builder.setNumber(data.badgeCount);
+        }
+    }
+
     protected Bitmap getBitmapFromResourceId(int resourceId) {
         return BitmapFactory.decodeResource(context.getResources(), resourceId);
     }
 
+    protected Bitmap getBitmapFromUrl(String url) throws ImageStore.CantGetImageException {
+        ImageStore is = new ImageStore(context, "Push Notifications");
```
jbwyme: s/"Push Notifications"/"MixpanelPushNotification"

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -153,25 +189,39 @@ protected void maybeSetChannel() {
...
+        return is.getImage(url);
+    }
+
     /* package */ NotificationData readInboundIntent(Intent inboundIntent) {
 
         final String message = inboundIntent.getStringExtra("mp_message");
         final String iconName = inboundIntent.getStringExtra("mp_icnm");
         final String largeIconName = inboundIntent.getStringExtra("mp_icnm_l");
         final String whiteIconName = inboundIntent.getStringExtra("mp_icnm_w");
+        final String expandableImageURL = inboundIntent.getStringExtra("mp_imgurl");
```
jbwyme: let's go with `mp_img`

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -153,25 +189,39 @@ protected void maybeSetChannel() {
...
         final String uriString = inboundIntent.getStringExtra("mp_cta");
         CharSequence notificationTitle = inboundIntent.getStringExtra("mp_title");
+        CharSequence notificationSubTitle = inboundIntent.getStringExtra("mp_stitle");
```
jbwyme: let's go with `mp_subtxt`

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -153,25 +189,39 @@ protected void maybeSetChannel() {
...
         final String colorName = inboundIntent.getStringExtra("mp_color");
         final String buttonsJsonStr = inboundIntent.getStringExtra("mp_buttons");
         final String campaignId = inboundIntent.getStringExtra("mp_campaign_id");
         final String messageId = inboundIntent.getStringExtra("mp_message_id");
         final String extraLogData = inboundIntent.getStringExtra("mp");
         int color = NotificationData.NOT_SET;
         List<NotificationButtonData> buttons = new ArrayList<>();
+        final String badgeCountStr = inboundIntent.getStringExtra("mp_badgecount");
```
jbwyme: let's go with `mp_bdgcnt`

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
aecast: Doesn't change behavior at all so I though I might as well just not call it

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
aecast: Changed these tests

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
aecast: Changed these tests they should actually verify useful calls in next PR

PR #631, opened 2019-09-23
Update push notifications

aecast:
Add support for event times and visibility to android push notifications


src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -342,27 +428,51 @@ protected Notification buildNotification() {
...
+        public final String channelId;
+        public final String tag;
+        public final String groupKey;
+        public final String ticker;
+        public final boolean sticky;
+        public final String timeString;
+        public final int visibility;
 
         public static final int NOT_SET = -1;
+        public static final int DEFAULT_PRIORITY = 3;
```
jbwyme: This doesn't seem to be used

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -1,37 +1,51 @@
...
     private final int DEFAULT_ICON_ID = android.R.drawable.sym_def_app_icon;
     private final Intent DEFAULT_INTENT = new Intent(Intent.ACTION_BUG_REPORT); // ACTION_BUG_REPORT is chosen because it's identifiably weird
+    private Context context;
 
     @Override
     public void setUp() throws Exception {
         super.setUp();
         System.setProperty("org.mockito.android.target", getContext().getCacheDir().getPath());
 
+        this.context = getContext();
```
jbwyme: do we need to do this? doesn't `getContext()` always return the same thing? If we do need to do this, can't we assign it at the same time we declare it (line 37)?

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -85,6 +99,14 @@ public void testNoTitle() {
         verify(builderSpy).setContentTitle(DEFAULT_TITLE);
     }
 
+    public void testSubTitle() {
```
jbwyme: We already have a test for this below called `testSubText`

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -200,6 +248,116 @@ public void testNoActionButtons() {
...
+        intent.putExtra("mp_message", "MESSAGE");
+        intent.putExtra("mp_bdgcnt", 2);
+        mpPushSpy.createNotification(intent);
+        verify(builderSpy).setNumber(2);
+    }
+
+    public void testInvalidNotificationBadge() {
+        final Intent intent = new Intent();
+        intent.putExtra("mp_message", "MESSAGE");
+        intent.putExtra("mp_bdgcnt", "0");
```
jbwyme: it's unclear whether this is invalid because 0 is not valid or because it's a string. I'd recommend being a bit more specific here or potentially having two tests if both are invalid to make it clear what our expectations are.

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -200,6 +248,116 @@ public void testNoActionButtons() {
...
+        Notification notification = mpPushSpy.createNotification(intent);
+        verify(builderSpy).setChannelId("12345");
+        assertEquals(notification.getChannelId(), "12345");
+    }
+
+    public void testNoChannelId() {
+        final Intent intent = new Intent();
+        intent.putExtra("mp_message", "MESSAGE");
+        Notification notification = mpPushSpy.createNotification(intent);
+        verify(builderSpy).setChannelId(anyString());
```
jbwyme: can't we be more specific here and verify that `setChannelId(MixpanelPushNotification.NotificationData.DEFAULT_CHANNEL_ID)`?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -57,22 +60,46 @@ public Notification createNotification(Intent inboundIntent) {
                 setWhen(now).
                 setContentTitle(data.title).
                 setContentText(data.message).
-                setTicker(data.message).
+                setTicker(data.ticker == null ? data.message : data.ticker).
                 setContentIntent(contentIntent);
 
+        maybeSetSubTitle(data);
```
jbwyme: let's just rename everything related to `SubTitle` to `SubText`

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -200,6 +248,116 @@ public void testNoActionButtons() {
...
+        intent.putExtra("mp_subtxt", "SUBTEXT");
+        Notification notification = mpPushSpy.createNotification(intent);
+        verify(builderSpy).setSubText("SUBTEXT");
+    }
+
+    public void testNoSubText() {
+        final Intent intent = new Intent();
+        intent.putExtra("mp_message", "MESSAGE");
+        Notification notification = mpPushSpy.createNotification(intent);
+        verify(builderSpy, never()).setSubText("MESSAGE");
```
jbwyme: This test would still pass if `setSubText` is called with anything other than `"Message"`. This should be verifying the it's not called with any thing.

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -200,6 +248,116 @@ public void testNoActionButtons() {
...
+        assertEquals(flag | Notification.FLAG_AUTO_CANCEL, flag);
+    }
+
+    public void testTimestamp() {
+        final Intent intent = new Intent();
+        intent.putExtra("mp_message", "MESSAGE");
+        intent.putExtra("mp_time", "2014-10-02T15:01:23.045123456Z");
+        Notification notification = mpPushSpy.createNotification(intent);
+        Instant instant = Instant.parse( "2011-05-03T11:58:01Z" );
+        verify(builderSpy).setWhen(instant.toEpochMilli());
```
jbwyme: I don't understand this test. Why do we set the time to be `"2014-10-02T15:01:23.045123456Z"` and then verify that it was set with `"2011-05-03T11:58:01Z"`?

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
```diff
@@ -200,6 +248,116 @@ public void testNoActionButtons() {
...
+    }
+
+    public void testNoTimestamp() {
+        final Intent intent = new Intent();
+        intent.putExtra("mp_message", "MESSAGE");
+        verify(builderSpy, never()).setWhen(any(Long.class));
```
jbwyme: Just curious: do we need to `setWhen` to the current time or does it just default to it?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java
```diff
@@ -140,7 +140,11 @@ public static void showPushNotification(Context context, Intent messageIntent) {
         Notification notification = mixpanelPushNotification.createNotification(messageIntent);
         if (null != notification) {
             final NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
-            notificationManager.notify(0, notification);
+            if (mixpanelPushNotification.tag != null) {
+                notificationManager.notify(mixpanelPushNotification.tag ,0, notification);
```
jbwyme: remove space after `.tag` and add space after the comma: `notificationManager.notify(mixpanelPushNotification.tag, 0, notification);`

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -57,22 +60,46 @@ public Notification createNotification(Intent inboundIntent) {
...
         return n;
     }
 
+    protected void maybeSetGroupKey(NotificationData data, PendingIntent contentIntent) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && data.groupKey != null) {
+            builder.setGroup(data.groupKey);
+        }
+    }
+
+    protected void maybeSetTag(NotificationData data) {
```
jbwyme: so this isn't a `maybeSetTag`, it always sets it. Can we just do this as a part of the base notification setup?

jbwyme: actually this is weird, it's basically a leaky abstraction

jbwyme: thought about it more, seems ok given the constraints

jbwyme: oh but we should still just do this in `createNotification` instead of in this method

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -126,52 +179,80 @@ protected PendingIntent createActionIntent(String uri) {
...
+    protected void maybeSetNotificationBadge(NotificationData data) {
+        if (data.badgeCount > 0) {
+            builder.setNumber(data.badgeCount);
+        }
+    }
 
-            mNotificationManager.createNotificationChannel(channel);
+    protected void maybeSetTime(NotificationData data) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && data.timeString != null) {
+            Instant instant = Instant.parse( "2011-05-03T11:58:01Z" );
```
jbwyme: ???? hard coded?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -126,52 +179,80 @@ protected PendingIntent createActionIntent(String uri) {
...
 
     protected Bitmap getBitmapFromResourceId(int resourceId) {
         return BitmapFactory.decodeResource(context.getResources(), resourceId);
     }
 
+    protected Bitmap getBitmapFromUrl(String url) {
+        ImageStore is = new ImageStore(context, "MixpanelPushNotification");
+        try {
+            Bitmap bm = is.getImage(url);
+            return bm;
```
jbwyme: just return `is.getImage(url)` no need to create a local var

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
aecast: I'm actually not sure of the reason but it breaks the testing with a null reference error if done purely as an instance variable

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
aecast: Deleted

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
aecast: Deleted

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
aecast: Invalid because 0 doesn't add any change to the notification; should've been an integer

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
aecast: I believe it the default is handled when the notification arrives to the device. I did also need to add setShowWhen(bool) as well though.

src/androidTest/java/com/mixpanel/android/mpmetrics/MixpanelNotificationBuilderTest.java
aecast: Stupid copy paste error on my end

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
aecast: I thought we wanted to return null instead of throwing the exception or does that happen anyways?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
aecast: Stupid mistake on my end

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
aecast: if we are making it part of the default setup what should the default tag val be?

PR #635, opened 2019-09-30
Fix id issue with push tags

aecast:
Generates a random id instead of using 0 so the device will, by default, show multiple notifications instead of always replacing the notification


src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java
```diff
@@ -144,10 +146,11 @@ public static void showPushNotification(Context context, Intent messageIntent) {
 
         if (null != notification) {
             final NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+            final int id = new Random().nextInt(Integer.MAX_VALUE);
             if (mixpanelPushNotification.data.tag != null) {
                 notificationManager.notify(mixpanelPushNotification.data.tag, 0, notification);
```
jbwyme: can we add some comments on why we pass 0 as an id for future reference?

PR #636, opened 2019-09-30
add silent push support

aecast:


src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -84,6 +85,8 @@ protected void parseIntent(Intent inboundIntent) {
             notificationSubText = null;
         }
 
+        boolean isSilent = silent != null && silent.equals("true") ? true : false;
```
jbwyme: simplified: `boolean isSilent = silent != null && silent.equals("true")`

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -187,7 +190,7 @@ protected void buildNotificationFromData() {
     protected Notification createNotification(Intent inboundIntent) {
         this.parseIntent(inboundIntent);
 
-        if (this.data == null) {
+        if (this.data == null || this.data.silent) {
```
jbwyme: it'd be useful to put this into it's own conditional just so we can log a message that says why we aren't showing a message. E.g. `MPLog.d(LOGTAG, "Notification will not be shown because 'mp_silent = true'")`

PR #637, opened 2019-09-30
Notification action button tracking support

aecast:
Before this change our notification system supported notification button actions but there was no way to track when they were clicked as android would launch the activity described in the button's intent in a context outside of our control. 

To introduce tracking we now use a the MixpanelNotificationRouteActivity as a liaison between our button actions and the OS launching the intended activity so that we can make tracking calls with the MixpanelAPI.


src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -0,0 +1,95 @@
...
+        MixpanelAPI.allInstances(new MixpanelAPI.InstanceProcessor() {
+            @Override
+            public void process(MixpanelAPI api) {
+                JSONObject pushProps = new JSONObject();
+
+                try {
+                    pushProps.put("actionId", actionId);
+                    pushProps.put("uri", uri);
+                    pushProps.put("messageId", messageId);
+                    pushProps.put("campaignId", campaignId);
```
jbwyme: we should use snake_case for these properties to match all of our other message tracking events: `action_id`, `message_id`, `campaign_id`. Also should we call it `button_id` instead of `action_id`? I'm trying to think through whether the `action` abstraction makes sense but I don't know.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -277,31 +295,51 @@ protected void maybeAddActionButtons() {
...
     }
 
-    protected PendingIntent createActionIntent(String uri) {
-        return PendingIntent.getActivity(
-                context,
-                0,
-                new Intent(Intent.ACTION_VIEW, Uri.parse(uri)),
-                PendingIntent.FLAG_UPDATE_CURRENT
-        );
+    protected PendingIntent createActionIntent(String uri, String actionId) {
```
jbwyme: Perhaps add some javadoc to this method describing what it does, particularly for things like `FLAG_ACTIVITY_NO_HISTORY`.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -0,0 +1,95 @@
...
+                } catch (JSONException e) {
+                    MPLog.e(LOGTAG, "Error loading tracking JSON properties.");
+                }
+                api.track("Notification Action Click", pushProps);
```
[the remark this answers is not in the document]
aecast: https://github.com/mixpanel/mixpanel-android/pull/642/files fixed here

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
aecast: https://github.com/mixpanel/mixpanel-android/pull/642/files fixed here

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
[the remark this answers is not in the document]
aecast: fixed here **https://github.com/mixpanel/mixpanel-android/pull/642/files**

src/main/java/com/mixpanel/android/mpmetrics/MixpanelWebViewActivity.java
```diff
@@ -0,0 +1,36 @@
...
+        WebSettings settings = webView.getSettings();
+
+        settings.setJavaScriptEnabled(true);
+        settings.setAllowFileAccess(false);
+        settings.setBuiltInZoomControls(false);
+        settings.setUseWideViewPort(true);
+        settings.setLoadWithOverviewMode(true);
+        settings.setDomStorageEnabled(true);
+
+        webView.loadUrl(getIntent().getExtras().getString("uri"));
```
jbwyme: we should validate somewhere in the call stack the URI is valid

jbwyme:
let's try adding these activities into our local manifest instead

src/main/AndroidManifest.xml
```diff
@@ -75,5 +74,16 @@
             </intent-filter>
         </receiver>
         -->
+        <activity android:name="com.mixpanel.android.mpmetrics.MixpanelWebViewActivity">
```
jbwyme: can we add comments above each of these explaining what they're used for so that developers integrating our library will understand?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelActivityLifecycleCallbacks.java
```diff
@@ -10,6 +10,7 @@
 import android.os.Handler;
 import android.os.Looper;
 
+import com.mixpanel.android.util.MPLog;
```
jbwyme: is this used?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -0,0 +1,163 @@
...
+                    return defaultIntent;
+                }
+            case DEEP_LINK:
+                return new Intent(Intent.ACTION_VIEW, Uri.parse(uri.toString()));
+            default:
+                return defaultIntent;
+        }
+    }
+
+    protected boolean isValidURL(CharSequence url, PushTapTarget target) {
```
jbwyme: the better api for this is probably `isValidURL(CharSequence url, boolean requireHttps)` instead of tightly coupling the PushTapTarget stuff

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -0,0 +1,163 @@
...
+
+        final String tapTarget = tapTargetChars.toString();
+        final String buttonId;
+        final String label;
+
+        if (tapTarget.equals(MixpanelPushNotification.TAP_TARGET_BUTTON)) {
+            CharSequence buttonIdChars = intentExtras.getCharSequence("buttonId");
+            if (null == buttonIdChars) {
+                MPLog.d(LOGTAG, "Notification action click logged with no buttonId");
+                return;
```
jbwyme: I would say in such a case, let's allow the track to continue, just set `buttonId` (and `label` below) to null

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -0,0 +1,163 @@
...
+
+            buttonId = buttonIdChars.toString();
+            label = labelChars.toString();
+        } else {
+            buttonId = null;
+            label = null;
+        }
+
+        CharSequence messageIdChars = intentExtras.getCharSequence("messageId");
+        if (null == messageIdChars) {
```
jbwyme: same here, let's just always track and let nulls through. It _shouldn't_ ever happen but I think in the case where it were to happen we'd prefer not to lose the track call completely.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -24,17 +25,40 @@
...
 import java.util.TimeZone;
 
+/**
+ * Routing activity to introduce mixpanel tracking to notification actions
+ *
+ * <p>This class acts as a liaison between a notification's actions and the
+ * handling of that action by Android so that we can make tracking calls to
+ * the Mixpanel API.
+ *
+ * <p>To enable this activity in your apps you must add the activity in the
```
jbwyme: i think this can be removed now since we're using the manifest merging

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -49,54 +73,64 @@ public MixpanelPushNotification(Context context, Notification.Builder builder, R
...
+        int visibility = Notification.VISIBILITY_PRIVATE;
+        if (null != visibilityStr) {
+            visibility = Integer.parseInt(visibilityStr);
+        }
+
+        int color = NotificationData.NOT_SET;
+        if (null != colorName) {
+            try {
+                color = Color.parseColor(colorName);
+            } catch (IllegalArgumentException e) {}
```
jbwyme: `MPLog.d` in the exception block

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -290,31 +289,193 @@ protected void maybeAddActionButtons() {
...
+                        }
+                    }
+
+                    // handle button label
+                    final String btnLabel = buttonObj.getString("lbl");
+
+                    // handle button action
+                    final JSONObject pushActionJSON = buttonObj.getJSONObject("ontap");
+                    final PushTapTarget target = PushTapTarget.fromString(pushActionJSON.getString("type"));
+                    final PushTapAction pushAction = new PushTapAction(target, pushActionJSON.getString("uri"));
```
jbwyme: `uri` can be null in the case of "homescreen" type right? Also, how do we pass in whether to open uri in `webview` or `browser` in the button case? Shouldn't we just be passing `ontap` to `buildOnTap(buttonObj.getString("ontap"))`? Let's add a test case for this if I'm correct.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -290,31 +289,193 @@ protected void maybeAddActionButtons() {
...
+
+                if (!typeFromJSON.equals(PushTapTarget.HOMESCREEN.getTarget())) {
+                    final String uriFromJSON = onTapJSON.getString("uri");
+                    onTap = new PushTapAction(PushTapTarget.fromString(typeFromJSON), uriFromJSON);
+                } else {
+                    onTap = new PushTapAction(PushTapTarget.fromString(typeFromJSON));
+                }
+
+            } catch (JSONException e){
+                MPLog.d(LOGTAG, "Couldn't parse JSON Object for \'mp_ontap\'");
```
jbwyme: if we do end up using this to parse button actions as well, let's make sure we update this comment so it doesn't specify "mp_ontap"

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -290,31 +289,193 @@ protected void maybeAddActionButtons() {
...
+                setClass(context, MixpanelNotificationRouteActivity.class).
+                putExtras(options).
+                setFlags(Intent.FLAG_ACTIVITY_NO_HISTORY);
+
+        verifyIntentPackage(routingIntent);
+        return routingIntent;
+    }
+
+    protected Bundle buildBundle(PushTapAction onTap) {
+        /**
```
jbwyme: why doesn't this just call through to `buildBundle(PushTapAction onTap, String buttonId, CharSequence buttonLabel)`?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -0,0 +1,167 @@
...
+
+        CharSequence campaignIdChars = intentExtras.getCharSequence("campaignId");
+        if (null == campaignIdChars) {
+            MPLog.d(LOGTAG, "Notification action click logged with no campaignId");
+        }
+
+        final String messageId = messageIdChars == null ? null : messageIdChars.toString();
+        final String campaignId = campaignIdChars == null ? null : campaignIdChars.toString();
+
+        MixpanelAPI.allInstances(new MixpanelAPI.InstanceProcessor() {
```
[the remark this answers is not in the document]
aecast: If the app is handling the notification isn't it already running? If not, how do you propose we track if it isn't

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -122,11 +136,38 @@ public MixpanelPushNotification(Context context, Notification.Builder builder, l
...
+
+        int badgeCount = MixpanelNotificationData.NOT_SET;
+        if (null != badgeCountStr) {
+            try {
+                badgeCount = Integer.parseInt(badgeCountStr);
+            } catch(NumberFormatException e) {
+                badgeCount = 1;
+            }
+        }
+        mData.setBadgeCount(badgeCount);
```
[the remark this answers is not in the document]
aecast: I've added resetting if <= 0 but from testing the OS takes care of changes to the displayed badge count when notifications are received or dismissed.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -320,18 +327,172 @@ protected void maybeAddActionButtons() {
...
+     *
+     */
+        Bundle options = buildBundle(onTap);
+        options.putCharSequence("tapTarget", TAP_TARGET_BUTTON);
+        options.putCharSequence("buttonId", buttonId);
+        options.putCharSequence("label", buttonLabel);
+        return options;
+    }
+
+    protected void verifyIntentPackage(Intent intent) {
```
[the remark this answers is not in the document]
aecast: That case shouldn't ever arise if this code is running. Method was left over from checking if there were multiple apps using the sdk to identity the correct one but this turned out to be something we never have to handle. I've removed the method.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationData.java
```diff
@@ -201,12 +210,64 @@ public void setLabel(String label) {
...
 
-        public void setUri(String uri) {
-            this.mUri = uri;
+        public static PushTapTarget fromString(String target) {
+            for (PushTapTarget entry : PushTapTarget.values()) {
+                if (entry.getTarget().equals(target)) {
+                    return entry;
+                }
+            }
+            throw new IllegalArgumentException("No enum found for string: " + target);
```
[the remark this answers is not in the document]
aecast: In other parts of the code we actually handle this by making HOMESCREEN the default action in the scenario that we are given a bad string so really this exception should never be thrown or crash the app.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -122,11 +137,42 @@ public MixpanelPushNotification(Context context, Notification.Builder builder, l
...
         mData.setTicker(ticker);
         mData.setTimeString(timeString);
+        mData.setCampaignId(campaignId);
+        mData.setMessageId(messageId);
+        mData.setButtons(buildButtons(buttonsJsonStr));
+
+        int badgeCount = MixpanelNotificationData.NOT_SET;
+        if (null != badgeCountStr) {
+            try {
+
```
patedit: extra line

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -122,11 +137,42 @@ public MixpanelPushNotification(Context context, Notification.Builder builder, l
...
+                badgeCount = Integer.parseInt(badgeCountStr);
+                if (badgeCount < 0) {
+                    badgeCount = 0;
+                }
+            } catch(NumberFormatException e) {
```
patedit:
```suggestion
            } catch (NumberFormatException e) {
```

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -122,11 +137,42 @@ public MixpanelPushNotification(Context context, Notification.Builder builder, l
...
+                badgeCount = 0;
+            }
+        }
+        mData.setBadgeCount(badgeCount);
+
+        int visibility = Notification.VISIBILITY_PRIVATE;
+        if (null != visibilityStr) {
+            switch (visibilityStr) {
+                case VISIBILITY_SECRET:
```
patedit: For clarity, I'd prefer if we are more explicit. eg. `MixpanelPushNotification.VISIBILITY_SECRET`. Applicable to all VISIBILITY_*

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -320,18 +332,162 @@ protected void maybeAddActionButtons() {
...
+
+                if (!typeFromJSON.equals(MixpanelNotificationData.PushTapTarget.HOMESCREEN.getTarget())) {
+                    final String uriFromJSON = onTapJSON.getString("uri");
+                    onTap = new MixpanelNotificationData.PushTapAction(MixpanelNotificationData.PushTapTarget.fromString(typeFromJSON), uriFromJSON);
+                } else {
+                    onTap = new MixpanelNotificationData.PushTapAction(MixpanelNotificationData.PushTapTarget.fromString(typeFromJSON));
+                }
+
+                if (onTap.getActionType().getTarget().equals(MixpanelNotificationData.PushTapTarget.ERROR.getTarget())) {
+                    hasOnTapError = true;
```
patedit: If an unknown target is passed, I think we should default it to open the HOMESCREEN (ie let the OS handle the tap) and still display it. We should log an error / warning in console, too. wdyt?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -320,18 +332,162 @@ protected void maybeAddActionButtons() {
...
+                setClass(mContext, MixpanelNotificationRouteActivity.class).
+                putExtras(options).
+                setFlags(Intent.FLAG_ACTIVITY_NO_HISTORY);
+
+        return routingIntent;
+    }
+
+    protected Bundle buildBundle(MixpanelNotificationData.PushTapAction onTap) {
+        /**
+         * Util method to let subclasses customize the payload through the push notification intent.
```
patedit: we add documentation above method signature, not inside the method :)

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -320,18 +332,162 @@ protected void maybeAddActionButtons() {
...
+
+                    //handle button id
+                    final String btnId = buttonObj.getString("id");
+
+                    if (pushAction == null || btnLabel == null || btnId == null) {
+                        MPLog.d(LOGTAG, "Null button data received. No buttons will be rendered.");
+                    } else {
+                        buttons.add(new MixpanelNotificationData.MixpanelNotificationButtonData(btnLabel, pushAction, btnId));
+                    }
+
```
patedit: extra line

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -0,0 +1,168 @@
...
+
+    protected boolean isValidURL(CharSequence url, boolean requireHttps) {
+        if (requireHttps) {
+            return URLUtil.isValidUrl(url.toString()) && url.toString().startsWith("https");
+        } else {
+            return URLUtil.isValidUrl(url.toString());
+        }
+    }
+
+    protected void cancelNotification(Bundle extras) {
```
patedit: Consider moving this method to `MixpanelFCMMessagingService` were notifications are shown and canceled.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
aecast: I think this behavior makes the most sense

PR #641, opened 2019-10-03
add webview support

aecast:
Allow notifications to open a webview activity from the app instead of switching to a browser.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
@@ -188,8 +188,8 @@ public class MixpanelPushNotification {
         final PendingIntent contentIntent = PendingIntent.getActivity(
                 context,
                 0,
-                data.intent,
-                PendingIntent.FLAG_UPDATE_CURRENT
+                getRoutingIntent(data.onTap, "notificationClick", "notification"),
+                PendingIntent.FLAG_CANCEL_CURRENT
         );
 
         builder.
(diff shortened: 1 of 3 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelWebViewActivity.java
```diff
@@ -1,6 +1,36 @@
...
 import android.app.Activity;
+import android.os.Bundle;
+import android.webkit.WebChromeClient;
+import android.webkit.WebSettings;
+import android.webkit.WebView;
+import android.webkit.WebViewClient;
+
+import com.mixpanel.android.R;
 
 public class MixpanelWebViewActivity extends Activity {
```
jbwyme: why did this class already exist? what was it used for?

aecast: It didn't, I think the class creation is either in another PR or wasn't tracked

jbwyme: ah, it should be in this one though shouldn't it? along with whatever code in the router that calls it?

aecast: yeah found it on other PR going to fix up

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -31,6 +31,7 @@
 
 public class MixpanelPushNotification {
     protected final String LOGTAG = "MixpanelAPI.MixpanelPushNotification";
+    protected final int ROUTING_REQUEST_CODE = 1;
```
jbwyme: add comment about what this does and why we chose 1

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -188,8 +189,8 @@ protected void buildNotificationFromData() {
         final PendingIntent contentIntent = PendingIntent.getActivity(
                 context,
                 0,
-                data.intent,
-                PendingIntent.FLAG_UPDATE_CURRENT
+                getRoutingIntent(data.onTap, "notificationClick", "notification"),
```
jbwyme: as mentioned in another PR, I think we should just be passing the `target` (as an enum) to `getRoutingIntent` and when the target is `notification`, we don't need an id or label.

PR #642, opened 2019-10-04
Notification onClick tracking

aecast:
Restructured notification building process to use routing activity and the new PushAction type, allowing us to track the event where a user taps the notification. Added handling of this event in the MixpanelNotificationRouteActivity.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
@@ -19,7 +19,7 @@ public class MixpanelNotificationRouteActivity extends Activity {
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        Intent routeIntent = getIntent();
+        final Intent routeIntent = getIntent();
         Bundle extras = routeIntent.getExtras();
 
         if (null == routeIntent) {
@@ -29,14 +29,31 @@ public class MixpanelNotificationRouteActivity extends Activity {
 
         trackAction(routeIntent);
 
-        Intent notificationIntent = new Intent(Intent.ACTION_VIEW, Uri.parse(routeIntent.getExtras().getCharSequence("uri").toString()));
+        final Intent notificationIntent = handleRouteIntent(routeIntent);
+
+
         if (!extras.getBoolean("sticky")) {
             cancelNotification(extras);
         }
         startActivity(notificationIntent);
     }
 
-    private void cancelNotification(Bundle extras) {
+    protected Intent handleRouteIntent(Intent routeIntent) {
+        String actionType = routeIntent.getExtras().getCharSequence("actionType").toString();
+
+        if (actionType.equals("browser") || actionType.equals("deeplink")) {
+            return new Intent(Intent.ACTION_VIEW, Uri.parse(routeIntent.getExtras().getCharSequence("uri").toString()));
+        } else if (actionType.equals("webview")) {
+            return new Intent(this.getApplicationContext(), MixpanelWebViewActivity.class).
+                    putExtra("uri", routeIntent.getExtras().getCharSequence("uri").toString()).
+                    addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+        } else {
+            //use homescreen as default if no actionType specified
+            return this.getPackageManager().getLaunchIntentForPackage(this.getPackageName());
+        }
+    }
+
+    protected void cancelNotification(Bundle extras) {
         int notificationId = extras.getInt("notificationId");
         NotificationManager notificationManager = (NotificationManager) getApplicationContext().getSystemService(getApplicationContext().NOTIFICATION_SERVICE);
         notificationManager.cancel(notificationId);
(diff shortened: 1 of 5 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -29,14 +29,31 @@ protected void onCreate(Bundle savedInstanceState) {
 
         trackAction(routeIntent);
 
-        Intent notificationIntent = new Intent(Intent.ACTION_VIEW, Uri.parse(routeIntent.getExtras().getCharSequence("uri").toString()));
+        final Intent notificationIntent = handleRouteIntent(routeIntent);
+
+
```
jbwyme: remove empty line here, don't need two blank lines

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -29,14 +29,31 @@ protected void onCreate(Bundle savedInstanceState) {
...
             cancelNotification(extras);
         }
         startActivity(notificationIntent);
     }
 
-    private void cancelNotification(Bundle extras) {
+    protected Intent handleRouteIntent(Intent routeIntent) {
+        String actionType = routeIntent.getExtras().getCharSequence("actionType").toString();
+
+        if (actionType.equals("browser") || actionType.equals("deeplink")) {
```
jbwyme: let's use an enum or at least constants for these action type values and use a switch/case instead of if/thens

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -29,14 +29,31 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
         if (!extras.getBoolean("sticky")) {
             cancelNotification(extras);
         }
         startActivity(notificationIntent);
     }
 
-    private void cancelNotification(Bundle extras) {
+    protected Intent handleRouteIntent(Intent routeIntent) {
+        String actionType = routeIntent.getExtras().getCharSequence("actionType").toString();
```
jbwyme: can this throw an NPE if `routeIntent.getExtras().getCharSequence("actionType")` is null?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -29,14 +29,31 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+        if (actionType.equals("browser") || actionType.equals("deeplink")) {
+            return new Intent(Intent.ACTION_VIEW, Uri.parse(routeIntent.getExtras().getCharSequence("uri").toString()));
```
jbwyme: same here, can this throw NPE?

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -47,47 +64,60 @@ protected void trackAction(Intent routeIntent) {
...
+            MPLog.i(LOGTAG, "Notification action click logged with no actionId");
             return;
         }
 
-        CharSequence uriChars = intentExtras.getCharSequence("uri");
-        if (null == uriChars) {
-            MPLog.i(LOGTAG, "Notification action click logged with no uri.");
+        CharSequence actionTypeChars = intentExtras.getCharSequence("actionType");
+        if (null == actionIdChars) {
+            MPLog.i(LOGTAG, "Notification action click logged with no actionType");
```
jbwyme: let's use `MPLog.d` for all of these

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -47,47 +64,60 @@ protected void trackAction(Intent routeIntent) {
...
+        final String label = labelChars.toString();
         final String messageId = messageIdChars.toString();
         final String campaignId = campaignIdChars.toString();
 
         MixpanelAPI.allInstances(new MixpanelAPI.InstanceProcessor() {
             @Override
             public void process(MixpanelAPI api) {
                 JSONObject pushProps = new JSONObject();
-
+                String tapTarget;
```
jbwyme: let's have an enum for `TapTarget` with `NOTIFICATION` and `ACTION_BUTTON` as entries.

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -47,47 +64,60 @@ protected void trackAction(Intent routeIntent) {
...
                 try {
-                    pushProps.put("actionId", actionId);
-                    pushProps.put("uri", uri);
-                    pushProps.put("messageId", messageId);
-                    pushProps.put("campaignId", campaignId);
+                    if (actionId.equals("notificationClick")) {
```
jbwyme: why don't we just set the tapTarget on the intent instead of determining it implicitly? then `actionId` can be null when the target is `notification`

src/main/java/com/mixpanel/android/mpmetrics/MixpanelNotificationRouteActivity.java
```diff
@@ -47,47 +64,60 @@ protected void trackAction(Intent routeIntent) {
...
+                        pushProps.put("tap_target", "notification");
+                    } else {
+                        pushProps.put("tap_target", "button");
+                        pushProps.put("button_id", actionId);
```
jbwyme: let's just use `buttonId` instead of `actionId` everywhere for consistency

src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -301,28 +254,99 @@ protected void maybeAddActionButtons() {
...
+        return (new Notification.Action.Builder(icon, title, createActionIntent(onTap, actionId, title))).build();
     }
 
-    protected PendingIntent createActionIntent(PushTapAction onTap, String actionId) {
+    protected PendingIntent createActionIntent(PushTapAction onTap, String actionId, CharSequence label) {
+        Intent routingIntent = getRoutingIntent(onTap, actionId, label);
+        return PendingIntent.getActivity(context, 0, routingIntent, PendingIntent.FLAG_CANCEL_CURRENT);
+    }
+
+    protected Intent getRoutingIntent(PushTapAction onTap, String actionId, CharSequence label) {
```
jbwyme: yeah let's make this signature: `protected Intent getRoutingIntent(PushTapTarget target, PushTapAction onTap, String buttonId = null, CharSequence label = null)`

aecast: if target is already a member of onTap then what is the point in putting it in the signature?

PR #643, opened 2019-10-08
Add android total sessions length

Ivansap:
Adding user fields like on [iOS SDK](https://github.com/mixpanel/mixpanel-iphone/blob/5be3385af99f3c4cd4c1971b834e45c41576f48e/Mixpanel/AutomaticEvents.m#L89) 

TOTAL_SESSIONS - Total user sections counter - 
TOTAL_SESSIONS_LENGTH - total user sections length in seconds

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelActivityLifecycleCallbacks.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelActivityLifecycleCallbacks.java
@@ -71,8 +71,10 @@ import java.util.Locale;
                             String sessionLengthString = nf.format((System.currentTimeMillis() - sStartSessionTime) / 1000);
                             JSONObject sessionProperties = new JSONObject();
                             sessionProperties.put(AutomaticEvents.SESSION_LENGTH, sessionLengthString);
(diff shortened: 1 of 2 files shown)


Ivansap:
hi, @aecast can you review this feature?

Ivansap:
Hi, @jbwyme, can you merge this commit?

jbwyme:
Thoughts on this @patedit? Seems generally ok to me. The two concerns I have at first glance are:
1) no tests
2) if customers are ok with two new automatic profile properties

src/main/java/com/mixpanel/android/mpmetrics/MixpanelActivityLifecycleCallbacks.java
```diff
@@ -71,6 +71,10 @@ public void run() {
                             String sessionLengthString = nf.format((System.currentTimeMillis() - sStartSessionTime) / 1000);
                             JSONObject sessionProperties = new JSONObject();
                             sessionProperties.put(AutomaticEvents.SESSION_LENGTH, sessionLengthString);
+                            if (mMpInstance.getPeople().isIdentified()) {
```
patedit: we don't need to check this. if the user is not identified, these updates will be queued in disk and flushed when `identify()` is called :)

Issue #567, opened 2018-09-24
Crash on formatting a timezone


arturdryomov:
* Device: BlackBerry BBB100-1
* OS: 8.1.0
* Mixpanel SDK: 5.4.1

```
Fatal Exception: java.lang.AssertionError: No NameTypeIndex match for SHORT_DAYLIGHT
       at android.icu.impl.TimeZoneNamesImpl$ZNames.getNameTypeIndex(TimeZoneNamesImpl.java:724)
       at android.icu.impl.TimeZoneNamesImpl$ZNames.getName(TimeZoneNamesImpl.java:790)
       at android.icu.impl.TimeZoneNamesImpl.getTimeZoneDisplayName(TimeZoneNamesImpl.java:183)
       at android.icu.text.TimeZoneNames.getDisplayName(TimeZoneNames.java:261)
       at java.util.TimeZone.getDisplayName(TimeZone.java:405)
       at java.util.Date.toString(Date.java:1066)
       at org.json.JSONStringer.value(JSONStringer.java:252)
       at org.json.JSONObject.writeTo(JSONObject.java:723)
       at org.json.JSONStringer.value(JSONStringer.java:237)
       at org.json.JSONObject.writeTo(JSONObject.java:723)
       at org.json.JSONObject.toString(JSONObject.java:692)
       at com.mixpanel.android.mpmetrics.AnalyticsMessages$PeopleDescription.toString(AnalyticsMessages.java:204)
       at com.mixpanel.android.mpmetrics.AnalyticsMessages$Worker$AnalyticsMessageHandler.handleMessage(AnalyticsMessages.java:317)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loop(Looper.java:164)
       at android.os.HandlerThread.run(HandlerThread.java:65)
```

dmitry-novikov:
Hey, MixPanel team, any change this bug to be fixed?

dmitry-novikov:
Can you please apply the same workaround like Facebook did?
https://stackoverflow.com/questions/51453091/android-facebook-sdk-internal-utility-java-lang-assertionerror/52035884#52035884

dmitry-novikov:
@patedit :question:

patedit:
Hey @dmitry-novikov - I will make sure to add this on the next release.

patedit:
I've created a PR here if you want to take a look: https://github.com/mixpanel/mixpanel-android/issues/567

[...]

PR #652, opened 2019-11-20
Fix AssertionError for some devices running Android 8.X

patedit:
Fixes https://github.com/mixpanel/mixpanel-android/issues/567

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
@@ -82,20 +82,31 @@ import javax.net.ssl.SSLSocketFactory;
 
     // Must be thread safe.
     public void peopleMessage(final PeopleDescription peopleDescription) {
-        final Message m = Message.obtain();
-        m.what = ENQUEUE_PEOPLE;
-        m.obj = peopleDescription;
+        try {
+            peopleDescription.getMessage().toString();
 
-        mWorker.runMessage(m);
+            final Message m = Message.obtain();
+            m.what = ENQUEUE_PEOPLE;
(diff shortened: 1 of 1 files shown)


src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -82,20 +82,31 @@ public void eventsMessage(final EventDescription eventDescription) {
 
     // Must be thread safe.
     public void peopleMessage(final PeopleDescription peopleDescription) {
-        final Message m = Message.obtain();
-        m.what = ENQUEUE_PEOPLE;
-        m.obj = peopleDescription;
+        try {
+            peopleDescription.getMessage().toString();
```
dmitry-novikov: Shouldn't it be already normalized in `MixpanelMessageDescription` constructor?

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -82,20 +82,31 @@ public void eventsMessage(final EventDescription eventDescription) {
...
 
     // Must be thread safe.
     public void groupMessage(final GroupDescription groupDescription) {
-        final Message m = Message.obtain();
-        m.what = ENQUEUE_GROUP;
-        m.obj = groupDescription;
-
-        mWorker.runMessage(m);
+        try {
+            groupDescription.getMessage().toString();
```
dmitry-novikov: The same, check in `MixpanelMessageDescription` constructor should already handle it.

patedit:
thanks @dmitry-novikov - totally missed removing that (was part of a different fix I had in place). Much appreciated you taking a look

PR #656, opened 2019-11-28
Add IP flag to engage call

codeflows:
I noticed that setting `<meta-data android:name="com.mixpanel.android.MPConfig.UseIpAddressForGeolocation" android:value="false" />` prevents IP geolocation data from being added to events, but it's still added to user profile data.

In the Mixpanel iOS library, setting `useIPAddressForGeoLocation = NO` prevents IP geolocation data from being added to events AND user profile. It seems that the `ip` parameter [is added to all outgoing network calls](https://github.com/mixpanel/mixpanel-iphone/blob/5ddcc639620f296023bed91b11a767ccbda9f3f7/Mixpanel/MPNetwork.m#L112).

This PR adds the IP flag to the `engage` call on Android as well which seems to prevent geolocation data from being added to new user profiles in my tests.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MPConfig.java
@@ -379,7 +379,7 @@ public class MPConfig {
     }
 
     public void setMixpanelPeopleEndpoint() {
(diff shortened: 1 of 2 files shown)


patedit:
Thanks @codeflows for the PR. Looks great to me - merging it now!

PR #657, opened 2019-12-02
Remove notification webview action support

aecast:
Removes the ability to launch web-views with an OnTap action from an Android notification.


src/main/AndroidManifest.xml
```diff
@@ -76,11 +76,6 @@
         -->
 
         <!-- This activity will alow your application to open URLs from Notifications in a Webview instead of a browser -->
```
patedit: remove this comment too plz

PR #674, opened 2020-03-03
add mipmap type to checks for resources

aecast:
Adds mipmap types to ReasourceReader so that if a check for an icon resource in Drawables fails when building a push notification, Mipmap resources are checked for that resource name before setting the icon to the default.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
@@ -206,6 +206,7 @@ public class MixpanelPushNotification {
                 notificationIcon = resource.idFromName(iconName);
             } else if ((resource = getResourceIds(ResourceReader.MIPMAP_TYPE, mContext)).knownIdName(iconName)) {
                 notificationIcon = resource.idFromName(iconName);
+                resource = mDrawableIds;
             }
         }
         if (notificationIcon == MixpanelNotificationData.NOT_SET) {
@@ -213,7 +214,6 @@ public class MixpanelPushNotification {
         }
         mData.setIcon(notificationIcon);
 
-        resource = mDrawableIds;
         int whiteNotificationIcon = MixpanelNotificationData.NOT_SET;
         if (whiteIconName != null) {
             if (mDrawableIds.knownIdName(whiteIconName)) {
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
@@ -687,6 +687,8 @@ public class MixpanelPushNotification {
         }
 
         switch (resourceType) {
+            case ResourceReader.ID_TYPE:
+                return new ResourceReader.Ids(resourcePackage, context);
             case ResourceReader.MIPMAP_TYPE:
                 return new ResourceReader.Mipmap(resourcePackage, context);
             case ResourceReader.DRAWABLE_TYPE:
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
@@ -63,7 +63,7 @@ public class MixpanelPushNotification {
     public MixpanelPushNotification(Context context, Notification.Builder builder, long now) {
         this.mContext = context;
         this.mBuilder = builder;
-        this.mDrawableIds = getResourceIds(context);
+        this.mDrawableIds = getResourceIds(ResourceReader.DRAWABLE_TYPE, context);
         this.mNow = now;
         this.ROUTING_REQUEST_CODE = (int) now;
         this.notificationId = (int) now;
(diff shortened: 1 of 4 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
```diff
@@ -199,9 +199,14 @@ public MixpanelPushNotification(Context context, Notification.Builder builder, l
         mData.setSticky(sticky);
 
         int notificationIcon = MixpanelNotificationData.NOT_SET;
+
+        ResourceIds resource = mDrawableIds;
         if (iconName != null) {
             if (mDrawableIds.knownIdName(iconName)) {
-                notificationIcon = mDrawableIds.idFromName(iconName);
+                notificationIcon = resource.idFromName(iconName);
+            } else if ((resource = getResourceIds(ResourceReader.MIPMAP_TYPE, mContext)).knownIdName(iconName)) {
```
jbwyme: so do you think it'd make sense to do this for all ResourceReader.*? I'm genuinely asking because I don't know what developers would expect here.

aecast: No, I don't think so. Other types include layout, color, string, navigation, animation, and animator so when creating an icon resource I think it makes sense for it to be covered by these two types. I think a large majority will be in drawable and experience no issues and the rest probably still use mipmap from older iterations of their code.

patedit: It makes sense to do it in ResourceReader. My proposal is to have something along these lines:

```
this.mDrawableIds = getResourceIds(ResourceReader.DRAWABLE_TYPE, context);
this.mMipmapIds = getResourceIds(ResourceReader.MIPMAP_TYPE, context);

...

if (mDrawableIds.knownIdName(iconName)) {
  //
} else if (mMipmapIds.knownIdName(iconName)) {
   //
}
```

Ideally, `ResourceReader` should be refactored so [getLocalClassName](https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/ResourceReader.java#L113) can return a list of classes (`getLocalClassNames`) and we don't have to check for both drawables and mipmaps everytime we want to find an image asset.

For example, you forgot to check for a mipmap image [in this function](https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java#L287) and the above solution would have covered that.

src/main/java/com/mixpanel/android/mpmetrics/ResourceReader.java
```diff
@@ -16,10 +19,9 @@
  */
 public abstract class ResourceReader implements ResourceIds {
 
-    public static class Ids extends ResourceReader {
-        public Ids(String resourcePackageName, Context context) {
-            super(context);
-            mResourcePackageName = resourcePackageName;
+    public static class Id extends ResourceReader {
+        public Id(String resourcePackageName, Context context) {
+            super(resourcePackageName, context);
```
patedit: Since you are doing this refactoring, it might make sense to do something like:

```
public interface ResourceIds {
    public void initialize();
    ...
}
```

and have the main class call that method that `Id` and `Drawable` need to override. So:

```
protected ResourceReader(String resourcePackageName, Context context) {
    mResourcePackageName = resourcePackageName;
    mContext = context;	        mContext = context;
    mIdNameToId = new HashMap<String, Integer>();
    mIdToIdName = new SparseArray<String>()
    initialize();
}
```

so we can remove these constructors

src/main/java/com/mixpanel/android/mpmetrics/ResourceReader.java
```diff
@@ -145,10 +175,15 @@ protected void initialize() {
         }
     }
 
+    private final String mResourcePackageName;
```
patedit: change this to `protected` and avoid using `super.mResourcePackageName`

src/main/java/com/mixpanel/android/mpmetrics/ResourceReader.java
```diff
@@ -109,6 +125,16 @@ private static void readClassIds(Class<?> platformIdClass, String namespace, Map
         }
     }
 
+    protected List<String> getLocalClassNames() {
```
patedit: this defeats the purpose of having subclasses. definitely not recommended.

src/main/java/com/mixpanel/android/mpmetrics/ResourceReader.java
```diff
@@ -109,6 +125,16 @@ private static void readClassIds(Class<?> platformIdClass, String namespace, Map
...
+        ArrayList<String> localClassNames = new ArrayList<>();
+        for (Class c : ResourceReader.class.getDeclaredClasses()) {
+            localClassNames.add(mResourcePackageName + ".R$" + c.getSimpleName().toLowerCase());
```
patedit: yikes, adding this hidden dependency here where the name of the class needs to match the name of the resource is very dangerous

src/main/java/com/mixpanel/android/mpmetrics/ResourceReader.java
aecast: I misunderstood what you initially meant. I was pretty confused as to why we would need subclasses but now I see I just horribly misinterpreted what you meant. Fixing now

src/main/java/com/mixpanel/android/mpmetrics/ResourceReader.java
aecast: If we remove these constructors, won't these subclasses look for the default constructor in ResourceReader which is undefined since we wrote an explicit constructor so the implicit default constructor isn't generated? If we explicitly define the default constructor, how will we capture the resourcePackageName and context args? I don't see a way to remove these constructors with the desired effect

patedit: you're right, there is no default constructor. a simple thing is to just call super
```
public Ids(String name, Context ctx) {
    super(name,ctx);
}
```

aecast: Do we still want to make initialize part of the interface and have sub classes of ResourceReader implement them given that the functionality is the same and we can't remove these constructors anyways since we need to call super?

PR #682, opened 2020-03-16
track using application context

aecast:
Use application context instead of activity context so that when checking isAppInForeground we are doing so on the main app context, not the current activity which will always be in the foreground while running this section of code.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelPushNotification.java
@@ -624,7 +624,7 @@ public class MixpanelPushNotification {
                     new JSONObject()
             );
 
-            MixpanelAPI instance = MixpanelAPI.getInstanceFromMpPayload(mContext, mpPayloadStr);
+            MixpanelAPI instance = MixpanelAPI.getInstanceFromMpPayload(mContext.getApplicationContext(), mpPayloadStr);
             if (instance != null && instance.isAppInForeground()) {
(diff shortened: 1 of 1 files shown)


aecast:
> were you able to confirm this tracks appropriately when the app is actually in the foreground?

Screen Shot 2020-03-16 at 12 18 22 PM

App closed (7min) vs App open (5min)

patedit:
we already [always pass](https://github.com/mixpanel/mixpanel-android/blob/b9a9e5694761984ac37bc32e59137319bf31fc7b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java#L149) the application context to `MixpanelPushNotification`

jbwyme:
> we already [always pass](https://github.com/mixpanel/mixpanel-android/blob/b9a9e5694761984ac37bc32e59137319bf31fc7b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java#L149) the application context to `MixpanelPushNotification`

@aecast can you explain this? it appears this change doesn't functionally change anything.

aecast:
> > we already [always pass](https://github.com/mixpanel/mixpanel-android/blob/b9a9e5694761984ac37bc32e59137319bf31fc7b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelFCMMessagingService.java#L149) the application context to `MixpanelPushNotification`
> 
> @aecast can you explain this? it appears this change doesn't functionally change anything.

debugging now, this change does correct the event behavior but probably for a reason I don't understand then

PR #686, opened 2020-03-18
Messaged received tracking fix

aecast:
State of MixpanelActivityLifecycleCallbacks does not persist when the app is in killed in Android causing isForeground to be reinitialized to true again when a message is received. Init isForeground to false instead.


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1952,6 +1952,26 @@ public boolean isAppInForeground() {
...
+    /**
+     * Based on the application's event lifecycle this method will determine whether the app
+     * is paused or not.
+     *
+     * If your build version is below 14 this method will always return false.
+     *
+     * @return True if the app is in paused state.
+     */
+    public boolean isAppPaused() {
+        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
```
patedit: we can safely remove this check since the creator of the lifecycle checks for the API version already and `mMixpanelActivityLifecycleCallbacks` will be null otherwise. we don't need to check everytime we access this var.

PR #687, opened 2020-03-23
add browsable category to push tap actions with browser type

aecast:
Adds browsable category to push tap action links with browser type so that they always open in browsers instead of behaving the same as deep links.


jbwyme:
This makes sense to me, let's get final approval from @patedit

PR #697, opened 2020-05-10
Use SecureRandom instead of Random

mohalaci:
Please use SecureRandom instead of Random because security analyzer tools mark mixpanel as an un-secure library.

### DIFF
--- a/acceptance/test-application/src/com/mixpanel/example/hello/MainActivity.java
+++ b/acceptance/test-application/src/com/mixpanel/example/hello/MainActivity.java
@@ -25,7 +25,6 @@ import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.util.Calendar;
 import java.util.Date;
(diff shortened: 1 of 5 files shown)


acceptance/test-application/src/com/mixpanel/example/hello/MainActivity.java
```diff
@@ -287,7 +286,7 @@ private String getTrackingDistinctId() {
     // in your users table to store mixpanel distinct_id, so it is easily
     // accesible for use in attributing cross platform or server side events.
     private String generateDistinctId() {
-        final Random random = new Random();
+        final SecureRandom random = new SecureRandom();
```
patedit, line 289: missing import

mohalaci: Ahh, I forgot the test app, thanks!

PR #723, opened 2021-02-18
Change Android mixpanel default properties ($app_release, $app_build_number) from number to string

tylerchaemp:
iOS's $app_build_number is string with XX.XX.XX format, and Android sdk sends number.

Although Mixpanel is designed to handle mixed types, unfortunately it's not free. our query engine needs to slow down when it encounters different types. especially for Mixpanel default properties, this impacts most of our clients. this PR matches the types of two default properties with iOS expecting overall performance improvements for properties query.


zihejia (approved):
LGTM

Issue #711, opened 2020-10-21
In-app messages cause android app crash "Software rendering doesn't support hardware bitmaps"


kauz-mobilepay:
Hi,

Mixpanel in-app messages are causing "Software rendering doesn't support hardware bitmaps" crashes when capturing screens that contain hardware accelerated images: 
```
com.mixpanel.android.util.ActivityImageUtils.getScaledScreenshot (ActivityImageUtils.java:15)
com.mixpanel.android.util.ActivityImageUtils.getHighlightColorFromBackground (ActivityImageUtils.java:43)
com.mixpanel.android.mpmetrics.MixpanelAPI$PeopleImpl$2.run (MixpanelAPI.java:2509)
android.app.Activity.runOnUiThread (Activity.java:6922)
com.mixpanel.android.mpmetrics.MixpanelAPI$PeopleImpl.showGivenOrAvailableNotification (MixpanelAPI.java:2482)

```

PixelCopy could be used for solving this.

clhols:
We are still seeing this crash our app in Mixpanel 5.8.5.

zihejia:
hi @kauz-mobilepay @clhols , we have released v5.8.7 which has a fix for this issue. I'm closing this one now, please feel free to reopen it if the problem still exists.

gosr:
Hi @zihejia, this crash is also occurring in v5.9.1:

> Fatal Exception: java.lang.IllegalArgumentException: Software rendering doesn't support hardware bitmaps
       at android.graphics.BaseCanvas.onHwBitmapInSwMode(BaseCanvas.java:673)
       at android.graphics.BaseCanvas.throwIfHwBitmapInSwMode(BaseCanvas.java:680)
       at android.graphics.BaseCanvas.throwIfCannotDraw(BaseCanvas.java:84)
       at android.graphics.BaseCanvas.drawBitmap(BaseCanvas.java:142)
       at android.graphics.Canvas.drawBitmap(Canvas.java:1648)
       at android.graphics.drawable.BitmapDrawable.draw(BitmapDrawable.java:548)
       at android.widget.ImageView.onDraw(ImageView.java:1453)
       at android.view.View.draw(View.java:23901)
       at android.view.View.draw(View.java:23773)
       at android.view.ViewGroup.drawChild(ViewGroup.java:5336)
       at android.view.ViewGroup.dispatchDraw(ViewGroup.java:5093)
[...]

PR #724, opened 2021-03-09
Replace deprecated view drawing cache api

zihejia:
Replace deprecated  "buildDrawingCache" API, this will lead to "Software rendering doesn't support hardware bitmaps" crashes.
This is to address https://github.com/mixpanel/mixpanel-android/issues/711


jaredmixpanel (approved):
LGTM

Issue #728, opened 2021-03-23
ActivityImageUtils.getScaledScreenshot crash


matteinn:
Hi!
I'm noticing some crashes on the 5.8.7 android sdk, mainly on Android 8 devices.

`Fatal Exception: java.lang.IllegalArgumentException
width and height must be > 0

com.mixpanel.android.util.ActivityImageUtils.getScaledScreenshot (ActivityImageUtils.java:16)`

Possibly related to https://github.com/mixpanel/mixpanel-android/issues/711

zihejia:
hi @matteinn , sorry for the inconvenience and thanks for providing the PR. I will do a patch release for this.

PR #729, opened 2021-03-24
Crash prevention on getScaledScreenshot when rootView is not measured yet

matteinn:
Fixes https://github.com/mixpanel/mixpanel-android/issues/728

Updated `ActivityImageUtils.getScaledScreenshot(...)` method like this:
- Added a check on `rootView` width and height attributes to perform an early exit in case that view is not measured yet
- Replaced the "// May return null." comment with a proper @Nullable annotation so that the compiler will warn about improper nullability handling
- Added a more explicit comment with reasons why this method may return `null`


zihejia (approved):
Thank you for providing this PR.

Issue #726, opened 2021-03-11
Allow sending event properties with null value


carstenhag:
Hello there, we have found a discrepancy between the mixpanel-ios and mixpanel-android libraries.

My iOS colleague log an event like this, in case a request is successful: `["evse_id": evSupplyEquipmentID, "success": true, "error": NSNull()]`. On the mixpanel website it shows up like `error:   <null>` which we want.

On AND I want to log the same thing, but the library has an additional check that doesn't allow `"error": null` to be sent. I think it's this line: https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java#L2886. On the mixpanel website, on Android it says `error: undefined`. So if we want to analyze stuff, we need to check for both null and undefined which is a bit annoying.

The code that we are using:

```kt
    private val gson = GsonBuilder()
            .serializeNulls() // Makes sure "error": null are also output
            .enableComplexMapKeySerialization()
            .create()

   fun propertiesAsJSONObject(
            properties: Map<AnalyticsPropertiesKey, Any?>?
    ): JSONObject? {
        return try {
            properties?.let { JSONObject(gson.toJson(it)) }
        } catch (exception: Exception) {
            Timber.e(exception, "Error while serializing AnalyticsProperties and Keys.")
            null
        }
    }

    // main function that actually calls mixpanel?.track()
    override fun trackEvent(event: AnalyticsEvent) {
        val serializedProperties = propertiesAsJSONObject(event.properties)

        mixpanel?.track(event.key, serializedProperties)
    }

    // JSONObject passed to MixpanelAPI.track():
    {"success":true,"evse_id":"123","error":null}
```

It would be great if there was an option to allow sending nulls (if the key is defined). Maybe it could also be made the default. Thanks in advance!

carstenhag:
I have deleted the `if (!properties.isNull(key)) {` line and now it seems to work. Of course I can't actually do this, so I can't use this as a workaround :/.

```kt
    @Test
    fun modifiedMixpanelPropertiesIteration() {
        val gson = GsonBuilder().serializeNulls().enableComplexMapKeySerialization().create()
        val properties = AnalyticsEvent.TrackRemoteStartResult("123", AnalyticsEvent.ChargingResult.Success).properties
        val propertiesJsonObject = properties?.let { JSONObject(gson.toJson(it)) }!!

        val messageProps = JSONObject()


[...]

PR #736, opened 2021-04-22
Enable tracking null property value

zihejia:
addressing the issue: https://github.com/mixpanel/mixpanel-android/issues/726


jaredmixpanel (approved):
lgtm

Issue #719, opened 2020-12-21
Fatal Exception: java.lang.OutOfMemoryError 


lorenc-tomasz:
Fatal Exception: java.lang.OutOfMemoryError: Failed to allocate a 2312 byte allocation with 2284928 free bytes and 2231KB until OOM, target footprint 268435456, growth limit 268435456; failed due to fragmentation (largest possible contiguous allocation 49020928 bytes)
       at java.util.Arrays.copyOf(Arrays.java:3257)
       at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:124)
       at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:649)
       at java.lang.StringBuilder.append(StringBuilder.java:203)
       at org.json.JSONStringer.string(JSONStringer.java:354)
       at org.json.JSONStringer.value(JSONStringer.java:261)
       at org.json.JSONObject.writeTo(JSONObject.java:733)
       at org.json.JSONStringer.value(JSONStringer.java:246)
       at org.json.JSONObject.writeTo(JSONObject.java:733)
       at org.json.JSONObject.toString(JSONObject.java:701)
       at com.mixpanel.android.mpmetrics.MPDbAdapter.addJSON(MPDbAdapter.java:322)
       at com.mixpanel.android.mpmetrics.AnalyticsMessages$Worker$AnalyticsMessageHandler.handleMessage(AnalyticsMessages.java:429)
       at android.os.Handler.dispatchMessage(Handler.java:107)
       at android.os.Looper.loop(Looper.java:237)
       at android.os.HandlerThread.run(HandlerThread.java:67)

zihejia:
fixed in v5.9.0. Thanks for bringing it up. Please feel free to reopen if the problem still exists.

PR #737, opened 2021-04-22
Prevent the out of memory crash when adding event to the SQLiteDatabase

zihejia:
addressing https://github.com/mixpanel/mixpanel-android/issues/719


jaredmixpanel (approved):
lgtm

PR #739, opened 2021-04-28
add setUseIpAddressForGeolocation, setEnableLogging, setServerURL

zihejia:
Provide new apis, in parity with Mixpanel iOS SDK.
setUseIpAddressForGeolocation, setEnableLogging and setServerURL. 

Previously, those settings are only available to be set in `AndrodManifest.xml`, not possible to be changed in run time and consumed by react-native/flutter SDK.


jaredmixpanel (approved):
lgtm!

Issue #627, opened 2019-09-18
"First app open" event 


maestr0ua:
Hello!
Despite the fact that I don't have any special settings in AndroidManifest file, I still get "First App Open" event with random distinctID.
The first question is why I receive that events, even with default value.
```
mDisableAppOpenEvent = metaData.getBoolean("com.mixpanel.android.MPConfig.DisableAppOpenEvent", true);
```
And the second question is why I get that events with random distinctID, despite on that I set my own distinctID at start the application.
```
mixpanelAPI = MixpanelAPI.getInstance(context,context.getString(R.string.mix_panel_token), false);
mixpanelAPI.reset();
mixpanelAPI.getPeople().identify(userId);
mixpanelAPI.identify(userId);
```

I need that events will come with my distinctID or will not come at all. Both variants are appropriate.

Thank you
<img width="1329" alt="Screen Shot 2019-09-12 at 3 11 36 PM-2" src="https://user-images.githubusercontent.com/11719504/65132579-d8e5a400-da09-11e9-96bf-c9b9a5507f7d.png">

maestr0ua:
Could anybody help?

patedit:
Hi @maestr0ua ! That "First App Open" event is one of our default mobile events. If you want to disable this feature, please read [this article](https://help.mixpanel.com/hc/en-us/articles/115004596186-Default-Mobile-Events-Collection).

OlivierFreyssinet:
Hi, sorry to reopen this issue but the linked article's solution is not very satisfying because it's not flexible at all.

Is there really no way to just disable this particular event and not other default mobile events ?

As the OP said, it would be great if at least there was a way to set the `distinctId` **before** this event is sent, because having one event with an ID that will never reappear kind of messes up the analytics ...
For instance if there was a way to get MixpanelAPI instance by specifying the `distinctId` I guess it could solve the issue. I'm saying that because I've seen that in the past, pull requests have been made to specify `superProperties` in `getInstance(...)`. I'm not familiar with the codebase at all so I can't really make that pull request myself for now ...

It also seems strange that the solution to not sending an event _on the client_ is to just filter it out (with other possibly very useful events) in the projects settings ...

This is quite frustrating and if anyone has a solid solution or even a hacky workaround to just force disable this event, I would be super grateful!

OlivierFreyssinet:
[...]

PR #748, opened 2021-06-11
Add param "distinctId" to constructor/getInstance to identify user before any event gets sent

OlivierFreyssinet:
## The issue
As mentioned in https://github.com/mixpanel/mixpanel-android/issues/627, — from my understanding at least — there is currently no way to avoid the initial "First app open" to be sent with a randomly generated user ID, at least not without disabling all default mobile events at all.

This can be quite disrupting as we might still want to set our own user ID (for instance in my case, we migrated our analytics from "Segment forwarding events to Mixpanel" to a direct integration of Mixpanel, so we are keeping the same user ID that was used for the Segment SDK, that allows us to avoid a new distinct user being created, which would mess up our analytics data etc.)

**If we just call `identify(distinctId)` after initializing the Mixpanel instance, then there will still be this "First app open" event being sent with a different ID, which would create a duplicate/ghost user in our data, making it harder to analyze than it should be.**

## The proposed solution
In this PR, I added a parameter `distinctId` to the main constructor and a new getInstance() method that uses this new parameter. This will identify the user prior to any default mobile events being sent.

I also updated the tests to use the new signature of the main `MixpanelAPI` constructor (with a `null` value for the `distinctId` parameter). Right now with these changes all tests are passing.

Disclaimer: I'm not familiar with the internal logic of this SDK and how the instance gets initialized so it's possible that this solution causes bugs or unwanted side-effects, there might also be a more elegant solution or maybe we'd want to make that `identify` call earlier in the `MixpanelAPI` constructor. 

For my company's app it did the job and works fine, we've been using that in prod since last December.

Regarding existing `getInstance` implementations, I made sure that it was a non breaking change for their type signature.

Thank you for your time!


scotmatson:
@zihejia this commit has likely gone stale after sitting for two years. Thoughts on closing this out or merging? Currently this is the only open PR.

zihejia:
closing it cause it's obsolete.

Issue #717, opened 2020-12-11
Mixpanel using legacy support libs in reflection calls - can't drop jetifier due to this


carstenhag:
See https://github.com/dipien/bye-bye-jetifier/discussions/15 and specifically the comment https://github.com/dipien/bye-bye-jetifier/discussions/15#discussioncomment-186714.

Jetifier is used on android to make old code that uses legacy android support libraries work with newer androidx libraries. It is really useful at that, but it comes with the disadvantage of increasing build times, see [this article](https://adambennett.dev/2020/08/disabling-jetifier/).

Mixpanel seems to use `android.support.v4.content.LocalBroadcastManager` inside https://github.com/mixpanel/mixpanel-android/blob/master/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java via reflections for something.

You should probably straight up either remove the old one and use androidx, or do like AppCenter and attempt the reflection calls with either the old or the new package/class. See https://github.com/microsoft/AppCenter-Test-Espresso-Extensions/blob/master/src/main/java/com/microsoft/appcenter/espresso/Factory.java for that.

carstenhag:
I could probably drop jetifier without this change/fix, as it's probably not checked at compile-time whether reflection calls will succeed. But then I am at risk of runtime crashes when Mixpanel calls that code.

PR #759, opened 2021-10-06
remove legacy support libs in reflection calls

zihejia:
address #717


jaredmixpanel (approved):

zihejia:
> @zihejia Should the class name really include *\n"? Seems strange, as it previously didn't.
oops, I will fix it.

PR #775, opened 2021-12-28
Update Apache Licence URL

magnusvs:
Updates licence name and URL, old one returns 404.

Feel free to revert back the name change if you're sure you want the old one.


zihejia (approved):
Thanks @magnusvs !

Issue #740, opened 2021-05-03
How to clear time Event? 


vinson1022:
I try to use `timeEvent()` to record a session from start editing to submit success.
And I want to cancel `timeEvent()` tracking when users exit editing.
But I can't find any API that can solve this situation on SDK. 

P.S. I find API `clearTimeEvent()` on iOS SDK can solve this situation.

ddk9499:
Hello @zihejia. Any news about this issue?

zihejia:
Hi @vinson1022 @ddk9499, we have released [6.1.0](https://github.com/mixpanel/mixpanel-android/releases/tag/v6.1.0) which contains the fix. I'm closing this one now, please feel free to reopen if the problem still exists.

PR #779, opened 2022-03-02
Add clearTimedEvent() and clearTimedEvents()

zihejia:
addresses #740

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-VERSION_NAME=6.0.0
+VERSION_NAME=6.0.1-SNAPSHOT
 
 POM_PACKAGING=aar
 GROUP=com.mixpanel.android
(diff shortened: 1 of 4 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -502,6 +502,29 @@ public void timeEvent(final String eventName) {
...
+        synchronized (mEventTimings) {
+            mEventTimings.clear();
+            mPersistentIdentity.clearTimedEvents();
+        }
+    }
+
+    /**
+     * Clears the event timing for an event.
+     *
+     * @param eventName the name of the event to track with timing.
```
jaredmixpanel, line 519:
```suggestion
     * @param eventName the name of the timed event to clear.
```

gradle.properties
```diff
@@ -1,4 +1,4 @@
-VERSION_NAME=6.0.0
+VERSION_NAME=6.0.1-SNAPSHOT
```
jaredmixpanel, line 1: okay to have `-SNAPSHOT` here in master?

jaredmixpanel (approved):
lgtm w/ one question and one suggestion

gradle.properties
zihejia: yes, it should. I will figure out why the release script did not change it.

PR #782, opened 2022-04-09
Deprecate  automatic property for "Radio($radio)"

zihejia:
Fix the warning: "Permission READ_PHONE_STATE not granted. Property $radio will not be available."

Since we now have `targetSdkVersion` as 30. To get the Radio value(3G, 4G, LTE, etc), the permission `READ_PHONE_STATE` is required which is considered "dangerous" after `targetSdkVersion` 23. Now, for users on Android 6+, runtime permissions are required. Therefore, we no longer provide this property automatically. You can check more details in this [doc](https://developer.android.com/training/permissions/requesting)  from Google. 

If you still need it, you can use methods like `checkSelfPermission()` and `requestPermissions()` to ask users and set the "$radio" property by yourself.


jaredmixpanel (approved):
lgtm

PR #784, opened 2022-05-03
Changes method signature for React Native

sidferreira:
This is needed to allow RN to access the Device ID


zihejia (approved):
Thanks @sidferreira !

Issue #703, opened 2020-08-03
How to identify multiple different users on the same device under the same mixpanel project token?


pikaboo:
I need to identify multiple users that run on the same device from the same mixpanel project token.

How to correctly do that?
In iOS it is possible because the MixpanelAPI has an instance name. In Android it seems not possible to have an instance name

zihejia:
hi @pikaboo , sorry for the late response. You can get the same instance with the same token. http://mixpanel.github.io/mixpanel-android/index.html#getInstance

pikaboo:
@zihejia Please reopen this.

You answer is not what I was looking for.
I do not want the same instance. I want the functionality that iOS version provides.

![image](https://user-images.githubusercontent.com/3405551/115134767-bff50880-a01b-11eb-8c79-174361d3e677.png)

Where i can get an instance of Mixpanel to report on several different "users" living on the same device.

For some reason iOS does have this functionality, and android does not. In android I need to pass the "mixpanel token" each time.

I dont want 
```         
 mMixpanel = MixpanelAPI.getInstance(this, "YOUR MIXPANEL API TOKEN");
```

I want 
```
          mMixpanel = MixpanelAPI.getInstance(this, "My Local user identifier");
```

zihejia:
hi @pikaboo ,  no worries, we will address this.

pikaboo:
@zihejia thanks, I look forward for some news.

zihejia:
hi @pikaboo , please try v6.4.0 which includes the feature you requested, sorry for the long waiting. Feel free to reopen if you find any issues.

PR #792, opened 2022-06-28
Support defining multiple instances by specifying `instanceName` in `getInstance()`

zihejia:
addresses #703. This will add the following APIs to MixpanelAPI:
```
/**
...
 * @param instanceName The name you want to uniquely identify the Mixpanel Instance.
   It is useful when you want more than one Mixpanel instance under the same project token.
...
**/
```
`getInstance(Context context, String token, String instanceName)`
`getInstance(Context context, String token, boolean optOutTrackingDefault, String instanceName)`
`getInstance(Context context, String token, JSONObject superProperties, String instanceName)`
`getInstance(Context context, String token, boolean optOutTrackingDefault, JSONObject superProperties, String instanceName)`

Please note: If you are going to add `instanceName` to `getInstance` on your existing implementation. `getInstance` will start using `instanceName` as the instance identifier rather than `token`, so you might lose some of the stored properties including the distinct Id under the `token`. We'd recommend using it when you need to create more than one instance under the same project token. You won't lose any events and user profile updates.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-VERSION_NAME=6.3.0
+VERSION_NAME=6.4.0-SNAPSHOT
 
 POM_PACKAGING=aar
 GROUP=com.mixpanel.android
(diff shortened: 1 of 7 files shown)


jaredmixpanel (approved):
LGTM

PR #798, opened 2022-09-08
Fix event being dropped if property value is custom object

zihejia:
Mixpanel does not support saving a custom class object as a property. Currently if you happened to save a custom object as a property, it will cause the event being silently dropped. The PR will prevent the event from being dropped if properties contain custom objects, however, the custom object still can't be sent to Mixpanel and will fall back to a null value. If you want to send property values from the custom class object or nested collection type, you need to flatten them and convert  them as [Mixpanel Supported Data Types.]( https://help.mixpanel.com/hc/en-us/articles/115004547063-Properties-Supported-Data-Types#:~:text=Mixpanel%20supports%20five%20data%20types,%2C%20boolean%2C%20date%20and%20list.)


jaredmixpanel (approved):
LGTM

PR #801, opened 2022-10-11
Upgrade SDK version to 31

zihejia:
Bump SDK version to 31 to meet Google Play's target API level requirement: https://developer.android.com/google/play/requirements/target-sdk


jaredmixpanel (approved):
lgtm

Issue #806, opened 2022-11-04
RemoveLegacyResidualFiles causes ANRs


yaroslav-v:
Hi!

In the latest v7.2.0 you've added the ability to remove residual image files left from the legacy SDK versions.

We've opted in and enabled com.mixpanel.android.MPConfig.RemoveLegacyResidualFiles. Since that time numerous ANRs appear in Firebase crashlogs for various devices, mostly on Android 12.

Here is a stack trace:
```
LegacyVersionUtils.java line 17
com.mixpanel.android.util.LegacyVersionUtils.removeLegacyResidualImageFiles

libopenjdk.so
Java_java_io_UnixFileSystem_getBooleanAttributes0 + 96
java.io.UnixFileSystem.getBooleanAttributes0 (UnixFileSystem.java)
java.io.UnixFileSystem.getBooleanAttributes (UnixFileSystem.java:254)
java.io.File.isDirectory (File.java:843)
com.mixpanel.android.util.LegacyVersionUtils.removeLegacyResidualImageFiles (LegacyVersionUtils.java:17)
com.mixpanel.android.util.LegacyVersionUtils.removeLegacyResidualImageFiles (LegacyVersionUtils.java:21)
com.mixpanel.android.util.LegacyVersionUtils.removeLegacyResidualImageFiles (LegacyVersionUtils.java:21)
com.mixpanel.android.util.LegacyVersionUtils.removeLegacyResidualImageFiles (LegacyVersionUtils.java:21)
com.mixpanel.android.mpmetrics.MixpanelAPI.<init> (MixpanelAPI.java:202)
com.mixpanel.android.mpmetrics.MixpanelAPI.<init> (MixpanelAPI.java:123)
com.mixpanel.android.mpmetrics.MixpanelAPI.getInstance (MixpanelAPI.java:505)
com.mixpanel.android.mpmetrics.MixpanelAPI.getInstance (MixpanelAPI.java:278)
...
android.app.Instrumentation.callApplicationOnCreate (Instrumentation.java:1213)
android.app.ActivityThread.handleBindApplication (ActivityThread.java:6884)
android.app.ActivityThread.access$1600 (ActivityThread.java:260)
android.app.ActivityThread$H.handleMessage (ActivityThread.java:2136)
android.os.Handler.dispatchMessage (Handler.java:106)
android.os.Looper.loopOnce (Looper.java:226)
android.os.Looper.loop (Looper.java:329)
[...]

PR #807, opened 2022-11-16
Fix ANRs issue when opt in and remove residual image files left from the legacy SDK versions

zihejia:
This PR should fix the ANRs issue, address #806


jaredmixpanel (approved):
lgtm

PR #808, opened 2023-01-30
replace original with distinct_id in alias

jaredmixpanel:
remove old `original` key and use the documented key `distinct_id`


zihejia (approved):
LGTM

PR #819, opened 2023-09-13
Fix integration event when opted out

argenisf:
Add a check for the integration event to only be sent when the library has not been initialized in the opted out state or the user has already opted out

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -176,7 +176,7 @@ public class MixpanelAPI {
             track("$app_open", null);
         }
 
-        if (!mPersistentIdentity.isFirstIntegration(mToken)) {
+        if (!mPersistentIdentity.isFirstIntegration(mToken) && !optOutTrackingDefault && !hasOptedOutTracking()) {
             try {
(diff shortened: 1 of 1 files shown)


zihejia (approved):
LGTM

carstenhag:
Thanks for making the change. I really wonder why this is not tested though. Are there no integration tests? Things like this are easy to break and to get wrong. 
I expected there to be a good test case around the very basics of opt in/opt out.

jaredmixpanel:
@carstenhag There are opt out tests: https://github.com/mixpanel/mixpanel-android/blob/master/src/androidTest/java/com/mixpanel/android/mpmetrics/OptOutTest.java

They just didn't cover the integration event, which is tracked directly via HTTP API, it doesn't use the tracking queue mechanism. Also worth pointing out that the integration event doesn't track anything about your end user to your project. Effectively it captures data about you (Mixpanel's user) and sends it to an internal Mixpanel project.  Specifically it just tells us that a given Mixpanel project token has been used to initialize an SDK.

carstenhag:
@jaredmixpanel Thanks for your answer. I see tests regarding the general opt out, but not for the integration event.

The argument "the integration event doesn't track anything about your end user to your project" is true. We as the company don't get any info.

But to privacy-aware users, journalists & courts this distinction probably does not matter. If we as the company use Mixpanel and you send data to yourself, without an opt in, it's still wrong.  
I'm happy this was fixed, but as a developer I am still curious about the test situation.

PR #833, opened 2024-03-27
feat: add headers support for mixpanel proxy api calls

abhilashdas-cred:
### Introducing Header Support for Mixpanel Proxy Servers

This pull request introduces a feature enabling proxy servers to request custom headers from the application.

### Problem:

Currently, when the proxy server URL is set, it makes API calls in a standard manner. However, if the application wishes to include certain headers, such as authentication tokens or device IDs, this isn't possible.

### Solution:

This pull request implements a callback mechanism. Before making an API call, the system requests headers and returns the API endpoint and status code. This information can be utilized by the application for tracking or other purposes.

### Implementation:

This implementation introduces a new property in the config class called "MixpanelServerCallback". It contains a method named "getHeaders()" which requests any headers to be passed just before the proxy server API call. Once the API call is complete, it returns the API endpoint and status code via "onResponse()".

### Benefits:

Enhanced Proxy API Control:
This solution grants the proxy API full control during Mixpanel proxy API calls.

### DIFF
--- a/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
+++ b/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
@@ -14,6 +14,7 @@ import android.os.Build;
 import android.os.Bundle;
 
 import com.mixpanel.android.util.MPLog;
+import com.mixpanel.android.util.MixpanelServerCallback;
 
 import org.json.JSONArray;
 import org.json.JSONException;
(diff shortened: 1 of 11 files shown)


src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -580,6 +581,17 @@ public void setServerURL(String serverURL) {
...
 
+    /**
+     * Set the base URL used for Mixpanel API requests.
+     * Useful if you need to proxy Mixpanel requests. Defaults to https://api.mixpanel.com.
+     * To route data to Mixpanel's EU servers, set to https://api-eu.mixpanel.com
+     *
+     * @param serverURL the base URL used for Mixpanel API requests
+     * @param callback the callback for mixpanel proxy server api headers and status
+     */
+    public void setServerURL(String serverURL, MixpanelServerCallback callback) {
```
zihejia, line 592: missing the definition of `MixpanelServerCallback`, are you meant to use `ProxyServerInteractor`?

abhilashdas-cred: yes. updated it

zihejia (approved):
LGTM

Issue #846, opened 2024-09-23
MixpanelAPI.getInstance() crash in instant apps on pre-Android 13 devices


MatthewWhitmer:
There is a crash from MixpanelAPI.getInstance() introduced with 8.0.0.  

The crash occurs with instant apps running on devices pre-Android 13.  The root cause is the androidx backwards compatible approach to some receivers getting wrapped with a custom permission ${packageName}.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION.  Since instant apps are not allowed to create permissions the associated registration (via ContextCompat) throws an exception when checking for that permission.  The permission declaration itself as well as the uses-permission are both properly merged into the manifest from androidx.core, the issue is simply that instant apps can't create permissions.  This is the security measure from Google introduced in Android 13 that is failing in their ContextCompat implementation: https://developer.android.com/about/versions/13/features#runtime-receivers 

Currently the only workaround instant apps running on pre-Android 13 devices is to not upgrade the MixPanel dependency to 8.0.0.

Caused by: java.lang.RuntimeException: Permission [your_instant_app].DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION is required by your application to receive broadcasts, please add it to your manifest
	at androidx.core.content.ContextCompat.obtainAndCheckReceiverPermission(ContextCompat.java:970)
	at androidx.core.content.ContextCompat$Api26Impl.registerReceiver(ContextCompat.java:1117)
	at androidx.core.content.ContextCompat.registerReceiver(ContextCompat.java:822)
	at androidx.core.content.ContextCompat.registerReceiver(ContextCompat.java:769)
	at com.mixpanel.android.mpmetrics.MixpanelAPI.<init>(MixpanelAPI.java:199)
	at com.mixpanel.android.mpmetrics.MixpanelAPI.<init>(MixpanelAPI.java:123)
	at com.mixpanel.android.mpmetrics.MixpanelAPI.getInstance(MixpanelAPI.java:463)
	at com.mixpanel.android.mpmetrics.MixpanelAPI.getInstance(MixpanelAPI.java:236)

jaredmixpanel:
@MatthewWhitmer Just released a fix for this in `v8.0.1` can you please test and confirm the issue is resolved for you?

MatthewWhitmer:
@jaredmixpanel  Looks like 8.0.1 fixed the issue for me, thanks for the quick turnaround!

PR #847, opened 2024-09-23
Only register Broadcast receiver on API Level 33+ and non-Instant Apps

jaredmixpanel:
Fix for https://github.com/mixpanel/mixpanel-android/issues/846


zihejia (approved):
LGTM

PR #859, opened 2025-05-09
Initial Feature Flags Support

jaredmixpanel:
This PR adds initial support for using Mixpanel Feature Flags to the Android SDK. As part of this work we introduce the `MixpanelOptions` object to encapsulate all optional initialization parameters, i.e. `instanceName`, `optOutTrackingDefault`, `superProperties`, `featureFlagsEnabled` and `featureFlagsContext`... in order to avoid further constructor and `getInstance` signature proliferation. All of the existing signatures will be supported for backwards compatibility by ultimately building a `MixpanelOptions` object to be passed into the new canonical/designated constructor which expects a `MixpanelOptions` instance.

Example usage (in Kotlin, note that Kotlin automatically converts Java getters like `getFlags()` into properties like `flags`):

```
val flagsContext = JSONObject()
flagsContext.put("my_context_key", "my_context_value")
val mpOptionsBuilder = MixpanelOptions.Builder().featureFlagsEnabled(true).featureFlagsContext(flagsContext)
val mpOptions = mpOptionsBuilder.build()
val mixpanel = MixpanelAPI.getInstance(context, MIXPANEL_PROJECT_TOKEN, true, mpOptions)
mixpanel.flags.loadFlags()  // called automatically during SDK initialization
val ready = mixpanel.flags.areFlagsReady()
if (ready) {
    val featureFlagData = mixpanel.flags.getVariantSync(flagName, fallback)
}
mixpanel.flags.getVariant(flagName, fallback,
    FlagCompletionCallback { result -> 
        println("Key: ${result.key}")
        println("Value: ${result.value}")
    }
)
if (mixpanel.flags.areFlagsReady()) {
    val value = mixpanel.flags.getVariantValueSync(flagName, fallbackValue)
}
mixpanel.flags.getVariantValue(flagName, fallbackValue,
    FlagCompletionCallback { value -> 
        println("Value: $value")
    }
)
if (mixpanel.flags.areFlagsReady()) {
    val isEnabled = mixpanel.flags.isEnabledSync(flagName, fallbackValue)
}
mixpanel.flags.isEnabled(flagName, fallbackValue,
    FlagCompletionCallback { isEnabled -> 
        println("Is Enabled: $isEnabled")
    }
)
```

### DIFF
--- a/mixpaneldemo/src/main/java/com/mixpanel/mixpaneldemo/TrackingPage.kt
+++ b/mixpaneldemo/src/main/java/com/mixpanel/mixpaneldemo/TrackingPage.kt
@@ -112,12 +112,12 @@ fun TrackingPage(navController: NavHostController) {
         // --- Feature Flag Actions ---
         Triple("Load Flags", "Action: Loading flags async...", {
             println(">>> Action: Loading flags")
-            mixpanel.loadFlags()
+            mixpanel.flags.loadFlags()
             // You might want to call getFeature async shortly after to see results
         }),
         Triple("Are Flags Ready?", "Action: Checking areFeaturesReady...", {
             println(">>> Action: Checking areFeaturesReady")
-            val ready = mixpanel.areFeaturesReady()
+            val ready = mixpanel.flags.areFlagsReady()
             println("areFeaturesReady Result: $ready")
             // Update dialog or state if needed based on 'ready'
             dialogMessage.value = "areFeaturesReady() returned: $ready"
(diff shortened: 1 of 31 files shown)


mixpaneldemo/src/main/java/com/mixpanel/mixpaneldemo/TrackingPage.kt
```diff
@@ -88,6 +108,103 @@ fun TrackingPage(navController: NavHostController) {
             println("Unregistering super property")
             mixpanel.unregisterSuperProperty("property_to_unregister")
             mixpanel.flush()
+        }),
+        // --- Feature Flag Actions ---
```
jaredmixpanel, line 112: This is just test/demo code. It's actually going to be removed before merging.

rahul-mixpanel (approved):

PR #893, opened 2026-02-06
Fix GitHub release notes formatting

rahul-mixpanel:
- Replace deprecated actions/create-release with softprops/action-gh-release
- Use GitHub's native generate_release_notes feature instead of changelog generator
- Create releases as drafts for review before publishing


rahul-mixpanel:
this outdated as we have new release process

PR #924, opened 2026-04-07
fix(analytics): Integrate mixpanel-android-common library for event bridging

rahul-mixpanel:
## Summary
- Add `mixpanel-android-common:1.0.0` dependency for cross-SDK event communication
- Add `mavenLocal()` repository for SNAPSHOT resolution
- Emit events to `MixpanelEventBridge` on `track()` calls (fire and forget)

This enables other SDKs (e.g., mixpanel-android-session-replay) to subscribe to tracked events via the shared `MixpanelEventBridge.events()` flow.

## Test plan
- [ ] Verify build compiles successfully
- [ ] Test event tracking still works as expected
- [ ] Verify events are emitted to MixpanelEventBridge when track() is called


src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
```diff
@@ -494,6 +496,10 @@ public void handleMessage(Message msg) {
                                         eventDescription.getEventName(),
                                         eventDescription.getProperties());
                             }
+
+                            // Notify event bridge listeners with the complete event properties
+                            JSONObject properties = prepareEventObject(eventDescription).getJSONObject("properties");
```
ketanmixpanel: Confirm this is right place to notify the bridge.

ketanmixpanel (approved):
LG

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
tylerjroach: This should be ok, but thank you for bringing this up. We need to add first time event  to notify just a few lines further down

build.gradle
```diff
@@ -75,6 +75,16 @@ allprojects {
         }
     }
     repositories {
+        mavenLocal()
+        maven {
```
tylerjroach: This needs removed

tylerjroach (changes requested):

src/main/java/com/mixpanel/android/mpmetrics/AnalyticsMessages.java
rahul-mixpanel: thanks for bringing this up. I have updated the code.

tylerjroach (approved):

tylerjroach (approved):

tylerjroach (approved):

PR #935, opened 2026-04-14
fix: Use LifecycleProcess to modernize how we load feature flags in foreground

tylerjroach:
This change ensures we load feature flags
* on first foreground after initialize
* If initialized after first foreground, and ff have not loaded yet, load


msiebert (approved):

jaredmixpanel (approved):
LGTM

PR #936, opened 2026-04-14
fix: track() and flags getVariant() sequential ordering

tylerjroach:
Prevent a race condition where track() and flags.getVariant are called sequentially.

Previously, the tracked event was getting posted to a handler. Inside the handler, we were notifying FirstTimeEventListener.

If a customer were to call getVariant before the handler was able to process the tracked event, it could result in a race condition and return an unexpected value.


jaredmixpanel (approved):
LGTM

PR #941, opened 2026-04-15
fix: convert projectId into Long, not String

tylerjroach:


src/main/java/com/mixpanel/android/mpmetrics/FirstTimeEventDefinition.java
```diff
@@ -12,7 +12,7 @@
 class FirstTimeEventDefinition {
     @NonNull final String flagKey;
     @NonNull final String flagId;
-    @NonNull final String projectId;
+    @NonNull final Long projectId;
```
ketanmixpanel, line 15: Confirm that the `projectId` coming from API is Integer.

ketanmixpanel (approved):
LG

src/main/java/com/mixpanel/android/mpmetrics/FirstTimeEventDefinition.java
tylerjroach: Yep, using long here just in case

PR #947, opened 2026-04-21
chore: Updated CI release process

tylerjroach:


rahul-mixpanel:
## Issues to Address

### 1. Invalid action version comment
The SHA `de0fac2e4500dabe0009e67214ff5f5447ce83dd` is annotated as `# v6.0.2`, but `actions/checkout` has no v6 release. The latest is v4.x. Update comments to match actual versions.

### 2. Missing `all` scope in modules.json
The changelog script (`generate-changelog.sh:42`) accepts scope `all` in commits (`feat(all): ...`), but `all` isn't in `modules.json`. The PR title check would reject `feat(all): description`. Consider either:
- Adding `"all"` to the valid scopes in the PR title check regex
- Documenting that `all` is only for changelog, not PR titles

### 3. Snapshot workflow runs for all modules on every push
`release-snapshot.yml` publishes snapshots for ALL modules on every push to master. This means unrelated changes trigger unnecessary builds. Consider:
- Using `paths` filters like the old OpenFeature workflow had
- Or making this acceptable tradeoff explicit in PR description

### 4. No test step for analytics module in release workflow
`release-maven-central.yml:116` runs `"${TASK_PREFIX}test"` which becomes `:test` for the analytics module. The analytics module uses instrumented tests only (`connectedAndroidTest`), not unit tests. This step would pass vacuously. Consider either:
- Skipping tests for analytics (since they need a device)
- Or adding a `test_task` field to modules.json

### 5. Changelog extraction regex may fail
In `release-maven-central.yml:167`:
```bash
NOTES=$(sed -n '/^## \['"${TAG}"'\]/,/^## \[/{...}' "$CHANGELOG")
```
If `TAG` contains characters special to regex (unlikely but possible with pre-release tags like `v1.0.0-beta.1`), this could fail. Consider escaping.

### 6. README sed pattern is brittle
`prepare-release.yml:95`:
```bash
sed -i -E "s|^##### _.*_ - \[.*\]\(.*\)|..." "$README"
```
This assumes a very specific format. If the README doesn't have this exact pattern, the update silently fails. Consider adding validation or making it more robust.

---

Overall the architecture looks solid. The action version comment issue (#1) should be fixed before merge. The other items could be addressed now or in follow-ups.

rahul-mixpanel (approved):

PR #975, opened 2026-05-27
feat(analytics): add excludeProperties API for events and People.set()

tylerjroach:
## Summary
- Adds `MixpanelOptions.Builder.excludeProperties(Set<String>)` so projects can opt into stripping specific keys from outgoing payloads — useful when a downstream contract or warehouse caps per-payload size.
- Filter is applied at two chokepoints:
  - **Events**: `AnalyticsMessages.prepareEventObject` — covers super properties, caller properties, referrer properties, and SDK auto-properties uniformly.
  - **`People.set()`**: filters the bag *after* `mDeviceInfo` is merged but *before* it reaches `stdPeopleMessage`, so the 8 auto-injected device keys (`$android_lib_version`, `$android_os`, `$android_model`, etc.) are subject to the same exclude set as event auto-properties.
- Matching is exact and case-sensitive; null/empty set disables filtering with zero per-payload overhead.
- Identity/routing keys are protected via `MixpanelOptions.RESERVED_PROPERTY_KEYS` (`token`, `time`, `distinct_id`, `$device_id`, `$user_id`, `$had_persisted_distinct_id`) — the constant is the single source of truth for the runtime filter, the Javadoc, and the test (which iterates the constant so a new reserved key tightens coverage automatically).

## Scope notes
- **`People.set_once` / `$add` / `$append` / `$union` / `$remove` / `$unset` / `$merge` / `$delete`** are pass-through. They don't merge `mDeviceInfo`, and filtering inside mutating operators would silently change semantics (e.g. dropping a name from an `$unset` list).
- **Group updates** are unaffected — they never merge `mDeviceInfo`.
- `$mp_metadata` (sibling of `properties` on the event JSON) is structurally outside the filter's traversal scope.

### DIFF
--- a/analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
+++ b/analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
@@ -10,7 +10,18 @@ import com.mixpanel.android.util.ProxyServerInteractor;
 
 import org.json.JSONObject;
 
+import java.util.Collections;
+import java.util.Set;
+
 public class MixpanelOptions {
+
+    /**
+     * Property keys that are required for ingestion or identity resolution and will never be
+     * stripped by {@link Builder#excludeProperties(Set)}, even if a customer lists them.
+     */
+    public static final Set<String> RESERVED_PROPERTY_KEYS = Set.of(
+        "token", "time", "distinct_id", "$device_id", "$user_id", "$had_persisted_distinct_id"
+    );
 
     private final String instanceName;
     private final boolean optOutTrackingDefault;
@@ -19,6 +30,7 @@ public class MixpanelOptions {
     private final DeviceIdProvider deviceIdProvider;
     private final String serverURL;
     private final ProxyServerInteractor proxyServerInteractor;
+    private final Set<String> excludeProperties;
 
     private MixpanelOptions(Builder builder) {
         this.instanceName = builder.instanceName;
@@ -28,6 +40,7 @@ public class MixpanelOptions {
         this.serverURL = builder.serverURL;
         this.proxyServerInteractor = builder.proxyServerInteractor;
         this.mFeatureFlagOptions = builder.mFeatureFlagOptions;
+        this.excludeProperties = builder.excludeProperties;
     }
 
     public String getInstanceName() {
@@ -95,6 +108,18 @@ public class MixpanelOptions {
         return proxyServerInteractor;
     }
 
+    /**
+     * Returns the set of property keys that will be stripped from every event before it is
+     * sent to Mixpanel. Returns an empty set if none were configured. The returned set is
+     * unmodifiable.
+     *
+     * @return The configured exclude set, or an empty unmodifiable set.
+     */
+    @NonNull
+    public Set<String> getExcludeProperties() {
+        return excludeProperties;
+    }
+
     public static class Builder {
         private String instanceName;
         private boolean optOutTrackingDefault = false;
@@ -103,6 +128,7 @@ public class MixpanelOptions {
         private DeviceIdProvider deviceIdProvider = null;
         private String serverURL;
         private ProxyServerInteractor proxyServerInteractor;
+        private Set<String> excludeProperties = Collections.emptySet();
 
         public Builder() {
         }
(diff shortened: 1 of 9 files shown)


analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
```diff
@@ -278,6 +304,44 @@ public Builder serverURL(
...
+         *
+         * <p>A {@code null} or empty set disables filtering entirely with zero per-event
+         * overhead.
+         *
+         * @param excludeProperties The set of property keys to strip.
+         * @return This Builder instance for chaining.
+         */
+        public Builder excludeProperties(@Nullable Set<String> excludeProperties) {
+            if (excludeProperties == null || excludeProperties.isEmpty()) {
+                this.excludeProperties = Collections.emptySet();
```
rahul-mixpanel, line 349: We do null check before reading from excludeProperties in applyExcludeProperties. Can't we avoid assigning this emptySet here?

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
```diff
@@ -278,6 +304,44 @@ public Builder serverURL(
...
+            } else {
+                // Defensive copy + immutable wrapper so callers can't mutate post-build.
+                this.excludeProperties = Set.copyOf(excludeProperties);
+            }
```
rahul-mixpanel, line 353: Same issue — `Set.copyOf()` requires API 30+. Replace with:
```java
this.excludeProperties = Collections.unmodifiableSet(new HashSet<>(excludeProperties));
```

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
```diff
@@ -278,6 +304,44 @@ public Builder serverURL(
             return this;
         }
 
+        /**
```
rahul-mixpanel, line 311: **Documentation gap:** The Javadoc should clarify this only matches **top-level keys**. Users may reasonably expect `"user.email"` to exclude a nested property, but it won't — and fails silently.

**Feature consideration:** Since the SDK preserves nested JSONObjects (no flattening), dot-notation support would be valuable. Users excluding `$screen_height` will likely want to exclude `location.city` too.

At minimum, add to Javadoc:
```java
<p>Note: Only top-level property keys are matched. To exclude nested properties
(e.g., a {@code city} field inside a {@code location} object), you must exclude
the entire parent key. Dot-notation paths like {@code "location.city"} are not
currently supported.
```

Alternatively, consider supporting dot-notation traversal — the implementation would iterate path segments and call `remove()` on the leaf's parent.

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
```diff
@@ -10,15 +10,27 @@
...
+import java.util.Collections;
+import java.util.Set;
+
 public class MixpanelOptions {
 
+    /**
+     * Property keys that are required for ingestion or identity resolution and will never be
+     * stripped by {@link Builder#excludeProperties(Set)}, even if a customer lists them.
+     */
+    public static final Set<String> RESERVED_PROPERTY_KEYS = Set.of(
```
rahul-mixpanel: `Set.of()` requires API 30+ but minSdk is 21. This will crash with `NoSuchMethodError` on Android 10 and below (~23% of devices).

Use:
```java
public static final Set<String> RESERVED_PROPERTY_KEYS = Collections.unmodifiableSet(
    new HashSet<>(Arrays.asList(
        "token", "time", "distinct_id", "$device_id", "$user_id", "$had_persisted_distinct_id"
    ))
);
```

rahul-mixpanel (changes requested):

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
tylerjroach: I'm realizing mixpanel-android does not ask customers to have desugaring enabled.  There are zero IDE warnings to this. I'll add a plugin to help detect these issues.

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
tylerjroach: fixed

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
tylerjroach: fixed

tylerjroach: fixed

rahul-mixpanel (dismissed):

santigracia (approved):

rahul-mixpanel (approved):

PR #979, opened 2026-06-11
fix(session-replay): respect remoteSettingsMode for event triggers

rahul-mixpanel:
## Summary
- Refactored remote settings handling to use a new `ResolvedRemoteSettings` data class
- Consolidates all values needed for SDK initialization in one place:
  - `isRecordingEnabled`: always preserved from remote settings
  - `config`: processed MPSessionReplayConfig based on mode
  - `recordingEventTriggers`: null in DISABLED mode, otherwise from remote config
- Simplifies initialization code by removing scattered mode checks
- Event triggers from remote config are now properly skipped when `remoteSettingsMode = DISABLED`

## Test plan
- [x] Unit tests pass (`./gradlew :session-replay:testDebugUnitTest`)
- [ ] Manual test: Initialize with `remoteSettingsMode = DISABLED` and verify event triggers are not applied
- [ ] Manual test: Initialize with `STRICT` and `FALLBACK` modes and verify event triggers work

### DIFF
--- a/session-replay/src/main/java/com/mixpanel/android/sessionreplay/MPSessionReplay.kt
+++ b/session-replay/src/main/java/com/mixpanel/android/sessionreplay/MPSessionReplay.kt
@@ -30,6 +30,16 @@ sealed class MPSessionReplayError : Exception() {
     data class InitializationError(override val cause: Throwable) : MPSessionReplayError()
 }
 
+/**
+ * Consolidated result from applying remote settings.
+ * Contains all values needed for SDK initialization in one place.
+ */
+internal data class ResolvedRemoteSettings(
+    val isRecordingEnabled: Boolean,
+    val config: MPSessionReplayConfig,
+    val recordingEventTriggers: Map<String, RecordingEventTrigger>?
+)
+
 open class MPSessionReplay {
     // Class-level properties and other methods can go here if needed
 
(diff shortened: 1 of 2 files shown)


session-replay/src/main/java/com/mixpanel/android/sessionreplay/MPSessionReplay.kt
```diff
@@ -158,26 +168,25 @@ object SessionReplayManager {
 
                 // Now we're in foreground, safe to make network call
                 val settingsResult = remoteSettingsService.fetchRemoteSettings(token)
-                val finalConfig = applyRemoteSettings(config, settingsResult)
+                val resolvedSettings = resolveRemoteSettings(config, settingsResult)
```
ketanmixpanel, line 171: check below alternate approach as well and see what works best for you. 

```
val settingsResult = remoteSettingsService.fetchRemoteSettings(token)
val (finalConfig, finalSettings) = applyRemoteSettings(config, settingsResult)


@VisibleForTesting
internal fun applyRemoteSettings(
    config: MPSessionReplayConfig,
    remoteSettings: RemoteSettingsResult
): Pair<MPSessionReplayConfig?, RemoteSettingsResult> = when (config.remoteSettingsMode) {
    RemoteSettingsMode.DISABLED -> Pair(
        config,
        remoteSettings.copy(sdkConfig = null)
    )

    RemoteSettingsMode.STRICT -> if (remoteSettings.isFromCache || remoteSettings.sdkConfig == null) {
        Pair(null, remoteSettings)
    } else {
        Pair(applyRemoteConfigValues(config, remoteSettings), remoteSettings)
    }

    RemoteSettingsMode.FALLBACK -> Pair(applyRemoteConfigValues(config, remoteSettings), remoteSettings)
}

```

ketanmixpanel: Use `finalSettings` instead of `settingsResult` in the initialization code.

ketanmixpanel (approved):
LGTM.

PR #980, opened 2026-06-24
feat(analytics): add trackScreenView and trackScreenLeave tracking methods

rahul-mixpanel:
## Summary
This PR adds screen view and screen leave tracking methods to the Mixpanel Android SDK, organized under a new `Autocapture` inner class on `MixpanelAPI`.

### Core SDK
- Added `MixpanelAPI.Autocapture` public inner class with:
  - `trackScreenView(String screenName, JSONObject properties)` method
  - `trackScreenView(String screenName)` convenience method
  - `trackScreenLeave(String screenName, JSONObject properties)` method
  - `trackScreenLeave(String screenName)` convenience method
- Added `getAutocapture()` accessor on `MixpanelAPI`
- Both methods delegate to existing `track()` method with event names `$mp_page_view` and `$mp_page_leave`
- Automatically adds `current_page_title` property with the screen name
- Automatically adds `$mp_autocapture: true` property

### Tests
- Added comprehensive instrumented tests in `MixpanelScreenTrackingTest.java`
- Tests cover: basic usage, with/without properties, property merging

### Demo App
- Added automatic screen tracking to the Android demo app
- Uses Jetpack Compose Navigation state observation
- Demonstrates best practice implementation

## API Usage

```kotlin
// Kotlin
mixpanel.autocapture.trackScreenView("HomeScreen")
mixpanel.autocapture.trackScreenView("HomeScreen", properties)
mixpanel.autocapture.trackScreenLeave("HomeScreen")
mixpanel.autocapture.trackScreenLeave("HomeScreen", properties)
```

```java
// Java
mixpanel.getAutocapture().trackScreenView("HomeScreen");
mixpanel.getAutocapture().trackScreenView("HomeScreen", properties);
mixpanel.getAutocapture().trackScreenLeave("HomeScreen");
mixpanel.getAutocapture().trackScreenLeave("HomeScreen", properties);
```

## Testing
- All existing tests pass
- New instrumented tests added for the new methods
- Demo app can be run to test automatic tracking

## Related
Part of SDK-103: Adding screenView/screenLeave APIs across all platforms


analytics/mixpaneldemo/src/main/java/com/mixpanel/mixpaneldemo/MainActivity.kt
```diff
@@ -29,14 +31,46 @@ class MainActivity : ComponentActivity() {
...
+
     Scaffold(
         modifier = Modifier.fillMaxSize()
     ) { innerPadding ->
         NavGraph(navController = navController, modifier = Modifier.padding(innerPadding))
     }
 }
 
+// Helper function to format route names into readable screen names
+private fun formatScreenName(route: String): String {
```
tylerjroach: For simplicity sake, it may be best not to demo this. We could just use better route names. Screens don't necessarily have to be capitalized and spaced.

tylerjroach (commented):
few small comments to align on

tylerjroach (dismissed):

analytics/mixpaneldemo/src/main/java/com/mixpanel/mixpaneldemo/MainActivity.kt
```diff
@@ -29,7 +31,29 @@ class MainActivity : ComponentActivity() {
 
 @Composable
 fun MyApp() {
+    val context = LocalContext.current
     val navController = rememberNavController()
+    val mixpanel = remember { MixpanelAPI.getInstance(context, MIXPANEL_PROJECT_TOKEN, true) }
```
tylerjroach: If the demo app is configurable to reset the instance you wouldn't want to remember an old instance here. Given this is just a demo app, this isn't too critical. The lookup is cheap, but I'd just recommend the instance lookup in LaunchedEffect, so you can still have the mixpanel variable and you aren't looking it up with every re-composition.

tylerjroach (dismissed):

analytics/mixpaneldemo/src/main/java/com/mixpanel/mixpaneldemo/MainActivity.kt
rahul-mixpanel: since this is our sample code, I will make this change as well for better clarity

tylerjroach (approved):

PR #982, opened 2026-06-29
feat(analytics): add autocapture frustration signals (click, rage click, dead click)

rahul-mixpanel:
## Summary

- Implements autocapture Phase 1: `$mp_click`, `$mp_rage_click`, and `$mp_dead_click` event detection for both XML Views and Jetpack Compose UIs
- Adds `AutocaptureOptions` configuration (disabled by default, opt-in via `MixpanelOptions` builder) with per-feature options (`ClickOptions`, `RageClickOptions`, `DeadClickOptions`)
- Intercepts touch events via `Window.Callback` wrapping, with `WindowSpy` for dialog/popup/menu window tracking
- Full Compose support via native `SemanticsNode` tree traversal (graceful degradation when Compose is not present)

### Privacy

Autocapture does **not** capture visible text content (`$el_text`) from tapped elements. Tracking text can be invasive and raise privacy concerns. The complexity of nested view hierarchies can also cause text extraction to capture content from unintended views — for example, tapping a `LinearLayout` container might extract text from a deeply nested `TextView` that isn't semantically related to the tap.

The captured properties (`$el_id`, `$el_tag_name`, `$attr-aria-label`, `$attr-role`, `$elements`, `$x`, `$y`) are purely structural UI metadata with no PII risk.

### Captured Properties

| Property | Description |
|----------|-------------|
| `$x`, `$y` | Touch coordinates |
| `$el_id` | Element identifier (contentDescription → resource ID → fallback hash) |
| `$el_tag_name` | Element class name (e.g., `Button`, `TextView`) |
| `$attr-aria-label` | Accessibility label (contentDescription) |
| `$attr-role` | Semantic role (e.g., `Button`, `Switch`, `Slider`) |
| `$elements` | View hierarchy string (max 5 levels) |

### Architecture

| Component | Responsibility |
|-----------|---------------|
| `TouchInterceptor` | Per-window touch interception via `Window.Callback` |
| `SemanticExtractor` | Element info extraction (XML View + Compose routing) |
| `ComposeSemanticHelper` | Compose-specific semantics via `RootForTest` / `SemanticsNode` |
| `RageClickTracker` | 4+ clicks within 1000ms / 44dp radius |
| `DeadClickDetector` | No UI change within 500ms (XML: view hash, Compose: semantic snapshot) |
| `AutocaptureManager` | Lifecycle coordination, event dispatch |
| `WindowSpy` | Root view tracking for dialogs, popups, menus |

### Tests

- 7 XML autocapture instrumented tests (`AutocaptureInstrumentedTest`)
- 7 Compose autocapture instrumented tests (`ComposeAutocaptureInstrumentedTest`)
- All 14 autocapture tests pass on emulator (API 35)

### Demo app

- XML and Compose test screens added to `mixpaneldemo` for manual validation

### Hardening (code review fixes)

- Depth-limited recursion (max 20) on all tree traversals to prevent `StackOverflowError`
- Fixed `WeakHashMap` iteration in `stop()` to avoid `ConcurrentModificationException`
- Fixed `WindowSpy.getRootViews()` returning stale data
- Fixed `ClickEvent.isComposeClick()` unreliability from `WeakReference` GC
- Fixed `AccessibilityNodeInfo` leak on exception

## Test plan

- [ ] Run XML tests: `./gradlew :analytics:connectedAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.mixpanel.android.mpmetrics.AutocaptureInstrumentedTest`
- [ ] Run Compose tests: `./gradlew :analytics:connectedAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.mixpanel.android.mpmetrics.ComposeAutocaptureInstrumentedTest`
- [ ] Manual validation using demo app XML and Compose test screens
- [ ] Verify existing SDK tests still pass: `./gradlew :analytics:connectedAndroidTest`


analytics/docs/autocapture.md
```diff
@@ -0,0 +1,620 @@
...
+        }
+    ) {
+        Text("Checkout")  // → $el_text: "Checkout"
+    }
+}
+```
+
+## Disabling for Specific Elements
+
+Mark elements as sensitive to exclude them from **all** autocapture events:
```
tylerjroach: I'm just wondering why we would consider an action as sensitive. In this case, If mp-sensitive and mp-no-track are the same, mp-no-track seems sufficient.

context/AUTOCAPTURE.md
```diff
@@ -0,0 +1,620 @@
...
+```
+
+#### Jetpack Compose
+
+```kotlin
+// Recommended: Set semantics for reliable tracking
+Button(
+    onClick = { /* ... */ },
+    modifier = Modifier.semantics {
+        contentDescription = "checkout_button"
```
tylerjroach, line 181: what does it look like if semantics aren't set. I assume the text inside the button doesn't get captured in semantics does it? Just thinking if it does in any way there are some mask concerns to consider.

context/AUTOCAPTURE.md
```diff
@@ -0,0 +1,620 @@
...
+
+### Best Practices
+
+#### XML Views
+
+```xml
+<!-- Recommended: Set contentDescription for reliable tracking -->
+<Button
+    android:id="@+id/checkout_button"
+    android:contentDescription="checkout_button"
```
tylerjroach, line 163: xml approach can still read the id thought right?

analytics/docs/autocapture.md
```diff
@@ -0,0 +1,620 @@
...
+These controls still emit `$mp_click` events.
+
+### What Counts as UI Change
+
+- View count change (new views added/removed)
+- Content change (text, button states, etc.)
+- Window count change (alerts, dialogs, bottom sheets)
+- Navigation events (Activity changes)
+
+**Note:** Keyboard appearance does not count as a UI change for the tapped element.
```
tylerjroach: why though? Couldn't we figure out the window size changed? I'd like to avoid calling something a dead click that open as keyboard (clearly not a dead click in that case)

context/AUTOCAPTURE.md
```diff
@@ -0,0 +1,620 @@
...
+
+**ProGuard Note:** The SDK includes ProGuard rules to preserve `WindowManagerGlobal` reflection:
+
+```proguard
+-keep class android.view.WindowManagerGlobal { *; }
+```
+
+## Signal UI Change API
+
+For JS-based navigation frameworks (React Native) or custom navigation that bypasses standard Activity lifecycle, use `signalUIChange()` to prevent false dead click positives:
```
tylerjroach, line 307: Are we saying that they would have to implement this with any and all ui changes? Surely we do something different here?

analytics/docs/autocapture.md
```diff
@@ -0,0 +1,620 @@
...
+
+## Privacy Considerations
+
+### What is Captured
+
+- Touch coordinates
+- View class names and hierarchy
+- Content descriptions (accessibility labels)
+- Resource ID names
+- Visible text content (redacted for sensitive patterns)
```
tylerjroach: I don't think manual redaction is enough here. This feels like a need to have auto masking similar to Session Replay, with ability to explicitly unmask. I didn't realize we were capturing actual text here, just thought it was ids

tylerjroach (commented):
I haven't finished reviewing the code, but had these questions on autocapture doc

analytics/README.md
```diff
@@ -113,6 +115,51 @@ Once data hits our API, it generally takes ~60 seconds for it to be processed, s
 
 👋 👋  Tell us about the Mixpanel developer experience! [https://www.mixpanel.com/devnps](https://www.mixpanel.com/devnps) 👍  👎
 
+# Features
```
tylerjroach: I think we should just leave this to our docs site to handle to make sure we don't get out of date.

context/AUTOCAPTURE.md
rahul-mixpanel: Yes, but I have kept the `$el_id` resolution order (mentioned above) same for xml and compose for consistency.

context/AUTOCAPTURE.md
rahul-mixpanel: For $el_id, we follow the resolution order and fall back to generating ClassName_view_<hashCode>. We also capture the button text in $el_text, similar to the JS SDK. This makes it much easier to identify the element without having to rely on coordinates or other metadata.

I think we should make $el_text opt-in (disabled by default), similar to the JS SDK.

rahul-mixpanel: I will make that change. Good catch!

context/AUTOCAPTURE.md
rahul-mixpanel: This is a method I added for react native during planning. Once I test it after implementation. We will take a final call on weather we should keep this or not.

analytics/docs/autocapture.md
rahul-mixpanel: We do track `$el_text`. I will make it an opt-in feature.

analytics/docs/autocapture.md
rahul-mixpanel: We have it on JS SDK. For mobile, I think we can just keep mp-no-track. The sensitive one doesn't make sense for mobile use cases.

analytics/README.md
rahul-mixpanel: yeah, we should also remove the rest of the sample docs from this readme in that case.

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
```diff
@@ -129,6 +144,12 @@ public static class Builder {
         private boolean optOutTrackingDefault = false;
         private JSONObject superProperties;
         private FeatureFlagOptions mFeatureFlagOptions = new FeatureFlagOptions.Builder().build();
+        // Autocapture is disabled by default - all event types disabled
+        private AutocaptureOptions mAutocaptureOptions = new AutocaptureOptions.Builder()
+                .clickOptions(new ClickOptions.Builder().enabled(false).build())
```
tylerjroach: AutocaptureOptions.Builder().build() should default all of this without having to provide all of the individual options.

analytics/src/main/java/com/mixpanel/android/autocapture/WindowSpy.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        synchronized (sLock) {
+            if (sInstalled) {
+                return;
+            }
+
+            // Requires API 19+ for WindowManagerGlobal
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.KITKAT) {
+                MPLog.d(TAG, "WindowSpy not supported below API 19");
```
tylerjroach: min sdk is 21? We don't need this

analytics/src/main/java/com/mixpanel/android/autocapture/WindowSpy.java
```diff
@@ -0,0 +1,206 @@
...
+                sInstalled = true; // Mark as installed to avoid retries
+                return;
+            }
+
+            try {
+                // Access WindowManagerGlobal singleton
+                Class<?> wmgClass = Class.forName("android.view.WindowManagerGlobal");
```
tylerjroach: Just curious what was missing from public methods that require reflection.

analytics/src/main/java/com/mixpanel/android/autocapture/TouchInterceptor.java
```diff
@@ -0,0 +1,282 @@
...
+        float y = event.getRawY();
+
+        mTouchListener.onTouchUp(x, y, decorView);
+    }
+
+    // Delegate all other Window.Callback methods to the original callback
+
+    @Override
+    public boolean dispatchKeyEvent(KeyEvent event) {
+        return mOriginalCallback != null && mOriginalCallback.dispatchKeyEvent(event);
```
tylerjroach: Does our listener really eat all of the callbacks where we have to implement every override and forward along? Were there no gesture detectors where we only implement what we need?

analytics/src/main/java/com/mixpanel/android/autocapture/AutocaptureManager.java
```diff
@@ -0,0 +1,358 @@
+package com.mixpanel.android.autocapture;
+
+import android.app.Activity;
+import android.app.Application;
+import android.content.Context;
+import android.os.Bundle;
+import android.view.View;
+import android.view.Window;
+
+import androidx.annotation.NonNull;
```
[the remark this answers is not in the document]
rahul-mixpanel: edited the PR description

analytics/docs/autocapture.md
rahul-mixpanel: This is a change needed in the doc. We don't consider input text fields for dead click evaluation. I will fix the doc.

analytics/src/main/java/com/mixpanel/android/autocapture/SemanticExtractor.java
```diff
@@ -0,0 +1,688 @@
...
+            for (int i = group.getChildCount() - 1; i >= 0; i--) {
+                View child = group.getChildAt(i);
+                View result = findViewAtPosition(child, x, y);
+                if (result != null) {
+                    return result;
+                }
+            }
+        }
+
+        return view;
```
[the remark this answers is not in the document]
rahul-mixpanel: The View hierarchy traversal is naturally bounded — Android view trees are typically 10-15 levels deep, and the recursion terminates as soon as children fall outside the tap coordinates (bounds check). The depth limit on the Compose path (`MAX_RECURSION_DEPTH = 20`) is purely defensive and would never fire in a real app either. That said, we'll add the same limit here for parity — it's a single int comparison per call, so zero performance cost.

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
```diff
@@ -226,6 +247,37 @@ public Builder featureFlagOptions(FeatureFlagOptions featureFlagOptions) {
...
+         * }</pre>
+         *
+         * @param autocaptureOptions The AutocaptureOptions configuration.
+         * @return This Builder instance for chaining.
+         * @see AutocaptureOptions
+         */
+        public Builder autocaptureOptions(AutocaptureOptions autocaptureOptions) {
+            this.mAutocaptureOptions = autocaptureOptions;
+            return this;
+        }
```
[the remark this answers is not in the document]
rahul-mixpanel: ignoring the null value. can't crash the host app.

analytics/src/main/java/com/mixpanel/android/autocapture/TouchInterceptor.java
rahul-mixpanel: Yes — we replace the Window.Callback via `window.setCallback()`, so we must forward all methods or they get silently swallowed (menus, action modes, etc. would break). View-level alternatives like `OnTouchListener` or `GestureDetector` are view-scoped and can't intercept touches on dialogs, popups, or spinners (separate windows), which we need for multi-window autocapture via WindowSpy. The interface is stable (last change was API 23) so the boilerplate is a one-time cost.

rahul-mixpanel: This was inspired by the source code of Curtains.

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
rahul-mixpanel: It already does — `new AutocaptureOptions.Builder().build()` enables all three event types (click, rage click, dead click) with default settings. The builder fields are pre-initialized (lines 104-106). Added a Javadoc on the constructor to make this clearer.

analytics/src/main/java/com/mixpanel/android/autocapture/WindowSpy.java
rahul-mixpanel: Android has no public API to observe window additions/removals (dialogs, popups, spinners, menus). `WindowManager` only lets you add/remove views from your own app — there's no listener for when the system adds windows. We need `WindowManagerGlobal.mViews` (the internal list of all root views) to intercept new windows as they appear and install touch interceptors on them. Without this, we'd only capture touches on Activity windows and miss dialogs, bottom sheets, spinners, etc. The reflection gracefully degrades — if it fails (e.g., on future Android versions), autocapture falls back to Activity-only tracking.

analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelOptions.java
rahul-mixpanel: Oh, on a second thought, I get your point here. But if we do so, then the developer will have to write this code to enable each one of them.

rahul-mixpanel: Now, `AutocaptureOptions.Builder().build()` enables all auto-capture events with their default configuration. Customers can then selectively disable the ones they don’t want.

analytics/README.md
rahul-mixpanel: removed my changes

context/AUTOCAPTURE.md
rahul-mixpanel: we don't track text anymore

analytics/src/main/java/com/mixpanel/android/autocapture/TouchInterceptor.java
tylerjroach: If curtains is also doing this, will our library and curtains still work together?

analytics/src/main/java/com/mixpanel/android/autocapture/AutocaptureManager.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    /**
+     * Starts autocapture.
+     *
+     * <p>Registers lifecycle callbacks and installs window tracking.
+     * This should be called once during SDK initialization.
+     */
+    public void start() {
+        if (mStarted) {
```
tylerjroach, line 97: Should probably sync on a lock, especially since we don't set mStarted right away

analytics/src/main/java/com/mixpanel/android/autocapture/AutocaptureManager.java
```diff
@@ -0,0 +1,358 @@
...
+            MPLog.e(TAG, "Failed to start autocapture", e);
+        }
+    }
+
+    /**
+     * Stops autocapture.
+     *
+     * <p>Unregisters lifecycle callbacks and removes all interceptors.
+     */
+    public void stop() {
```
tylerjroach, line 135: sync on same lock started needs.

analytics/src/main/java/com/mixpanel/android/autocapture/AutocaptureManager.java
```diff
@@ -0,0 +1,358 @@
...
+
+    private void processTouchEvent(float x, float y, @NonNull View decorView) {
+        // Extract semantics from the touched view
+        ClickEvent.Builder builder = SemanticExtractor.extract(decorView, x, y);
+        if (builder == null) {
+            // No view found at position
+            return;
+        }
+
+        ClickEvent clickEvent = builder.build();
```
tylerjroach: Why is ClickEvent using a builder? If it makes it easier to construct a ClickEvent thats fine, but I don't know if it makes sense for methods such as the one above to return the builder, rather than the full ClickEvent.

analytics/src/main/java/com/mixpanel/android/autocapture/ComposeSemanticHelper.java
```diff
@@ -0,0 +1,485 @@
...
+                ", isRootForTest: " + (view instanceof RootForTest) + ", x: " + x + ", y: " + y);
+
+        if (!(view instanceof RootForTest)) {
+            MPLog.d(TAG, "View is not RootForTest, returning NOT_FOUND");
+            return ExtractResult.notFound();
+        }
+
+        RootForTest root = (RootForTest) view;
+        SemanticsNode rootNode = root.getSemanticsOwner().getRootSemanticsNode();
+        Rect rootBounds = rootNode.getBoundsInWindow();
```
tylerjroach: Lets do some manual testing on here to make sure that the coordinates we calculate take into virtual device bars (status bars, virtual navigation buttons on the bottom, notches, etc. Want to make sure when we are finding a node at a position, its all using the same calculations so we don't end up calculating slightly off values.

analytics/src/main/java/com/mixpanel/android/autocapture/ComposeSemanticHelper.java
```diff
@@ -0,0 +1,485 @@
...
+            // Content hash changed = state change (text, labels, etc.)
+            if (this.contentHash != current.contentHash) {
+                MPLog.d(TAG, "Snapshot: contentHash changed");
+                return true;
+            }
+
+            // Node count increased by 5+ = likely navigation (new screen added)
+            // During navigation animation, both screens may be present
+            if (current.nodeCount >= this.nodeCount + 5) {
+                MPLog.d(TAG, "Snapshot: nodeCount increased significantly (" +
```
tylerjroach: This node count feels very arbitrary. Is there a reason 5 was selected? I assume this is for dead click detection, but wouldn't any node changes at all mean no dead click? Wouldn't even a change of 1 cause contentHash to be calculated as different?

tylerjroach:
Not to overwhelm here but some notable findings to look into from Fable.  

### 1. Autocapture events silently dropped when `trackAutomaticEvents=false`
  **`MixpanelAPI.java:2204`** — The emit callback calls `track(eventName, properties, /*isAutomaticEvent=*/true)`, and `track()` (line 2805) early-returns on `(isAutomaticEvent && !mTrackAutomaticEvents)`. An app created with `getInstance(context, token, false, optionsWithAutocaptureEnabled)` starts the full autocapture machinery (interceptors, WindowSpy, timers) yet emits zero events — "Autocapture initialized" is logged and nothing hints at the drop. Autocapture opt-in should gate itself rather than piggyback on the automatic-events flag.

  ### 2. `WindowSpy` mViews swap is an unsynchronized race that can crash the host app
  **`WindowSpy.java:80-84`** — `install()` copies `WindowManagerGlobal.mViews` into a new list and writes the field back while holding only WindowSpy's own lock (not the framework's `mLock`), running synchronously on whatever thread calls `getInstance()`. A root view added by the main thread between the copy and the field write vanishes from `mViews`; its later `removeView` throws `IllegalArgumentException("View not attached to window manager")` in the host app — violating the never-crash rule. This swap technique is only safe when performed on the main thread.

### 3. Every single-pointer `ACTION_UP` is treated as a click — scrolls and swipes emit `$mp_click`
  **`TouchInterceptor.java:107-125`** — `ACTION_DOWN` is never recorded, so there is no touch-slop or duration check, and `ACTION_CANCEL`/`ACTION_MOVE` are ignored (no mitigation exists anywhere downstream). Every scroll stroke on a `RecyclerView`/`LazyColumn` produces a spurious click on whatever was under the finger at lift-off, and a few quick strokes in one region synthesize a false `$mp_rage_click`. A slop/duration check at this layer is standard.

### 4. Responsive buttons falsely reported as dead clicks (XML path)
  **`DeadClickDetector.java:201-224`** — The baseline snapshot is taken 150ms *after* the click, and `onGlobalLayout`/`onScrollChanged` explicitly ignore anything before `mBaselineCaptured` ("expected settling"). A click handler that updates the UI within a frame or two — the common case — has its response absorbed into the baseline; at the 500ms check the tree matches and a false `$mp_dead_click` fires. The Compose path does this correctly by snapshotting synchronously at click time.

### 5. No retroactive attach — deferred SDK init captures nothing on the current screen
  **`AutocaptureManager.java:101-122`** — `start()` only registers forward-looking hooks: `registerActivityLifecycleCallbacks` doesn't replay `onActivityResumed` for an already-resumed activity, and `DelegatingViewList`'s copy-constructor bulk copy bypasses the overridden `add()`, so existing root views never fire `onRootViewChanged`. With deferred init (e.g., after a consent dialog), the foreground activity gets no interceptor until navigation or a config change.

### 6. Dialogs, popups, and menus are never captured — WindowSpy attaches to nothing
  **`AutocaptureManager.java:339-357`** — `getWindowFromView()` only handles `context instanceof Activity` directly (no `ContextWrapper` unwrapping; the "reflection fallback" comment at line 341 has no code behind it). Dialog decor views carry a `DecorContext`/`ContextThemeWrapper`, so the method returns null and no interceptor attaches. When the context *is* the Activity (popups), it returns the activity's already-registered window, so the `containsKey` guard skips attachment. Net effect: the entire WindowSpy mechanism currently attaches interceptors to nothing.

### 7. Compose hit-testing mixes coordinate spaces
  **`ComposeSemanticHelper.java:226`** — `TouchInterceptor` captures `getRawX()/getRawY()` (screen space) and passes them unconverted to `findNodeAtPositionRecursive`, which tests against `SemanticsNode.getBoundsInWindow()` (window space). The XML and accessibility paths correctly use screen space (`getLocationOnScreen`, `getBoundsInScreen`). For any Compose window not at the screen origin — dialogs, popups, multi-window — the tap misses or attributes to the wrong node.

### 8. Two `Window.Callback` methods are not delegated
  **`TouchInterceptor.java:128-281`** — `onProvideKeyboardShortcuts` (API 24) and `onPointerCaptureChanged` (API 26) are not overridden. Both are `default` no-ops in the interface, so this compiles silently, but once wrapped those framework notifications never reach the host Activity — breaking the keyboard-shortcuts sheet (Meta+/) and pointer-capture notifications. The existing `Api23Helper` + animalsniffer-ignore pattern covers this cleanly.

  #### 9 `$tap_count` documented but never emitted
  **`RageClickTracker.java:92-95`** — `autocapture.md:150` documents `$tap_count` ("Number of taps in the rage click sequence") on `$mp_rage_click`, but `recordClick` clears the click history and returns the triggering event unchanged, and `ClickEvent.toProperties()` has no tap-count field. The documented property never appears in any emitted event, and once cleared the actual count is unrecoverable.

  #### 10. `stop()` / `uninstall()` latent defects
  **`TouchInterceptor.java:67-76`** — `uninstall()` only restores the callback when `getCallback() == this`; if another SDK wrapped the callback after Mixpanel, it silently no-ops and the stale interceptor keeps emitting clicks forever — there is no detached/stopped flag anywhere on the event path. `install()` is also not idempotent (no check for an
  already-installed `TouchInterceptor`), so a stop/start cycle would double-wrap and double-emit every click. Currently unreachable — nothing in the SDK calls the public `stop()` — but worth fixing before it becomes reachable.

  ## Autocapture PR Review — Additional Findings (design, performance, cleanup)

  ### Design / architecture

  #### D1. "Disabled by default" is a hand-built template, not a real off switch
  **`MixpanelOptions.java:147-152`** — The default is implemented by building an `AutocaptureOptions` that individually disables each of the three current sub-options, and `AutocaptureOptions.isEnabled()` is an OR over sub-options. When Phase 2 adds a new sub-option
  whose builder defaults to `enabled=true` (the existing convention in `ClickOptions`/`RageClickOptions`/`DeadClickOptions`), every app that never called `autocaptureOptions()` silently gets that capture enabled. Fix: a master enabled flag or an
  `AutocaptureOptions.disabled()` factory so "off" is defined once, in the class that owns the sub-options. *(flagged independently by 2 agents)*

  #### D2. Dead-click detection is two divergent mechanisms instead of one abstraction
  **`DeadClickDetector.java:188`** — `DetectionSession` branches on `mIsComposeClick` throughout start/captureBaseline/checkResult/cleanup, and the paths already disagree: `baselineDelayMs` silently doesn't apply to Compose, layout/scroll cancellation signals only exist
  for XML, and `onWindowFocusChanged()` (line 124) is never called by anyone — dead code. Every new change signal or Phase 2 surface must be added to both branches. Fix: a UI-change-monitor interface (captureBaseline / hasChanged / attach / detach) chosen once per
  click.

  #### D3. Compose change-detection sensitivity is a hardcoded heuristic
  **`ComposeSemanticHelper.java:136`** — The Compose path classifies a click as dead unless the node count changes by more than ±5, bypassing both `AutocaptureDefaults` and `DeadClickOptions`, while the XML path uses exact count/hash equality. A Compose click that
  adds/removes 1-4 nodes (toggling an icon, revealing a badge) is a false dead click, and there's no option to tune it. Sensitivity should be defined once and honored by both snapshot implementations.

  #### D4. `MAX_ACCESSIBILITY_NODES` doesn't enforce its documented contract
  **`SemanticExtractor.java:215`** — The constant (500, documented as "maximum number of accessibility nodes to probe") is actually used as a recursion *depth* bound in `findNodeAtPosition`. A broad-but-shallow tree probes unbounded nodes per tap, while 500 as a depth
  limit is far too deep to prevent the StackOverflowError it presumably targets (other traversals cap depth at 20). Fix: a shared visited-node counter for total work, with depth reusing `MAX_RECURSION_DEPTH`.

  #### D5. `WindowSpy` install is irreversible and failure is silent-permanent
  **`WindowSpy.java:118`** — There is no uninstall path (`stop()` removes listeners but the swapped list stays forever), `sInstalled` is set true even when the reflection fails, so degradation is permanent and undetectable by the caller, and `DelegatingViewList` only
  intercepts `add`/`remove` — not `addAll`/`clear`/`set` — so coverage depends on OS internals. If another library uses the same mViews-swap technique, whichever installs second wraps the other and neither can be removed. Fix: keep the original list for restore-on-stop,
  and surface install failure.

  #### D6. Duplicate `ActivityLifecycleCallbacks` registration
  **`AutocaptureManager.java:109`** — The SDK already registers `MixpanelActivityLifecycleCallbacks` on the same Application (`MixpanelAPI.java:2178`); `AutocaptureManager` registers a second, and with multiple Mixpanel instances each adds another. Dispatching
  autocapture's resume/pause/destroy hooks from the existing callbacks would keep one lifecycle path to maintain.

  #### D7. XML hit-testing picks the deepest view, not the clickable target
  **`SemanticExtractor.java:506`** — `findViewAtPosition` returns the deepest visible view with no preference for clickable views or walk-up to a clickable ancestor (unlike the accessibility path). A tap on a clickable container built as ViewGroup + TextView is
  attributed to the inner TextView: wrong `$el_id`, role `text` instead of `button`, and `isInteractive=false` — which also silently skips dead-click detection for that element.

  ### Performance (tap hot path — all on the main thread)

  #### P1. Extraction runs before the app receives the touch event
  **`TouchInterceptor.java:90`** — On every ACTION_UP, the full hit test, hierarchy string build, and (for Compose) semantics traversal + snapshot run synchronously *before* `mOriginalCallback.dispatchTouchEvent(event)`, adding latency between finger-up and the app's 

  ### Performance (tap hot path — all on the main thread)

  #### P1. Extraction runs before the app receives the touch event
  **`TouchInterceptor.java:90`** — On every ACTION_UP, the full hit test, hierarchy string build, and (for Compose) semantics traversal + snapshot run synchronously *before* `mOriginalCallback.dispatchTouchEvent(event)`, adding latency between finger-up and the app's click handling. Nothing in the extraction depends on running first — forward the event,
  then extract (or post it).

  #### P2. Per-node `getLocationOnScreen` during hit-test DFS
  **`SemanticExtractor.java:483`** — Each visited view allocates an `int[2]` and calls `getLocationOnScreen` (which walks the parent chain), making the hit test O(visited × depth) per tap. Converting the tap to root coordinates once and translating incrementally while descending (as `ViewGroup.dispatchTouchEvent` does) eliminates both costs.

  #### P3. Dead-click snapshots walk the tree twice, twice
  **`DeadClickDetector.java:250`** — `captureBaseline()` and `checkResult()` each traverse the full hierarchy twice (`countViews` + `computeContentHash`) — 4 full traversals per interactive click where 1-2 suffice. Compute count and hash in a single pass (as `ComposeSemanticHelper.computeTreeHash` already does), or drop `countViews` since the hash
  already reflects count changes.

  #### P4. Per-node allocation in Compose snapshots
  **`ComposeSemanticHelper.java:207`** — `computeTreeHash` allocates an `int[2]` per semantics node, and the snapshot runs at click time and again at timeout — hundreds of short-lived arrays per interactive click on large screens. Pack count+hash into a single long return, or use one holder per snapshot.

  #### P5. Hot-path debug logs build strings regardless of log level
  **`SemanticExtractor.java:60`** (also ComposeSemanticHelper.extract, captureSnapshot, AutocaptureManager.emitEvent) — `MPLog.d` checks the level inside the call, so concatenation and `getSimpleName()` run on every tap even at the default WARN level. Guard with a level check or remove per-tap logs.

  ### Dead code / simplification

  - **`AutocaptureManager.java:66`** — `mCurrentActivityRef` is write-only: assigned every `onActivityResumed`, never read. Delete. *(flagged independently by 3 agents)*
  - **`WindowSpy.java:128`** — `getRootViews()` and the `sOriginalViews` field exist only to serve it; zero callers. Delete both. *(flagged independently by 2 agents)*
  - **`ClickEvent.java:49`** — `timestamp` is set, threaded through the 10-arg constructor, and never serialized or read (`RageClickTracker` keeps its own timestamps). Drop it.
  - **`DeadClickDetector.java:167`** — `DetectionSession` caches `mIsComposeClick`/`mComposeRootRef` duplicating state already available from the `mClickEvent` it holds (double-wrapping an already-weak reference). Use the ClickEvent accessors.
  - **`ComposeSemanticHelper.java:46`** — `ExtractResult` wrapper + two-value `ExtractionResult` enum encode exactly "nullable builder". Return `@Nullable ClickEvent.Builder` and delete both types.
  - **`SemanticExtractor.java:158,183,293`** — Dead SDK-version guards for API 16/18; minSdk is 21, so the early-return is unreachable and the gates are always true.
  - **`SemanticExtractor.java:516` / `ComposeSemanticHelper.java:303`** — The `$el_id` priority chain (contentDescription > id/testTag > fallback hash) and role mapping are copy-pasted across the XML, accessibility, and Compose paths — three places to edit, already drifting. Extract one shared resolver.

  ### Conventions (per CLAUDE.md)

  - **`AutocaptureOptions.java:91`** — Builder setters (`clickOptions`/`rageClickOptions`/`deadClickOptions`) store `@NonNull` params with no null guard; a Java caller passing null NPEs later in `AutocaptureOptions.isEnabled()`, which runs in the `MixpanelAPI` constructor outside any try/catch — reachable host-app crash during `getInstance()`. The
  sibling `MixpanelOptions.Builder.autocaptureOptions` added in this PR does guard; mirror it.
  - **`AutocaptureManager.java:57,59`** — `mRageClickTracker` and `mDeadClickDetector` are assigned only in the constructor but declared non-final ("Final fields for immutability").
  - **`ClickEvent.java:23`** — Instance fields (`x`, `y`, `elementId`, …) omit the `m` prefix used by every other new class in this PR ("Member variables prefixed with 'm'").

analytics/src/main/java/com/mixpanel/android/autocapture/TouchInterceptor.java
rahul-mixpanel: identified that it may cause issues when used along with curtains. Since the SR SDK depends on Curtains, I am bringing the same to analytics to avoid this issue in future when customers use both of these SDKs together.

analytics/src/main/java/com/mixpanel/android/autocapture/WindowSpy.java
```diff
@@ -0,0 +1,128 @@
...
+                return;
+            }
+
+            try {
+                sCurtainsListener = (view, added) -> notifyListeners(view, added);
+                Curtains.getOnRootViewsChangedListeners().add(sCurtainsListener);
+                sInstalled = true;
+                MPLog.d(TAG, "WindowSpy installed via Curtains");
+            } catch (Exception e) {
+                MPLog.e(TAG, "Failed to install WindowSpy via Curtains, "
```
tylerjroach, line 74: Is this still accurate that there is other tracking? Is there any exception to even be caught from curtains?

analytics/src/main/java/com/mixpanel/android/autocapture/DeadClickDetector.java
```diff
@@ -0,0 +1,408 @@
...
+
+        // Only monitor interactive elements
+        if (!clickEvent.isInteractive) {
+            return;
+        }
+
+        try {
+            // Choose the appropriate monitor based on click type
+            UiChangeMonitor monitor;
+            if (clickEvent.isComposeClick()) {
```
tylerjroach: Many apps are mixed between compose and xml. Will the change monitor detect a click that goes from a compose view to an xml view since these monitors look to be specific for the view type that came from the click?

 Or a better example would be if a compose view is added to an xml view. If a button came from xml but changed the contents inside compose (or vise versa) would we detect this?

analytics/src/main/java/com/mixpanel/android/autocapture/AutocaptureManager.java
rahul-mixpanel: All the public APIs in this class needs to be called from Main thread. So there is no need to sync this. I have added `@MainThread` notations to make it explicitly clear.

analytics/src/main/java/com/mixpanel/android/autocapture/AutocaptureManager.java
rahul-mixpanel: same response as above comment

analytics/src/main/java/com/mixpanel/android/autocapture/AutocaptureManager.java
rahul-mixpanel: The Builder is confined to the extraction phase — `SemanticExtractor` has three paths (XML view, accessibility, Compose semantics) that each populate different fields, and the caller sometimes augments the result after extraction (e.g., setting `composeRoot`). Once extraction is done, `.build()` is called exactly once in `processTouchEvent` and everything downstream (rage click, dead click, emission) receives the finalized immutable `ClickEvent`. Happy to refactor if you feel strongly, but the current boundary keeps it to one allocation per tap.

analytics/src/main/java/com/mixpanel/android/autocapture/ComposeSemanticHelper.java
rahul-mixpanel: Good catch — you're right. `computeTreeHash` combines child hashes sequentially with `31 * hash + childResult[1]`, so any structural change (adding/removing even a text-less node) produces a different hash. The ±5 node count check was redundant. Removed it — `hasChanged()` now relies solely on `contentHash` comparison.

analytics/src/main/java/com/mixpanel/android/autocapture/WindowSpy.java
rahul-mixpanel: The try-catch is defensive — Curtains relies on `WindowManagerGlobal.mViews` reflection which has been stable since API 17, but the library is no longer actively maintained. If a future Android version changes or restricts this internal API, the catch ensures we degrade gracefully (activity windows still work via lifecycle callbacks, only dialogs/popups lose coverage) rather than crashing the host app. Added a comment to clarify.

analytics/src/main/java/com/mixpanel/android/mpmetrics/AutocaptureOptions.java
```diff
@@ -0,0 +1,171 @@
...
+         */
+        public Builder deadClickOptions(@NonNull DeadClickOptions deadClickOptions) {
+            this.mDeadClickOptions = deadClickOptions;
+            return this;
+        }
+
+        /**
+         * Builds and returns an {@link AutocaptureOptions} instance with the configured settings.
+         *
+         * @return A new {@link AutocaptureOptions} instance.
```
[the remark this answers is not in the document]
rahul-mixpanel: added null checks

analytics/src/main/java/com/mixpanel/android/autocapture/CurtainsHelper.kt
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        window.touchEventInterceptors += listener
+        return listener
+    }
+
+    /**
+     * Removes a previously installed touch listener from a [Window].
+     */
+    @JvmStatic
```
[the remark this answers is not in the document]
rahul-mixpanel: addressed

analytics/src/main/java/com/mixpanel/android/autocapture/ComposeSemanticHelper.java
rahul-mixpanel: identified gaps on split-screen use cases and fixed them.

context/AUTOCAPTURE.md
```diff
@@ -0,0 +1,542 @@
+# Android Autocapture
```
tylerjroach, line 1: This was useful for review to fully understand the feature set, but can we delete this file in favor of our public docs. Is easy for these pages to become outdated and forgotten

analytics/src/androidTest/java/com/mixpanel/android/mpmetrics/AutocaptureComposeInstrumentedTest.kt
```diff
@@ -0,0 +1,385 @@
...
+    @Before
+    fun setUp() {
+        mContext = InstrumentationRegistry.getInstrumentation().context
+
+        val messages = AnalyticsMessages.getInstance(
+            mContext,
+            MPConfig.getInstance(mContext, null)
+        )
+        messages.hardKill()
+        Thread.sleep(2000)
```
tylerjroach: There are a lot of sleeps on this test file. Can we replace with something better?

analytics/src/androidTest/java/com/mixpanel/android/mpmetrics/AutocaptureMixedFrameworkInstrumentedTest.kt
```diff
@@ -0,0 +1,356 @@
...
+    @Before
+    fun setUp() {
+        mContext = InstrumentationRegistry.getInstrumentation().context
+
+        val messages = AnalyticsMessages.getInstance(
+            mContext,
+            MPConfig.getInstance(mContext, null)
+        )
+        messages.hardKill()
+        Thread.sleep(2000)
```
tylerjroach: Lets try to replace the sleep's

analytics/src/androidTest/java/com/mixpanel/android/mpmetrics/AutocaptureXmlInstrumentedTest.java
```diff
@@ -0,0 +1,474 @@
...
+    @Before
+    public void setUp() throws Exception {
+        mContext = InstrumentationRegistry.getInstrumentation().getContext();
+
+        // Hard-kill existing AnalyticsMessages instance for clean state
+        AnalyticsMessages messages = AnalyticsMessages.getInstance(
+                mContext,
+                MPConfig.getInstance(mContext, null));
+        messages.hardKill();
+        Thread.sleep(2000);
```
tylerjroach: Try to replace sleeps

analytics/src/main/java/com/mixpanel/android/autocapture/ClickEvent.java
```diff
@@ -0,0 +1,269 @@
...
+     * Defaults to {@code null} if not provided.
+     */
+    @Nullable
+    public final String elements;
+
+    /** Whether the clicked view is considered interactive (clickable/longClickable). */
+    public final boolean isInteractive;
+
+    /** Whether this click originated from a Compose element (set at construction, not affected by GC). */
+    final boolean isCompose;
```
tylerjroach: We've got isCompose and composeRootRef. Would it make sense to have a ComposeClickEvent extension so that we don't potentially clutter a single class with variables that may or may not apply to the view system type?

analytics/src/main/java/com/mixpanel/android/autocapture/CurtainsHelper.kt
```diff
@@ -0,0 +1,119 @@
...
+                    MotionEvent.ACTION_CANCEL -> {
+                        downTime = 0L
+                    }
+                    MotionEvent.ACTION_UP -> {
+                        if (event.pointerCount == 1) {
+                            val duration = event.eventTime - downTime
+                            val dx = event.rawX - downX
+                            val dy = event.rawY - downY
+                            if (downTime > 0 && duration <= MAX_TAP_DURATION_MS
+                                && (dx * dx + dy * dy) <= touchSlopSq
```
tylerjroach: touch slop check here is not ideal because this will give a false tap report if someone touches + quickly swipes down and back up to the same position. We should be looking at the move event and if at any time there is a move that extended past our slop threshold, we should cancel the touch.

tylerjroach (commented):
While the below is from Claude, this is because I was recognizing this same concern. We've made an effort on SR to ensure each node is visited once. There are a couple of methods that are not obvious in their usage that cause a fully tree traversal. Even though we are only running this on click detections, on slow devices this can be expensive and could cause frame drops. Lets see if we can optimize this.

  SemanticExtractor.extract currently walks the View tree three times per tap:

  1. findViewAtPosition walks down from root to target (SemanticExtractor.java:490-528)
  2. findComposeRoot walks up from target back to root (SemanticExtractor.java:92-115)
  3. buildHierarchyString / buildHierarchyFromNode walks up again to build the ancestor chain (SemanticExtractor.java:389-423 and 565-585)

  On top of that, each descent step in #1 calls view.getLocationOnScreen(location) (line 505), which itself walks up the parent chain internally — so the "single descent" is actually O(depth × visited) rather than O(visited).

  SensitiveViewManager.processSubviews in session-replay is deliberate about visiting each node exactly once — see the ViewContext pattern at SensitiveViewManager.kt:420-423, and the explicit "avoids walking up tree for every view" rationale at line 500. Parent-derived state is threaded downward through
  the traversal instead of being re-derived by walking back up.

  Suggest restructuring extract around the same discipline: a single iterative descent that carries the state it needs — local coordinates translated on the way down, a composeRoot reference set when we cross an AndroidComposeView, and the ancestor chain captured in traversal order. At the leaf we'd
  already have the target view, the compose root (or null), and the hierarchy chain — no post-hoc parent walks needed.

  That change dissolves the O(depth × visited) hit-test, findComposeRoot's ancestor scan, and the O(depth²) insert(0, …) in hierarchy building in one pass, and keeps this file consistent with the "visit each node once" rule already established in session replay.

rahul-mixpanel:
Re: deleting `context/AUTOCAPTURE.md` — this file is intentionally kept as a context file for Claude Code (our AI coding assistant), not as public-facing documentation. It helps Claude understand the autocapture feature when working on this codebase. It won't be linked from public docs.

context/AUTOCAPTURE.md
rahul-mixpanel: this is added as a context file for Claude

rahul-mixpanel:
Good call — addressed in 42c76eea. Here's what changed:

1. **Merged `findComposeRoot` into `findTargetView`** — Compose root is now detected during the downward descent, eliminating the separate O(h) upward walk.

2. **Reused `int[2]` buffer** — single allocation passed through all recursive calls instead of `new int[2]` per node.

3. **Pre-built hierarchy passed via `HitResult`** — `findTargetView` returns a `HitResult(target, composeRoot, hierarchy)` so `extractFromView` doesn't need a separate parent walk.

The `getLocationOnScreen` cost is still there — eliminating it requires manually translating coordinates during descent, which introduces correctness risk with scroll offsets, transforms, and clipping. In practice, the entire hit-test takes ~0.015-0.06ms for a typical screen, or ~0.05-0.2ms for a complex one — well under 1% of a 16.6ms frame budget. And unlike SR which snapshots up to 2/sec on every UI change, autocapture only runs this once per tap. Happy to revisit if profiling shows it's a concern on slower devices.

All 24 autocapture instrumented tests pass.

tylerjroach (approved):

PR #984, opened 2026-06-30
fix(analytics): expose Flags.shutdown() to release worker thread + executor (SDK-88)

tylerjroach:
## Summary

`FeatureFlagManager` creates a non-daemon `HandlerThread` + single-thread `ExecutorService` that previously had no public release path. The package-private `close()` existed only as a test-only escape hatch. OpenFeature's `FeatureProvider.shutdown()` lifecycle hook — the primary motivating caller — has no way to release those resources today.

- Add `shutdown()` to the public `MixpanelAPI.Flags` interface
- Rename `FeatureFlagManager.close()` → public `shutdown()` (same body, idempotent)
- Reframe docs around the OpenFeature lifecycle motivation

The `MixpanelProvider.shutdown()` wiring that actually invokes this is split into a follow-up PR — `openfeature-provider`'s `build.gradle.kts` pins a Maven coordinate for `mixpanel-android`, so it cannot compile against the new interface method until this ships as a release.

## Compatibility

Adding a method to a public interface is a source and binary break for anyone implementing `MixpanelAPI.Flags` themselves. Mockito mocks are unaffected. External implementers are rare — the interface exists mainly to be consumed via `mixpanel.getFlags()` — but this is worth calling out in the CHANGELOG for the release cutting the change.

## Test plan

- [x] `:analytics:testDebugUnitTest` passes (`FeatureFlagManagerTest.tearDown()` exercises `shutdown()` on every test)
- [x] `:openfeature-provider:compileDebugKotlin` still resolves against `mixpanel-android:8.5.0` on Maven and compiles cleanly (this branch no longer touches the provider)

## Context

Linear: SDK-88

### DIFF
--- a/analytics/src/main/java/com/mixpanel/android/mpmetrics/FeatureFlagManager.java
+++ b/analytics/src/main/java/com/mixpanel/android/mpmetrics/FeatureFlagManager.java
@@ -175,10 +175,14 @@ class FeatureFlagManager implements MixpanelAPI.Flags {
     }
   }
 
-  // Stop the worker HandlerThread and network executor so they don't outlive a test JVM
-  // worker (non-daemon threads would otherwise keep gradle's test JVM alive after a failure).
-  // Package-private; idempotent.
-  void close() {
+  /**
+   * Stops the worker {@link HandlerThread} and the network request executor.
+   * Invoked from lifecycle hooks that own this flag instance — notably
+   * an OpenFeature {@code FeatureProvider.shutdown()} hook — so the
+   * non-daemon threads don't outlive their intended owner. Idempotent.
+   */
+  @Override
+  public void shutdown() {
     try {
       mHandler.getLooper().quitSafely();
     } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


tylerjroach:
Pushed `896f371` addressing the Greptile P1.

`shutdown()` now gates on `mixpanel != null` — the internal flag set by the convenience constructor when we create the `MixpanelAPI` ourselves. When constructed via `MixpanelProvider(mixpanel.getFlags())` (the documented pattern), `mixpanel` is null and `shutdown()` is a no-op, preserving the shared `MixpanelAPI`'s flags HandlerThread. When constructed via `MixpanelProvider(context, token, options)`, `mixpanel != null` and we release the resources we own.

Updated `MixpanelProviderTest.shutdown` to verify `flags.shutdown()` is not called when flags were injected (renamed to `shutdown does not tear down injected flags`).

Note: the openfeature-provider build itself is currently red on this branch because `MixpanelAPI.Flags.shutdown()` isn't in the published `8.5.0` yet — this is the standard two-step release just like SDK-79 Android. Verified locally with `implementation(project(":analytics"))` — 41 tests pass.

rahul-mixpanel (approved):

PR #985, opened 2026-07-01
fix(analytics): isolate AnalyticsMessages and HttpService per instance [SDK-109]

tylerjroach:
## Summary

Two `MixpanelAPI` instances configured with different tokens and no explicit `instanceName` silently shared a single `AnalyticsMessages` worker (and therefore its `MPConfig` and `HttpService`), so events for the second instance were POSTed to the first instance's server URL. Independently, runtime `setServerURL` calls updated `MPConfig` endpoints but never refreshed the `HttpService` `serverHost` captured at construction.

## Changes

- `AnalyticsMessages.getInstance` now takes a `token` and keys `sInstances` by `MPConfig.resolveInstanceKey(instanceName, token)`, mirroring `MixpanelAPI.getInstance` and matching iOS `MixpanelManager` behavior.
- `AnalyticsMessages.getPoster` re-syncs `serverHost` (in addition to `backupHost` and the error listener) on every call, so `setServerURL` propagates to the poster.
- Added `HttpService.setServerHost` / `getServerHost`; the constructor delegates to `setServerHost` so the empty-check + `DEFAULT_SERVER_HOST` fallback lives in one place.
- Extracted the `instanceName ?? token` fallback into `MPConfig.resolveInstanceKey` — the five prior inline copies (plus three in tests) now call it, avoiding future drift.
- Removed the dead `mInstanceName` field on `AnalyticsMessages`.

## Workaround

Users on older SDK versions can side-step both bugs by giving each `MixpanelOptions` a distinct `instanceName` and setting the server URL via `MixpanelOptions.serverURL(...)` (which is applied before the first `getPoster()` call, so `HttpService` is constructed with the right host on first creation).

### DIFF
--- a/analytics/src/androidTest/java/com/mixpanel/android/mpmetrics/AutomaticEventsTest.java
+++ b/analytics/src/androidTest/java/com/mixpanel/android/mpmetrics/AutomaticEventsTest.java
@@ -156,7 +156,7 @@ public class AutomaticEventsTest {
 
             @Override
                 /* package */ PersistentIdentity getPersistentIdentity(final Context context, final Future<SharedPreferences> referrerPreferences, final String token, final String instanceName, final DeviceIdProvider deviceIdProvider) {
-                String instanceKey = instanceName != null ? instanceName : token;
+                String instanceKey = MPConfig.resolveInstanceKey(instanceName, token);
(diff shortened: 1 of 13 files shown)


tylerjroach:
Pushed `f9407b5` — swapped `assertEquals` → `assertSame` and imported the helper. 2 tests pass locally.

PR #988, opened 2026-07-21
feat(openfeature): map Source.Fallback.Reason to OpenFeature codes (SDK-130)

tylerjroach:
## Summary

Completes SDK-79 for Android at the OpenFeature layer. #981 landed \`Source.Fallback.Reason\` in the analytics SDK (shipped in 8.9.0) but the OpenFeature wrapper never actually consumed it — every fallback path still collapses to \`ErrorCode.FLAG_NOT_FOUND\` via the pre-SDK-79 reference-equality trick. This PR wires the wrapper up to the reason.

## Changes

- \`MixpanelProvider.kt\`: replace \`if (variant === fallbackVariant)\` with a check on \`variant.source\`; add \`mapFallback\` that dispatches on \`Source.Fallback.Reason\`.
- \`build.gradle.kts\`: bump \`mixpanel-android\` dep from \`8.5.0\` → \`8.9.0\` (the release that ships \`Source.Fallback.Reason\`). **This overlaps with [#986](https://github.com/mixpanel/mixpanel-android/pull/986), which independently bumps the same line to 8.9.0.** Whichever lands first, the other rebases trivially.
- \`gradle.properties\`: bump wrapper version \`0.1.0\` → \`0.2.0\` (new capability surface).
- \`MixpanelProviderTest.kt\`: \`setupFlag\` now stamps \`Source.network()\` on returned variants (previously they inherited the default \`Fallback(FLAG_NOT_FOUND)\` source and worked only because of the removed reference-equality check). New \`setupFallback\` helper + three tests, one per reason.

## Mapping table

| Reason | OpenFeature ErrorCode | Reason |
|---|---|---|
| \`FLAG_NOT_FOUND\` | \`FLAG_NOT_FOUND\` | \`DEFAULT\` |
| \`NOT_READY\` | \`PROVIDER_NOT_READY\` | \`ERROR\` |
| \`BACKEND_ERROR\` | \`GENERAL\` | \`ERROR\` |

Android's enum is narrower than PHP/Python/Ruby (no \`MISSING_CONTEXT_KEY\` / \`NO_ROLLOUT_MATCH\` — network-cache SDKs can't distinguish those). Adding a new reason to the base SDK will fail the exhaustive Kotlin \`when\` at compile time rather than silently succeeding.

## Note on sync-path coverage

The wrapper uses \`getVariantSync\`, which today can only return \`FLAG_NOT_FOUND\` or (bypassed by the wrapper's own \`areFlagsReady\` guard) \`NOT_READY\` — \`BACKEND_ERROR\` is only produced on the async path. The full mapping is still needed for forward-compatibility if the wrapper is ever migrated to async, and for correctness if the sync path is enhanced.

## Test plan

- [ ] CI passes (Android tests + code-quality checks)
- [ ] Verify existing test coverage still passes (all setup-flag tests should still work with \`Source.network()\` stamped)
- [ ] Verify the three new reason-mapping tests pass

Draft while [#986](https://github.com/mixpanel/mixpanel-android/pull/986) is open so we can coordinate the shared \`build.gradle.kts\` bump.

Refs: [SDK-130](https://linear.app/mixpanel/issue/SDK-130), [SDK-79](https://linear.app/mixpanel/issue/SDK-79)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/openfeature-provider/src/main/java/com/mixpanel/android/openfeature/MixpanelProvider.kt
+++ b/openfeature-provider/src/main/java/com/mixpanel/android/openfeature/MixpanelProvider.kt
@@ -208,6 +208,11 @@ class MixpanelProvider(private val flags: MixpanelAPI.Flags) : FeatureProvider {
                 providerNotReady(defaultValue)
             MixpanelFlagVariant.Source.Fallback.Reason.BACKEND_ERROR ->
                 generalError(defaultValue, "Flag \"$flagKey\" evaluation failed at backend")
+            // Fail closed on unrecognized reasons so a new value added to the
+            // base SDK enum doesn't compile-break this wrapper (Kotlin `when`
+            // exhaustiveness) nor throw NoWhenBranchMatchedException at
+            // runtime if an old wrapper JAR meets a new base value.
+            else -> generalError(defaultValue, "Unrecognized fallback reason $reason for flag: $flagKey")
         }
     }
 
--- a/openfeature-provider/src/main/java/com/mixpanel/android/openfeature/MixpanelProvider.kt
+++ b/openfeature-provider/src/main/java/com/mixpanel/android/openfeature/MixpanelProvider.kt
@@ -184,11 +184,31 @@ class MixpanelProvider(private val flags: MixpanelAPI.Flags) : FeatureProvider {
             return generalError(defaultValue, e.message)
         }
 
-        if (variant === fallbackVariant) {
-            return flagNotFound(key, defaultValue)
+        // A fallback source means the SDK had no real variant to serve. The
+        // reason discriminates why (missing key, not-ready, backend error) so
+        // callers get the OpenFeature error code the spec assigns to each,
+        // instead of collapsing every fallback to FLAG_NOT_FOUND.
+        val source = variant.source
+        if (source is MixpanelFlagVariant.Source.Fallback) {
+            return mapFallback(source.reason, key, defaultValue)
         }
 
         return coerce(variant)
+    }
+
+    private fun <T> mapFallback(
+        reason: MixpanelFlagVariant.Source.Fallback.Reason,
+        flagKey: String,
+        defaultValue: T
+    ): ProviderEvaluation<T> {
+        return when (reason) {
+            MixpanelFlagVariant.Source.Fallback.Reason.FLAG_NOT_FOUND ->
+                flagNotFound(flagKey, defaultValue)
+            MixpanelFlagVariant.Source.Fallback.Reason.NOT_READY ->
+                providerNotReady(defaultValue)
+            MixpanelFlagVariant.Source.Fallback.Reason.BACKEND_ERROR ->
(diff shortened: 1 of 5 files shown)


openfeature-provider/src/main/java/com/mixpanel/android/openfeature/MixpanelProvider.kt
```diff
@@ -184,13 +184,33 @@ class MixpanelProvider(private val flags: MixpanelAPI.Flags) : FeatureProvider {
...
+        flagKey: String,
+        defaultValue: T
+    ): ProviderEvaluation<T> {
+        return when (reason) {
+            MixpanelFlagVariant.Source.Fallback.Reason.FLAG_NOT_FOUND ->
+                flagNotFound(flagKey, defaultValue)
+            MixpanelFlagVariant.Source.Fallback.Reason.NOT_READY ->
+                providerNotReady(defaultValue)
+            MixpanelFlagVariant.Source.Fallback.Reason.BACKEND_ERROR ->
+                generalError(defaultValue, "Flag \"$flagKey\" evaluation failed at backend")
```
rahul-mixpanel, line 210: should we return an else case that returns a generalError that gives whatever new error that comes in here?

openfeature-provider/src/test/java/com/mixpanel/android/openfeature/MixpanelProviderTest.kt
```diff
@@ -449,4 +469,33 @@ class MixpanelProviderTest {
...
+
+    @Test
+    fun `fallback with BACKEND_ERROR reason maps to ErrorCode GENERAL`() {
+        setupFallback("backend-fail", MixpanelFlagVariant.Source.Fallback.Reason.BACKEND_ERROR)
+        val result = provider.getBooleanEvaluation("backend-fail", false, ImmutableContext())
+        assertEquals(false, result.value)
+        assertEquals(ErrorCode.GENERAL, result.errorCode)
+        assertEquals(Reason.ERROR.toString(), result.reason)
+    }
+
```
rahul-mixpanel, line 500: maybe add a test to cover a case how an unknown reason is handled. in mixed versions cases.

rahul-mixpanel (changes requested):

openfeature-provider/src/test/java/com/mixpanel/android/openfeature/MixpanelProviderTest.kt
tylerjroach: Can't do this because I can't provide an enum value that doesn't exist.

openfeature-provider/src/main/java/com/mixpanel/android/openfeature/MixpanelProvider.kt
tylerjroach: added. Especially if we don't consider enum additions as breaking changes, we need to make sure we don't lock ourselves into being unable to add reasons in the future.

rahul-mixpanel (approved):

PR #992, opened 2026-08-03
feat(analytics): walk-up-to-clickable-parent for autocapture

rahul-mixpanel:
## Summary
- When a non-interactive leaf view is tapped, walk up the view hierarchy to the nearest clickable ancestor and use its identity for `$el_id`
- Walk-up is always-on and uses interactivity checks (`isClickable()`/`isLongClickable()`), not identity checks — consistent with iOS behavior
- Extract `resolveIdentity()` from `resolveElementId()` to eliminate duplicate resource lookups per tap
- Fix ancestor hierarchy string to be built from the ancestor view, not the leaf
- Add walk-up demo test screens for both XML and Compose
- Add instrumented tests validating walk-up behavior across XML and Compose views

## Test plan
- [x] Walk-up resolves to clickable parent's `$el_id` for non-interactive leaves
- [x] Non-interactive leaf with own `contentDescription` still walks up (leaf identity ignored)
- [x] Clickable leaf does NOT walk up (owns its own click)
- [x] Nested clickables stop at the innermost clickable parent
- [x] Deep nesting (9 levels) finds clickable ancestor within 10-level limit
- [x] Non-interactive leaf with no clickable ancestor falls back to hash
- [x] Android build compiles successfully
- [ ] Run `./gradlew :analytics:connectedAndroidTest` on device/emulator

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/analytics/src/main/java/com/mixpanel/android/autocapture/SemanticExtractor.java
+++ b/analytics/src/main/java/com/mixpanel/android/autocapture/SemanticExtractor.java
@@ -103,8 +103,8 @@ final class SemanticExtractor {
             // Fall back to direct view extraction (XML views)
             ClickEvent.Builder viewResult = extractFromView(hit.target, hit.hierarchy, x, y);
 
-            // Walk up to clickable parent if enabled and the tapped view resolved to a hash fallback
-            if (viewResult != null && autocaptureOptions.isWalkUpToClickableParent()) {
+            // Walk up to clickable parent if the tapped view resolved to a hash fallback
+            if (viewResult != null) {
                 viewResult = walkUpToClickableParent(hit.target, viewResult, hit.hierarchy, x, y);
             }
 
--- a/analytics/src/main/java/com/mixpanel/android/autocapture/SemanticExtractor.java
+++ b/analytics/src/main/java/com/mixpanel/android/autocapture/SemanticExtractor.java
@@ -8,6 +8,7 @@ import android.view.accessibility.AccessibilityNodeInfo;
 import android.view.accessibility.AccessibilityNodeProvider;
 import android.widget.Button;
 import android.widget.CheckBox;
+import android.widget.Checkable;
 import android.widget.CompoundButton;
 import android.widget.EditText;
 import android.widget.ImageButton;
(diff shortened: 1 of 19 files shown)


analytics/src/main/java/com/mixpanel/android/mpmetrics/MixpanelAPI.java
```diff
@@ -1083,104 +1083,6 @@ private void initializeAutocapture(AutocaptureOptions options) {
     }
 
      /**
-     * Provides methods for tracking autocapture events. Events tracked through this class
-     * are automatically tagged with the {@code $mp_autocapture} property, which causes them
-     * to appear with an "[Auto]" prefix in the Mixpanel web app.
-     *
-     * <p>Access this via {@link MixpanelAPI#getAutocapture()}.
-     */
-    public class Autocapture {
```
rahul-mixpanel, line 1092: The previous PR ended up with duplicate `Autocapture` classes due to a rebase.

analytics/src/main/java/com/mixpanel/android/autocapture/SemanticExtractor.java
```diff
@@ -0,0 +1,802 @@
...
+
+        // Walk up to nearest clickable ancestor
+        ViewParent parent = tappedView.getParent();
+        int depth = 0;
+        while (parent instanceof View && depth < AutocaptureDefaults.MAX_ANCESTOR_SEARCH_DEPTH) {
+            View ancestor = (View) parent;
+            if (ancestor.isClickable() || ancestor.isLongClickable()) {
+                // Found a clickable ancestor — rebuild hierarchy from its perspective
+                return extractFromView(ancestor, buildHierarchyString(ancestor), x, y);
+            }
```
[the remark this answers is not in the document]
tylerjroach: This is valid.

analytics/src/main/java/com/mixpanel/android/autocapture/SemanticExtractor.java
```diff
@@ -739,6 +759,43 @@ private static String getSimpleClassName(@NonNull String fullyQualifiedName) {
...
+     * not identity. A non-interactive leaf (e.g., TextView inside a clickable
+     * LinearLayout) will always resolve to its clickable parent's identity,
+     * even if the leaf has its own contentDescription.
+     */
+    @NonNull
+    private static ClickEvent.Builder walkUpToClickableParent(
+            @NonNull View tappedView, @NonNull ClickEvent.Builder original,
+            @Nullable String hierarchy, float x, float y) {
+        // Only walk up if the tapped view is not interactive
+        if (tappedView.isClickable() || tappedView.isLongClickable()) {
```
tylerjroach: Can we add isContextClickable as well. Edge case but it is another option. Can you also add a test for a CheckBox and RadioButton, etc. `isCheckable` is available on these, and I'm not sure if they report as clickable or not. If not, we may need to type check for these.

Should point out that if a customer made their own touchListener, we may walk up when we shouldn't, but I think this is more of an edge case, and we have to make best guesses here.

tylerjroach (changes requested):
small addition ask.

analytics/src/main/java/com/mixpanel/android/autocapture/SemanticExtractor.java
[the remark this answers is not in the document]
rahul-mixpanel: When a clickable card (for example, a cart item) contains a disabled **Delete** button, tapping the disabled button should trigger a dead click and capture the button’s identity, not the parent card’s. That’s why we intentionally skip the enabled check here.

analytics/src/main/java/com/mixpanel/android/autocapture/SemanticExtractor.java
rahul-mixpanel: Good call raising `isContextClickable` — I looked into it and decided to leave it out intentionally. Context clicks come from right-click or stylus secondary button interactions, which are separate from taps. Since this walk-up only runs on the tap capture path, a view that's context-clickable but not tap-clickable wouldn't produce a tap event anyway, so including it wouldn't change behavior in practice. This holds in the disabled case too — `isClickable()` still returns `true` on a disabled view, so walk-up correctly stops there regardless of context-clickability.

For `isCheckable`, added `instanceof Checkable` to both guards in `walkUpToClickableParent` to match the accessibility node path. Also added tests for `CheckBox` and `RadioButton` — they're both clickable by default so they correctly keep their own identity without walking up.

tylerjroach (approved):

PR #996, opened 2026-08-19
feat(analytics): resolve $el_id for View and Compose autocapture paths

rahul-mixpanel:
## Summary

Implements `$el_id` resolution for autocapture across both rendering paths, as **SDK-owned behavior**. One resolver per path, since a Compose element is a semantics node rather than a `View`:

| Resolver | Governs |
|---|---|
| `DefaultViewElementIdExtractor` | XML layouts + everything React Native renders |
| `DefaultComposeElementIdExtractor` | Jetpack Compose |

Both are internal (package-private). This release deliberately ships **no host-app extension point** for `$el_id`: while autocapture is in beta the resolution rules need to stay free to change, and a pluggable interface would freeze them into public API before we have real-world feedback. `AutocaptureOptions` and `AutocaptureManager` are untouched by this PR.

Rebased onto `feat/autocapture`. Companion PR: the same resolution on iOS — mixpanel/mixpanel-swift#764.

## Resolution order

**View path** — `DefaultViewElementIdExtractor`

1. React Native `nativeID` — the JS-side prop, read from the view tag keyed by a dynamically resolved `view_tag_native_id` resource id (result cached; no compile-time React Native dependency)
2. Android resource entry name — e.g. `checkout_button` for `@+id/checkout_button`
3. `contentDescription` — only when the developer intentionally exposed it (see the guard below)
4. `<SimpleClassName>_<hash>`

**Compose path** — `DefaultComposeElementIdExtractor`

1. `Modifier.testTag("…")` — the Compose analogue of a resource id: developer-assigned, never user-visible
2. `contentDescription` from `semantics { }`
3. `<TagName>_<hash>`

`$attr-aria-label` still comes from `contentDescription` alone, regardless of which source won `$el_id`. There is no `nativeID` step on the Compose path — React Native renders through the legacy View hierarchy.

The anonymous fallback hashes the element's **position in the hierarchy**, not its identity. `View.hashCode()` is an identity hash: it changes on every launch and as list rows are recycled, so an element with no id would get a different `$el_id` every session and could never be grouped. `String.hashCode()` is JDK-specified, so the structural hash is stable across processes and devices.

`ComposeElementInfo` carries the tapped node's semantics as **plain values** (`testTag`, `contentDescription`, `role`, `tagName`, anonymous id). That keeps Compose types out of the resolution logic, so it can be tested without a Compose runtime and apps without Compose never load a Compose class because of it.

## PII guard: labels on views excluded from accessibility

React Native expresses `accessible={false}` by clearing **focusability**, not by clearing `importantForAccessibility` — the view stays important for accessibility with its `contentDescription` intact. The previous guard only tested importance, so a label on a view the developer excluded from accessibility was reported as both `$el_id` and `$attr-aria-label`. Measured in the sample app: a button with `accessible={false}` and `accessibilityLabel="Account ending 4321"` reported that string verbatim.

`AutocaptureDefaults.intentionalContentDescription(View)` now requires `isImportantForAccessibility()` **and**, for React Native views only, `isFocusable()`. The focusability requirement is scoped to React Native views (`com.facebook.react.` class prefix) on purpose: a native Android view can be clickable without being focusable while carrying a perfectly intentional `android:contentDescription`. Both `$el_id` resolution and the `$attr-aria-label` property call the shared helper, so the two can no longer disagree about whether a label is safe to report.

## Behavior changes

- Resource entry name now outranks `contentDescription` on the View path (previously reversed)
- Compose `testTag` now outranks `contentDescription` (previously reversed)
- A label on a React Native view with `accessible={false}` is no longer reported

Existing screens that rely on `contentDescription` are unaffected in practice: React Native's dynamic ids have no resource entry name, so the label still wins there.

## Testing

79/79 autocapture instrumented tests pass on a Pixel 4 (API 35) emulator.

- `DefaultViewElementIdExtractorTest` (14) — each priority step, the priority pairs, the accessibility guards (including the React Native focusability shape, and a native clickable-but-not-focusable view that must keep its label), empty and non-String `nativeID` tags, hash shape and stability. Declares `view_tag_native_id` in `androidTest/res` so the real `getIdentifier`/`getTag` path runs unmocked.
- `DefaultComposeElementIdExtractorTest` (5) — `testTag` > `contentDescription` > anonymous, empty values treated as absent, `ComposeElementInfo` accessors. Needs no Compose runtime, which is the point of the plain-value holder.
- `ElementIdInstrumentedTest` (4) + `ElementIdTestActivity` — real Espresso taps for the View path, including that an `accessible={false}` label reaches neither `$el_id` nor `$attr-aria-label`.
- `ComposeElementIdInstrumentedTest` (1) + `ElementIdComposeTestActivity` — a real tap proving end-to-end that `testTag` wins over `contentDescription`. Kept as a separate activity from `AutocaptureComposeTestActivity`: that one lays its elements out in a long scrolling column and several tests tap by coordinate, so inserting anything into it pushes later elements off-screen.
- `FakeReactViewGroup`, declared in `com.facebook.react.views.view` under `androidTest`, lets the React-Native-specific guard be tested without adding a React Native dependency to the SDK.

Also verified in a real app: `mixpanel-react-native`'s `MixpanelStarter` sample, built against a local `mavenLocal` publish of this branch. `nativeID="checkout_button"` reports `checkout_button` while the accessibility label still travels as `$attr-aria-label`; `accessible={false}` now yields `ReactViewGroup_<hash>` with no label; `accessible={true}` and labels with no `accessible` prop are unchanged. The structural hash was confirmed identical across two launches on a physical device.

## Drive-by test fixes

Two coordinate-based taps that Android 14+ targeted input injection rejects when the target is off-screen or behind the system bars:

- `AutocaptureXmlInstrumentedTest.testInvisibleView_NoEventCaptured` — threw `IllegalArgumentException: Targeted input event injection ... not directed at a window owned by uid ...`, and could silently skip its assertion
- `AutocaptureWalkUpXmlInstrumentedTest.tapViewById` (shared helper) — silently reported `navigationBarBackground` as `$el_id` for the last child of its `ScrollView`

Both now scroll the target in with `requestRectangleOnScreen` and assert the point falls inside `getWindowVisibleDisplayFrame` before injecting, so a mis-aimed tap fails with a clear message instead of a confusing value mismatch.


tylerjroach:
Who is creating the non default extractors? Will that code be on the RN repo?

tylerjroach (approved):

PR #1001, opened 2026-08-25
feat(analytics): Add semver and date custom operator support

efahk:
In common lib so that they could eventually be used by the hand rolled json logic fxns


common/src/main/java/com/mixpanel/android/customoperators/Rfc3339.kt
```diff
@@ -0,0 +1,83 @@
...
+        val calendar: Calendar = GregorianCalendar(TimeZone.getTimeZone("UTC"))
+        calendar.clear()
+        // A lenient calendar rolls an impossible date such as 2026-02-30 forward into a real instant,
+        // so a malformed property would match a date rule instead of failing closed.
+        calendar.isLenient = false
+        calendar.set(year, month - 1, day, hour, minute, second)
+        val wallSeconds =
+            try {
+                calendar.timeInMillis / 1000L
+            } catch (e: IllegalArgumentException) {
```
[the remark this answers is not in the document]
tylerjroach: Don't think we are worrying about pre-epoch here

common/src/main/java/com/mixpanel/android/customoperators/SemanticVersion.kt
```diff
@@ -0,0 +1,150 @@
...
+
+import androidx.annotation.RestrictTo
+
+/**
+ * Semantic Versioning 2.0.0 comparison, per [semver.org](https://semver.org).
+ *
+ * Kept free of any JsonLogic type so the same ordering can back either engine.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object SemanticVersion {
```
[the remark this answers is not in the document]
tylerjroach: A common release needs to be pushed before mixpanel-android can then implement into jsonlogic custom operator. This is just housing the inner implementation.

common/src/test/java/com/mixpanel/android/customoperators/CustomOperatorVectorsTest.kt
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private fun datetimeCompare(): Boolean {
+        val actual = (subject as? String)?.let { Rfc3339.toUnixSeconds(it) } ?: return false
+        val wanted = (target as? Number)?.let { Rfc3339.epochMillisToUnixSeconds(it.toDouble()) } ?: return false
+        return comparatorMatches(actual - wanted, symbol)
+    }
+
+    @Test
+    fun goldenVector() {
```
tylerjroach, line 93: Only concern here would be what if we fail to load the list, would we actually get a test failure, or would no tests run and it would look like the build is passing.

tylerjroach (dismissed):

common/gradle.properties
```diff
@@ -1,4 +1,4 @@
-VERSION_NAME=1.0.1
+VERSION_NAME=1.0.2
```
tylerjroach: revert

common/src/test/java/com/mixpanel/android/customoperators/CustomOperatorVectorsTest.kt
efahk: If we fail to load the list, we'd get a test failure. But i've added an additional check to also fail if there are no cases read

tylerjroach (approved):

PR #1003, opened 2026-08-28
feat(session-replay): wireframes (beta)

tylerjroach:
## What a wireframe is

Session replay sends screenshots, which give Mixpanel's AI summaries nothing to read. A wireframe is a lightweight text outline of each captured screen — the visible elements, what kind each one is, what it says, and where it sits — shipped alongside the screenshot as an rrweb Custom event, so a session can be summarized without image recognition.

Off by default. Nothing is captured until `MPSessionReplayConfig.wireframesOptions` is set, and existing integrations see no change on upgrade.

## Masking

Enforced structurally in four layers, so anything hidden in the replay video is textless in the wireframe. The element still ships as a `role + bounds` shell, so the shape of the screen survives without its content:

1. **View-level** — auto-masked types, `mpReplaySensitive`, and text-entry fields, which are always masked and never overridable.
2. **Geometric leak-prevention** — an element whose bounds intersect any mask rect the screenshot is drawing loses its text, even if it was never marked sensitive itself.
3. **Declared text** — `mpWireframeText` substitutes copy the developer authored.
4. **Sensitive rules** — the caller's own content rules, run last over everything above.

## Also in here

- Server-side wireframe kill switch, requested only when the app opted in. A `false` drops the wireframe payload while replay keeps recording.
- Debug emitter handing each captured wireframe back with the reason every element's text was kept or removed.
- A `session-replay:wireframe-goldens` module with **98 coordinate goldens** pinning roles, bounds and masking decisions off-device via Paparazzi.

## Beta framing

Every public wireframe type carries the same note, and there is a matching one-time notice at `initialize` in debuggable builds:

> **Beta.** Wireframes are in beta. Before shipping to production, inspect the wireframes your app produces with the wireframe debug emitter and confirm that no sensitive information is captured.


tylerjroach:
@greptileai review

tylerjroach:
@greptileai

tylerjroach:
@greptileai re-review

tylerjroach:
@greptileai

tylerjroach:
@greptileai

session-replay/src/main/java/com/mixpanel/android/sessionreplay/sensitive_views/SensitiveViewManager.kt
```diff
@@ -412,69 +533,347 @@ object SensitiveViewManager {
...
 
         // Skip nodes with zero or negative dimensions after int conversion
-        if (rect.width() <= 0 || rect.height() <= 0) return
+        if (rect.width() <= 0 || rect.height() <= 0) return null
 
-        addOrUpdateEntry(accumulator, rect, maskDecision)
+        return rect
     }
 
     private data class ViewContext(
```
rahul-mixpanel, line 545: **Masking gap: masked-ancestor provenance stops at toolkit boundaries.**

The `maskedAncestor` threading added to the Compose semantics walk fixes exactly the right bug — but the same hole is still open on the other two boundaries:

1. **View→View:** `ViewContext` carries only `isInsideSafeContainer`. A child's masking is decided purely from its own attributes, so with `AutoMaskedView.Text` off, a `TextView` inside an `addSensitiveView`'d container is protected only by the Layer-2 geometric strip — the mechanism the Compose KDoc in this PR explicitly calls unreliable. On API 26+, `getGlobalVisibleRect` doesn't clip to a parent with `clipChildren=false`, so a translated child lands outside every mask rect and its text ships with `MaskDecision.NONE`.
2. **View→Compose:** `collectMaskableNodes` always seeds the semantics walk with `maskedAncestor = null`, even when the enclosing `ComposeView` (or a View ancestor) was itself masked — the walk never learns it. A `ComposeView` partially scrolled off inside an Android `ScrollView` leaks the same way: the mask rect covers only the visible portion while `boundsInWindow` is clamped to the compose root, not the window.

Suggestion: thread the enclosing view's mask decision into `ViewContext` (next to `isInsideSafeContainer` — the propagation channel already exists) and pass it into the `collectMaskableNodes` seed, so Layer 1 is authoritative on both toolkits and Layer 2 becomes pure defense-in-depth.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/sensitive_views/SensitiveViewManager.kt
```diff
@@ -489,6 +888,14 @@ object SensitiveViewManager {
             // Skip views that aren't visible (including children of GONE/INVISIBLE parents)
             if (!currentView.isShown) continue
 
+            // A fully transparent view paints nothing, so neither its pixels nor its text may
+            // reach the replay — `isShown` only covers GONE/INVISIBLE. Alpha is multiplicative
+            // down the hierarchy, so `continue` correctly drops the whole subtree: children are
+            // only enqueued at the bottom of this loop. It also means such a view contributes no
+            // mask region, which is right — there is nothing painted to cover.
+            // Matches Flutter's `Opacity(0)` filter (reference suite fixture 44).
+            if (currentView.alpha <= 0f) continue
```
rahul-mixpanel, line 902: **Masking gap: the alpha-0 skip drops the mask rect for fading sensitive views.**

This skip runs *before* mask-rect collection, so a sensitive view at alpha 0 contributes no `boundsSnapshot` entry. If it was the only sensitive view on screen, `needsMasking` is false and the whole revalidate/discard block in `ScreenRecorder` is skipped. Meanwhile `captureUsingPixelCopy` suspends up to 300 ms and reads the *newest* surface frame — a login/payment form fading in via `ViewPropertyAnimator` (which animates `View.alpha`, so neither the `transitionAlpha` guard nor the legacy-Animation check fires) can be partially visible in the captured pixels with no mask painted over it. On master, alpha-0 sensitive views still produced mask rects.

Suggestion: keep the skip for wireframe/text extraction, but still record the mask rect for sensitive candidates at `alpha <= 0` — or treat it like `hasActiveTransition` and discard the frame.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/tracking/EventPublisher.kt
```diff
@@ -27,4 +28,12 @@ class EventPublisher {
     fun publishSessionEvent(event: RawScreenshotEvent) {
         subscribers.forEach { it.receivedScreenshotEvent(event) }
     }
+
+    /**
+     * Publishes a fully-formed [SessionEvent] (e.g. an rrweb Custom event) that doesn't need
+     * any further server-side encoding. The wireframe pipeline uses this path.
+     */
+    fun publishCustomEvent(event: SessionEvent) {
```
rahul-mixpanel, line 37: **Thread safety: this PR adds the first off-main-thread publisher over an unsynchronized list.**

`WireframeEmitter.emit` runs in `withContext(Dispatchers.Default)` and calls `publishCustomEvent`, while `subscribe`/`unsubscribe` run on Main during instance init/deinit (`MPSessionReplay.initialize` re-init path). Pre-PR, all publishers were main-thread-only, so the plain `mutableListOf` was fine; now a re-init while an emit is mid-`forEach` can throw `ConcurrentModificationException` (caught in `ScreenRecorder`, so no crash — the wireframe is just silently dropped), and with no happens-before edge the Default worker may never observe a Main-thread `subscribe`, silently dropping all wireframe events for the session.

Suggestion: `CopyOnWriteArrayList`, or synchronize subscribe/unsubscribe/publish.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/MPSessionReplayInstance.kt
```diff
@@ -354,6 +371,14 @@ class MPSessionReplayInstance(
             // Enable debug mask overlay if configured
             debugMaskOverlayManager?.enable()
 
+            // Wireframe dedup is per *session*, not per SDK lifetime. The emitter is built
+            // in init and outlives a stop/start cycle, so without this the new replay's
+            // first frame is compared against the previous replay's last one — and since
+            // stopRecording cleared initialScreenshotCaptured, scheduleScreenshotCapture()
+            // below always captures, so a background/foreground onto an unchanged screen
+            // reliably ships an opening screenshot with no mp_wireframe to describe it.
+            ScreenRecorder.shared.wireframeEmitter?.resetDedup()
```
rahul-mixpanel, line 380: **Lifecycle leak: stopping mid-gesture orphans the 150 ms `recordTimer` and leaves stale touch state.** (Anchored here since `stopRecording()` itself isn't in the diff.)

`stopRecording()` sets `hasStartedRecording = false` but neither cancels `recordTimer` nor resets `TouchEventRecorder`. The `if (hasStartedRecording)` guard in `onTouch` then swallows the gesture's terminal UP/CANCEL, so `onTouchEnd` — the only cancel site for `recordTimer` — never fires. The `Timer` thread wakes every 150 ms indefinitely (each `scheduleScreenshotCapture` early-returns, but the wakeups persist) and survives even `deinitialize()`, which cancels only `pendingScreenshotJob`. Master's 200 ms `endOfScrollRunnable` watchdog used to self-heal the common scroll case; this PR removed it and relies solely on delivered UP/CANCEL.

Also, `isTracking` stays true with buffered `pendingSamples`, so a finger held across a stop/start cycle feeds MOVEs with no TOUCH_START and mixes old-session samples into the new session's first TOUCH_MOVE batch.

Suggestion: have `stopRecording()` cancel `recordTimer` and call a new `TouchEventRecorder.reset()` (drop pending samples, clear `isTracking`).

session-replay/src/main/java/com/mixpanel/android/sessionreplay/utils/MPSessionReplayEncoder.kt
```diff
@@ -94,16 +102,20 @@ object SessionReplayEncoder {
...
         val replayJson = ReplayJSONTemplate.mainEventJSON(
             imageBase64 = android.util.Base64.encodeToString(image, android.util.Base64.NO_WRAP),
-            timestamp = getCurrentTimestampMillis()
+            timestamp = timestamp
         )
-        val x = json.decodeFromString<SessionEvent>(replayJson).also { it.timestamp = getCurrentTimestampMillis() }
+        val x = json.decodeFromString<SessionEvent>(replayJson).also { it.timestamp = timestamp }
         x
     } catch (e: Exception) {
         Logger.error("Error decoding main event JSON: ${e.message}")
```
rahul-mixpanel, line 111: **The "image is now guaranteed to ship" invariant the wireframe emit relies on is false.**

`ScreenRecorder`'s comment says a non-null return is always published — but published ≠ enqueued. The screenshot still flows through `mainSessionEvent`, whose catch returns `null` (silently dropped at `EventHandler`'s `event?.let { enqueue }`), and `incrementalSessionEvent` has no try/catch at all. Custom events bypass the encoder entirely, so on an encode failure (e.g. OOM base64-ing a large frame) the `mp_wireframe` event ships while its screenshot doesn't — and since `lastPayloadHash` was already committed, the next successful capture of the same static screen is deduped and ships wireframe-less.

The nastiest variant: on the *initial* frame, a `mainSessionEvent` failure drops the rrweb FullSnapshot, yet `captureScreenshot` returned non-null so `initialScreenshotCaptured = true` and the full snapshot is never retried.

Suggestion: make the encode path symmetric with the wireframe (catch + retry/report instead of null-drop; add try/catch to `incrementalSessionEvent`), or defer the wireframe emit/hash-commit until the screenshot event is actually enqueued.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/wireframe/WireframeEmitter.kt
```diff
@@ -0,0 +1,413 @@
...
+            // Deliberately content-free: element text is customer data, and the SDK log is
+            // a channel they did not opt into and cannot filter. Anyone who wants to see
+            // what a wireframe says installs DebugOptions.wireframeEmitter, which is
+            // opt-in, stays on device, and reports each element's maskDecision alongside
+            // its text. Counts and viewport are enough to confirm capture is running.
+            Logger.debug {
+                "Emitting wireframe (elements=${elements.size}, viewport=${viewportW}x$viewportH)"
+            }
+
+            lastPayloadHash = payloadHash
```
rahul-mixpanel, line 168: **Dedup hash is committed before the publish it's supposed to represent.**

If `publishCustomEvent` throws (a CME from the unsynchronized subscriber list, or `RejectedExecutionException` from an `eventSerialQueue` shut down by a concurrent deinit), the exception is swallowed by `ScreenRecorder`'s catch — but the hash stays committed. Every subsequent identical frame then hits the dedup check and is suppressed, so screenshots keep shipping with no `mp_wireframe` until the screen visibly changes or the next `startRecording()` resets dedup.

Suggestion: move the hash commit after a successful publish — one-line reorder.

session-replay/CLAUDE.md
```diff
@@ -0,0 +1,708 @@
...
+
+## Wireframe Capture Notes (2026-07-29)
+
+**The `mp_wireframe` event is a cross-platform contract — Android, Flutter, and
+iOS must all follow it.** This section records how the Android implementation
+realizes that contract and where the reasoning isn't obvious from the code; the
+same section exists in the Flutter session-replay `CLAUDE.md`. If the platforms
+disagree, that is a bug in one of them, not a local Android choice.
+
+- **Only four semantic roles are emitted:** `text`, `button`, `input`, `image`
```
rahul-mixpanel: **Stale cross-platform contract: docs say four roles, the code ships ten.**

`WireframeType` now has Text/Input/Image/Button/Link/Header/Checkbox/Switch/Radio/Tab (and this file's own 2026-08-26 section documents that change), but this line — inside the section declared "a cross-platform contract — Android, Flutter, and iOS must all follow it" — still says only `text`/`button`/`input`/`image`. That's exactly the sentence the Flutter/iOS authors will implement against. The same stale four-role wording is in `WireframePayload.kt:20` ("role: one of …"), `WireframesOptions.kt:19`, and the comment at `SensitiveViewManager.kt:992`.

Related stale comment: `WireframePaparazziHarness.kt:110` says "`deinitialize()` deliberately keeps registered *classes*" — the exact opposite of what `deinitialize()` now does (`retainAll(setOf(EditText))` + `_customerSensitiveClasses.clear()`, decided 2026-08-26 per this same CLAUDE.md).

Suggestion: update all five sites in this PR before the wrong contract propagates to the sibling SDKs.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/tracking/TouchEventRecorder.kt
```diff
@@ -2,124 +2,127 @@ package com.mixpanel.android.sessionreplay.tracking
...
+ *
+ * Nothing here is deferred: batches drain on the next sampled move or when the gesture
+ * ends, so no position is held behind a timer and every event carries the `eventTime` of
+ * the `MotionEvent` that produced it.
+ */
 class TouchEventRecorder(
-    private val context: Context,
+    context: Context,
     private val touchEventListener: TouchEventListener,
     internal val windowOffsetProvider: () -> IntArray = {
```
rahul-mixpanel, line 34: **Perf: the window offset is re-resolved per touch sample inside dispatch.**

`scalePoint` calls `windowOffsetProvider()` for every accepted sample (one per 50 ms during a drag, ~20/s) and every interaction, on the UI thread inside touch dispatch. This default provider allocates an `IntArray(2)`, scans `Curtains.rootViews` with a filter, and calls `getLocationOnScreen` (a parent-chain walk) each time — and the window doesn't move mid-gesture (if it did, screenshot alignment would be broken anyway).

Suggestion: resolve the offset once per gesture in `onGestureStart` (keeping the injectable provider for tests, just called once) and reuse it for the gesture's samples and end event. Same could apply to `epochOffsetProvider` — though computing that per-sample is what protects against sleep drift, so if you cache it per-gesture, consider clamping `timeOffset` with `coerceAtMost(0)` in `EventHandler` either way.

session-replay/CLAUDE.md
tylerjroach: There is no contract to role values

session-replay/src/main/java/com/mixpanel/android/sessionreplay/wireframe/WireframeEmitter.kt
tylerjroach: No need for this.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/utils/MPSessionReplayEncoder.kt
tylerjroach: This would be pre-existing if an issue.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/sensitive_views/SensitiveViewManager.kt
tylerjroach: Will push a change to do this since it matches more with Flutter but this would have been expected on Android because the actual content will be exposed on the screenshot.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/sensitive_views/SensitiveViewManager.kt
tylerjroach: This would break the screenshot side. If a change happened, it would fail the validation diff and the whole screen would be discarded.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/tracking/EventPublisher.kt
tylerjroach: not much practical impact here but will push to avoid more call outs.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/MPSessionReplayInstance.kt
tylerjroach: pushing fix.

session-replay/src/main/java/com/mixpanel/android/sessionreplay/tracking/TouchEventRecorder.kt
tylerjroach: nice optimization. pushed.

tylerjroach:
@greptileai

rahul-mixpanel (approved):

PR #1008, opened 2026-09-02
feat(session-replay): send app bundle_id and build_number as settings API query params

ketanmixpanel:
This pull request enhances the Session Replay SDK's remote settings logic by including the host app's bundle ID and build number in settings requests, enabling server-side SDK blocking per app and build version. It also introduces robust handling for missing app metadata and adds comprehensive unit tests for the new logic.

**Remote Settings Request Enhancements:**

* The `RemoteSettingsService` now includes the app's bundle ID and build number as query parameters in settings requests, using new helper methods in `DeviceInfo`. These values are omitted if unavailable, and a warning is logged in that case. [[1]](diffhunk://#diff-80f1346c166996f725a08f803653e79b35acc0567e8a301fd3d23f9243981a44L31-R34) [[2]](diffhunk://#diff-80f1346c166996f725a08f803653e79b35acc0567e8a301fd3d23f9243981a44R46-R48) [[3]](diffhunk://#diff-80f1346c166996f725a08f803653e79b35acc0567e8a301fd3d23f9243981a44L60-R86)

* Optional overrides for bundle ID and build number can be passed to `RemoteSettingsService`, allowing for custom or test values.

**Device Info Utilities:**

* The `DeviceInfo` utility adds `bundleId(context)` and `buildNumber(context)` methods to retrieve the app's package name and build number, with safe error handling to return `null` if unavailable.

**Testing Improvements:**

* New unit tests in `DeviceInfoTest` cover normal and error cases for bundle ID and build number retrieval.

* The `RemoteSettingsServiceTest` is extended to verify that settings requests include or omit bundle ID and build number as appropriate, and that existing parameters remain unaffected. [[1]](diffhunk://#diff-0c33ce2ac5667d61b8c5099672108a57a784c69e71506e6861d0143032228c77L157-R174) [[2]](diffhunk://#diff-0c33ce2ac5667d61b8c5099672108a57a784c69e71506e6861d0143032228c77R334-R397)

**Versioning:**

* The demo app's `versionCode` and `versionName` are updated to reflect a new release.## Summary

Android counterpart to the iOS change. Sends the host app's package name and build
number as query parameters on the `/settings` request, so the backend can extend
server-side SDK blocking to work **per app ID and per app build version**.

## What changed

- `DeviceInfo` gains `bundleId(context)` and `buildNumber(context)`, mirroring where
  iOS puts them. They take a `Context` because `DeviceInfo` is a stateless object and
  Android has no global equivalent of iOS's `Bundle.main`.
  - `bundleId` → `context.packageName`
  - `buildNumber` → `PackageInfo.longVersionCode` on API 28+, `versionCode` below
    (minSdk is 21). This is the analog of iOS `CFBundleVersion` — deliberately
    `versionCode`, not `versionName`.
  - Both return `null` on failure rather than the `0` / `"Unknown"` sentinels its
    siblings use, because omitting the param is better than sending a fake value.
- `RemoteSettingsService` appends each param only when non-nil, and gains optional
  trailing `bundleIdOverride` / `buildNumberOverride` constructor params for testing.
  Being optional and trailing, every existing call site compiles unchanged — including
  the `remoteSettingsServiceFactory` lambda in `MPSessionReplay`.

### API URL
```
2026-09-02 20:05:36.442  4272-4297  MixpanelSessionReplay   com.mixpanel.sessionreplaydemo       D  [Mixpanel Session Replay - Network.kt - func executeAPIRequest-0E7RQCE] (DEBUG) - https://api.mixpanel.com/settings?recording=1&sdk_config=1&%24os=Android&mp_lib=android-sr&%24lib_version=1.4.0&bundle_id=com.mixpanel.sessionreplaydemo&build_number=22

```

### Note on `BuildConfig.VERSION_CODE`

Not usable here. `session-replay` is a `com.android.library` module whose `BuildConfig`
has one hand-declared field (`VERSION_NAME`); AGP does not generate `VERSION_CODE` for
library modules, so it would not compile. Even if declared it would carry the *SDK's*
version — duplicating the existing `$lib_version` param — rather than the host app's
build, so it could not support blocking per app build version.

## Test plan

`./gradlew :session-replay:testDebugUnitTest :session-replay:ktlintCheck`

- **169 tests, 0 failures** (was 164), ktlint clean
- New `DeviceInfoTest` (5 tests): package name returned, versionCode returned, and null
  on each failure mode — `packageName` throwing, `NameNotFoundException`, and
  `packageManager` throwing
- New `RemoteSettingsServiceTest` cases: params present when injected, both absent when
  the context can't supply them
- Updated `testCheckSettingsIncludesSdkConfigQueryParam`, which asserts the whole
  ordered `queryItems` list

**Known gap:** the API-28+ `longVersionCode` branch isn't covered — JVM unit tests
report `Build.VERSION.SDK_INT` as `0`, so tests exercise the deprecated `versionCode`
path only. A Robolectric test with `@Config(sdk = [28])` would close it.

Fixes SDK-153

### DIFF
--- a/session-replay/src/main/java/com/mixpanel/android/sessionreplay/services/RemoteSettingsService.kt
+++ b/session-replay/src/main/java/com/mixpanel/android/sessionreplay/services/RemoteSettingsService.kt
@@ -73,8 +73,8 @@ internal open class RemoteSettingsService(
 
         // Include app bundle ID and build number to enable server-side SDK blocking
         // per app ID and app build version. Either is omitted if unavailable.
-        bundleId?.let { queryItems.add("bundleId" to it) }
-        buildNumber?.let { queryItems.add("buildNumber" to it) }
+        bundleId?.let { queryItems.add("bundle_id" to it) }
+        buildNumber?.let { queryItems.add("build_number" to it) }
         if (bundleId == null || buildNumber == null) {
             Logger.warn("Incomplete app info for settings request: bundleId=$bundleId, buildNumber=$buildNumber")
         }
--- a/session-replay/src/main/java/com/mixpanel/android/sessionreplay/services/RemoteSettingsService.kt
+++ b/session-replay/src/main/java/com/mixpanel/android/sessionreplay/services/RemoteSettingsService.kt
@@ -10,6 +10,7 @@ import com.mixpanel.android.sessionreplay.network.Network
 import com.mixpanel.android.sessionreplay.network.RequestMethod
 import com.mixpanel.android.sessionreplay.network.SdkConfig
 import com.mixpanel.android.sessionreplay.network.SettingsResponse
+import com.mixpanel.android.sessionreplay.utils.DeviceInfo
 import com.mixpanel.android.sessionreplay.utils.EndPoints
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
@@ -28,7 +29,9 @@ internal open class RemoteSettingsService(
     private val network: Network = Network(),
     private val version: String,
     private val mpLib: String,
-    private val serverUrl: String = EndPoints.DEFAULT_BASE_URL
+    private val serverUrl: String = EndPoints.DEFAULT_BASE_URL,
+    bundleIdOverride: String? = null,
+    buildNumberOverride: String? = null
 ) {
     companion object {
         private const val SETTINGS_TIMEOUT_MS = 5000L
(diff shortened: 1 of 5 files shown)


tylerjroach (dismissed):

session-replay/src/main/java/com/mixpanel/android/sessionreplay/services/RemoteSettingsService.kt
```diff
@@ -33,7 +34,9 @@ internal open class RemoteSettingsService(
     private val network: Network = Network(),
     private val version: String,
     private val mpLib: String,
-    private val serverUrl: String = EndPoints.DEFAULT_BASE_URL
+    private val serverUrl: String = EndPoints.DEFAULT_BASE_URL,
+    bundleIdOverride: String? = null,
+    buildNumberOverride: String? = null
```
rahul-mixpanel: can we make these `private val` and assign null as default value? trying to preserve the immutability of these params once the `RemoteSettingsService` is created.

rahul-mixpanel (changes requested):

session-replay/src/main/java/com/mixpanel/android/sessionreplay/services/RemoteSettingsService.kt
ketanmixpanel: fixed

rahul-mixpanel (approved):

